PR #281, opened 2014-07-22
Add API version discovery

emonty:
Since docker-py handles API differences well already, in many cases
it's entirely possible to write scripts that can talk to different
versions of docker without the user knowing in advance what the API
version is. Add an opt-in ability for a consumer to request that
docker-py figure out what version the server is, by making an
unversioned call to /version and retreiving the version if
Client(version=False)


shin-:
Closing in favor of #505

PR #284, opened 2014-07-29
Add ability to set timeout on client.

wt:
Signed-off-by: Warren Turkal <wt@signalfuse.com>

### DIFF
--- a/docker/client.py
+++ b/docker/client.py
@@ -757,6 +757,9 @@ class Client(requests.Session):
                                       params={'term': term}),
                             True)
 
+    def set_timeout(self, timeout):
+        self._timeout = timeout
+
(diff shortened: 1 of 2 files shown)


gtaylor:
@wt Agreed. Referencing a private attribute gets flagged as a warning in Pylint and PEP8, IIRC. 

Perhaps we could rename ``_timeout`` to ``timeout``, thus bringing it public? Probably not enough going on here to warrant a full method. A property or a public attrib would be peachy.

I don't know if docker-py has a policy on this yet, but Django and many other projects offer no backwards compatibility guarantees for private vars/attribs/funcs/methods. Worst case, we could easily alias ``_timeout`` to ``timeout`` if we were concerned.

shin-:
WIth #475 merged, we can now close this. Thanks!

PR #306, opened 2014-08-27
Add support for 'cpuset'

### LINKED ISSUE #254: CPU Affinity
Is there a way to use the new `cpuset` flag of the `run` command? Example:

```
docker run --rm --cpuset 0-1 busybox true
```

See: https://github.com/dotcloud/docker/issues/4587

groundeffect:
Fairly simple addition. Also tested it on the shell: `/sys/fs/cgroup/cpuset/docker/{container_id}/cpuset.cpus` reflected the cores that I passed in as arguments to `create_container()`.

Let me know if i missed something.
Would close #254.

### DIFF
--- a/docker/client.py
+++ b/docker/client.py
@@ -102,8 +102,8 @@ class Client(requests.Session):
                           mem_limit=0, ports=None, environment=None, dns=None,
                           volumes=None, volumes_from=None,
                           network_disabled=False, entrypoint=None,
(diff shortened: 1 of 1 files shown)


cmdelatorre:
FYI, the documentation is out-of-date here. The reference for [create container](http://docker-py.readthedocs.org/en/latest/api/#create_container) does not specify the *cpuset* param.

PR #317, opened 2014-09-09
Accept invalid SSL certificates on insecure registry pings.

merll:
When validation of the SSL certificate fails, the current implementation tries to connect via HTTP instead. However, a connect via unvalidated HTTPS should be attempted prior to using HTTP.
This PR accepts invalid SSL certificates, if `insecure_registry` in `pull` or `push` operations is set to `True`.

Possibly this is not the best workaround. The ping is performed from the client running the `docker-py`-based application; the actual connect however takes place from the remote-controlled client, which may have different certificate settings and therefore different standards of "secure" connections.


shin-:
*Cleaning up, sorry about the necro-posting*
Closing via #681

Issue #357, opened 2014-10-16
_socket.socket has no attribute _response


carlsverre:
Not sure if the latest release is 100% safe to deploy:

```
Exception in thread Thread-37:
Traceback (most recent call last):
  File "/usr/lib/python2.7/threading.py", line 551, in __bootstrap_inner
    self.run()
  File "/usr/lib/python2.7/threading.py", line 504, in run
    self.__target(*self.__args, **self.__kwargs)
  File "/plugins/psyduck/scheduler.py", line 190, in _exec
    for status in self.client.pull(image, stream=True, insecure_registry=True):
  File "/src/docker/docker/client.py", line 271, in _stream_helper
    socket_fp = socket_obj(_sock=self._get_raw_response_socket(response))
  File "/src/docker/docker/client.py", line 266, in _get_raw_response_socket
    sock._response = response
AttributeError: '_socket.socket' object has no attribute '_response'
```

PR #413, opened 2014-12-04
keep unix socket alive with python3

rmohr:
Hi,

I am currently working on p3 support for https://github.com/d11wtq/dockerpty. While working on this, I ran into strange socket errors.
It turned out, that the client.attach_socket method was broken because it always returned a closed socket when using python3. The problem is related to #357. The patch below keeps the socket open when using python3 or python2.
To achieve this, I am passing arount the SocketIO object instead of the raw socket.

What do you think?

Best Regards,
Roman

### DIFF
--- a/docker/client.py
+++ b/docker/client.py
@@ -274,18 +274,19 @@ class Client(requests.Session):
     def _get_raw_response_socket(self, response):
         self._raise_for_status(response)
         if six.PY3:
-            sock = response.raw._fp.fp.raw._sock
+            sock = response.raw._fp.fp.raw
+            sock._response = response
         else:
             sock = response.raw._fp.fp._sock
-        try:
-            # Keep a reference to the response to stop it being garbage
-            # collected. If the response is garbage collected, it will close
-            # TLS sockets.
-            sock._response = response
-        except AttributeError:
-            # UNIX sockets can't have attributes set on them, but that's fine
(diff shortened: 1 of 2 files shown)


docker/client.py
```diff
@@ -274,18 +274,19 @@ def _warn_deprecated(self, arg_name, version):
...
             sock = response.raw._fp.fp._sock
-        try:
-            # Keep a reference to the response to stop it being garbage
-            # collected. If the response is garbage collected, it will close
-            # TLS sockets.
-            sock._response = response
-        except AttributeError:
-            # UNIX sockets can't have attributes set on them, but that's fine
-            # because we won't be doing TLS over them
-            pass
```
shin-: Can you expand on those changes here? It seems to me that both versions would do exactly the same thing, am I mistaken?

shin-:
Just a small comment, otherwise LGTM

shin-:
And sorry for the delay here, haven't had much time for docker-py in the past few weeks.

shin-:
Thanks!

docker/client.py
rmohr: In Python2 the socket stays open anyway. But it would be cool to get rid of the try-except block. I will look into it.

rmohr:
:+1:

Issue #451, opened 2015-01-08
exception thrown when pulling repo from registry.hub.docker.com


kkleidal:
Example:  `client.pull('registry.hub.docker.com/my/repo')`
would raise docker.errors.InvalidRepository: Invalid repository name, try "my/repo" instead

The if body starting at line 62 in docker-py/docker/auth/auth.py is where this occurs.  It prohibits us from explicitly naming registry.hub.docker.com or index.docker.io as registries.  I assume this is because they are implied.

What is the rationale behind throwing an exception when the registry is stated explicitly?  The `docker pull` bash command allows us to explicitly name the registry even if it is implied.

shin-:
It used to be prevented in mainline docker. Indeed if that's not the case anymore I see no reason for this code to stay.

shin-:
[Actually still there](https://github.com/docker/docker/blob/master/registry/registry.go#L214), just doesn't prevent `registry.hub.docker.com` for some reason.

kkleidal:
Updated PR so that index.docker.io is still not allowed, but registry.hub.docker.com is.  #453

kkleidal:
PR #453 merged

PR #452, opened 2015-01-08
removed unnecessary registry host check as proposed in #451

kkleidal:
See #451 for reasoning.  The main point is that by not allowing index.docker.io and registry.hub.docker.com to be explicitly provided, client.pull is inconsistent with the docker command line client.


kkleidal:
Replaced by #453

PR #453, opened 2015-01-08
removed exception thrown if registry.hub.docker.com is explicitly define...

kkleidal:
...d in repository param for client.pull

Replaces #452 since index.docker.io is still not allowed to be explicitly provided when pulling with the docker client


shin-:
Thanks!

PR #455, opened 2015-01-11
Adds a fix for the bug where a fully-qualified URI given as the registry...

aurynn:
..., with trailing /v1, will be placed as-is in the auth_config directory, whereas subsequent lookups to auth_config will be performed with a registry URI that lacks the trailing /v1.

### DIFF
--- a/docker/auth/auth.py
+++ b/docker/auth/auth.py
@@ -22,7 +22,12 @@ import six
 from ..utils import utils
 from .. import errors
 
-import urlparse
+try:
+    import urlparse
+except ImportError:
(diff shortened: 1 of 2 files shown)


aurynn:
Provides new test cases directly for auth.auth.expand_registry_url.

shin-:
Thanks! IIUC that's the same thing as #468 and #463 ? If so we could still use the additional tests.

shin-:
Closing as this was fixed by another PR in 0.7.2. Thanks again for submitting a fix!

PR #462, opened 2015-01-16
fix create_host_config method

bernardopericacho:
I had an issue trying to start a stopped container changing it's links. If I start a container with links and then I stop and start it again setting this time `links=[]` to start it without links, before it doesn't change them. With this fix, it will set no links. This happens to other attributes also changed in the commit.


shin-:
Sensible. Thank you!

PR #463, opened 2015-01-19
Fix authconfig resolve for private registries

fermayo:
Due to d8eb04084b6631c1d2b9c864ddcd171a183b6025, private registry auth configs were not resolved. This commit fixes it.

### DIFF
--- a/docker/auth/auth.py
+++ b/docker/auth/auth.py
@@ -83,7 +83,7 @@ def resolve_authconfig(authconfig, registry=None):
     #
     # as there is only one auth entry which is fully qualified we need to start
     # parsing and matching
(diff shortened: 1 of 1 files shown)


shin-:
LGTM, thanks!

Would it make sense to add regression tests to prevent breaking it again in the future?

fermayo:
It makes a lot of sense, If I get the time!

shin-:
I know how it is. If I find some time before you do I'll take care of it :)

fermayo:
This is now a duplicate of #468

shin-:
Thanks, the fix and regression tests made it into 0.7.2 released today.

PR #468, opened 2015-01-29
Fix registry url canonicalization in resolve_authconfig

jamespharaoh:
This fixes a problem where URLs aren't canonicalised properly and so authentication isn't pulled out of the file correctly in some cases, causing a 403, even though the correct details are in the .dockercfg file. The problem was that the check whether the URL included the path element was too general and was fooled by the "/" in the method.


mpetazzoni:
Just ran into this, can confirm this is the root cause (and the proper fix).

PR #474, opened 2015-02-04
Fix crash when building from remote sources

gierschv:
Avoid crashing when the Docker python client is able to load
registry credentials when building using a remote source


shin-:
LGTM, thanks!

PR #475, opened 2015-02-05
Rename Client._timeout to timeout.

gtaylor:
As proposed in #284 , here's a PR that renames Client._timeout to be public. This initial PR does not make any attempt to alias ``_timeout`` to ``timeout``, but that's easily added if we want/need to do that.

I wasn't sure what the policy was on documenting these changes. I'd be glad to add that to the change log as well if someone lets me know what version number we're shooting for next.

### DIFF
--- a/docker/client.py
+++ b/docker/client.py
@@ -56,8 +56,8 @@ class Client(requests.Session):
                 )
             )
         self.base_url = base_url
+        self.timeout = timeout
         self._version = version
-        self._timeout = timeout
         self._auth_configs = auth.load_config()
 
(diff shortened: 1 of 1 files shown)


shin-:
Thanks!

Um, I think that's all fine. Regarding the Changelog, I usually compile it myself when doing a release, so don't worry about it :)

gtaylor:
I just realized that I renamed ``_timeout``, but now ``_version`` ends up being inconsistent in that it is also a kwarg and of general interest, yet private.

Would you accept another PR that renamed ``_version`` to ``version``?

PR #476, opened 2015-02-05
Update Client.build() docs to note rm default mis-match

### LINKED ISSUE #332: Client.build parameters do not match docker build parameters
Client.build has a default rm=False parameter whereas it is --rm=true for docker build.
Also, Client.build does not have a force-rm option. I don't use it myself, but I'm not sure if it got left out.

gtaylor:
I figured we probably wanted to include some context for the mis-match, so this ended up being a little bit of a long param docstring. Seemed worth explaining so people won't repeatedly bring this up in the issue tracker. Fixes #332 .


shin-:
LGTM, thanks! =)

PR #480, opened 2015-02-07
Add missing options to the events command

ggtools:
- Add since, until and filters parameters to `Client.events`
- Add missing `events`command in the documentation

Signed-off-by: Christophe Labouisse <christophe@labouisse.org>

### DIFF
--- a/docker/client.py
+++ b/docker/client.py
@@ -23,6 +23,8 @@ import requests
 import requests.exceptions
 import six
 
+from datetime import datetime
+
 from .auth import auth
 from .unixconn import unixconn
 from .ssladapter import ssladapter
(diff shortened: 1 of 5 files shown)


shin-:
Hi, thanks for contributing! This looks good, my only issue is that, if I understand correctly, this would break for people currently using the events endpoint because of the different output format (strings VS dicts). Can we maybe add the decoding on the fly as an option, and leave the default as is to avoid breaking?

Thanks!

ggtools:
Yes you're definitely right I adding this to the PR right now.

shin-:
Awesome, thanks! LGTM.

abhat:
@ggtools and @shin- , this is of huge help. Should the docs/api.md document a dictionary of non-unicode strings if decode is None and the current version if True? If you agree, I will make the edit so that the documentation is more complete.

PR #484, opened 2015-02-11
support for stats retrieval

nir0s:
This is a very initial implementation of the stats API.
I wanted to get the ball rolling on this implementation and get your input so that we can perfect this.

Documentation is in place and a stats aggregation feature was added.

### DIFF
--- a/docker/client.py
+++ b/docker/client.py
@@ -966,28 +966,11 @@ class Client(requests.Session):
         res = self._post_json(url, data=start_config)
         self._raise_for_status(res)
 
-    def stats(self, container, aggregate=0):
+    def stats(self, container):
         if isinstance(container, dict):
             container = container.get('Id')
         url = self._url("/containers/{0}/stats".format(container))
-        res = self._get(url, stream=True)
-        self._raise_for_status(res)
-
-        stats_stream = res.iter_lines()
-        stats_array = []
-        stats_aggregated = 0
-
-        if aggregate:
-            for stat_obj in stats_stream:
-                if stats_aggregated >= aggregate:
-                    stats_aggregated = 0
-                    yield stats_array
-                    stats_array = []
-                stats_array.append(stat_obj)
-                stats_aggregated += 1
-        else:
-            for stat_obj in stats_stream:
-                yield stat_obj
+        return self._stream_helper(self._get(url, stream=True))
 
     def stop(self, container, timeout=10):
         if isinstance(container, dict):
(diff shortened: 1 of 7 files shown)


nir0s:
Some comments.

- Need to add unit tests
- Need to add integration tests
- I have code in place to add a maximum number of stats messages to return in case you don't want to stream indefinitely. Thinking whether this is relevant or not.

docker/client.py
```diff
@@ -966,11 +966,35 @@ def start(self, container, binds=None, port_bindings=None, lxc_conf=None,
...
         self._raise_for_status(res)
 
+    def stats(self, container, aggregate=0):
+        if isinstance(container, dict):
+            container = container.get('Id')
+        url = self._url("/containers/{0}/stats".format(container))
+        res = self._get(url, stream=True)
+        self._raise_for_status(res)
+
+        stats_stream = res.iter_lines()
```
shin-: Should we use `_stream_helper` here instead?

docs/api.md
```diff
@@ -664,6 +664,32 @@ from. Optionally a single string joining container id's with commas
 None
 ```
 
+## stats
+
+The Docker API parallel to the `docker stats` command.
+This will stream statistics for a specific container.
+
+`aggregate`, if 0 will return an iterable stream object of messages.
```
shin-: I'm unsure what the use case is for this param, can you clarify what your thought process was?

shin-:
Thank you for jumpstarting this! I left a couple comments on the code.

> - Need to add unit tests
> - Need to add integration tests

Aye aye, let me know if I can help with those.

> - I have code in place to add a maximum number of stats messages to return in case you don't want to stream indefinitely. Thinking whether this is relevant or not.

Is this something that can't be achieved by:
```python
i = 0
for line in client.stats():
    if i >= MAX:
        break
    ...
```
?

docs/api.md
m1keil: I believe this is mechanism for "batching" stats. 
But I think it's a bit problematic since it's blocking and can actually get deadlock if stats stream stopped?

docker/client.py
nir0s: oops. my bad. will do.

nir0s:
I'll see if I get to the tests soon.. if not, i'd appreciate the help.
Regarding the max number of messages.. I was just wondering if it would be useful. Don't wanna add useless code :)

nir0s:
Well then.. I added a unittest and am now using stream helper.
I removed aggregation until I have a clearer view of how to implement this.
The reasoning behind it is that you might want to aggregate stats and return only then return them so that you're not blocked by the stream.

Obviously, my previous implementation was blocking. Comes to prove that you shouldn't work long hours :) makes your judgment fuzzy.

shin-:
Thank you!

PR #485, opened 2015-02-11
Support for PID mode

shin-:


dreamcat4:
Thanks @shin- !
Do we know yet for which minimum version of fig for this will be released ? Perhaps Docker 1.6 ?

PR #486, opened 2015-02-11
Fix old style string formatting

nir0s:
For consistency sake.. and because it's just the right thing to do :)


shin-:
LGTM, thanks!

PR #490, opened 2015-02-12
split tests to different tox envs

nir0s:
For consistency sake.. and because it's just the right thing to do :)


shin-:
Thanks, LGTM

PR #494, opened 2015-02-16
Fix several DOCKER_HOST assumptions in integration_test.py

tianon:
See https://github.com/docker/docker/pull/10797#issuecomment-74591463.

### DIFF
--- a/tests/integration_test.py
+++ b/tests/integration_test.py
@@ -1429,7 +1429,7 @@ class UnixconnTestCase(unittest.TestCase):
         with warnings.catch_warnings(record=True) as w:
             warnings.simplefilter('always')
 
-            client = docker.Client()
+            client = docker.Client(base_url=DEFAULT_BASE_URL)
             client.images()
             client.close()
             del client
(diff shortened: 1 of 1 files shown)


tests/integration_test.py
```diff
@@ -1444,7 +1444,7 @@ def test_resource_warnings(self):
 
 class TestRegressions(unittest.TestCase):
     def setUp(self):
-        self.client = docker.client.Client(timeout=5)
+        self.client = docker.Client(timeout=5, base_url=DEFAULT_BASE_URL)
```
tianon: Is there a particular reason this one has a custom `timeout` value?

shin-: Per https://github.com/tianon/docker-py/blob/fix-host-assumptions/tests/integration_test.py#L50, all of the timeouts are set to 5 - any longer probably indicates a problem with the test anyway. Did you have any particular issue with this, or just curiosity?

shin-:
LGTM!

tests/integration_test.py
tianon: Just curiosity! :)

PR #496, opened 2015-02-18
Add optional auth to pull

hibooboo2:
Goal here is to allow pull command to be made with onetime credentials so that a user may pull using different credentials without logging in and out.


hibooboo2:
@shin- Is there any chance you can take a look at this?

shin-:
Thank you!

hibooboo2:
Thank you shin.

PR #502, opened 2015-02-24
Add special cases for .dockerignore

### LINKED ISSUE #498: .dockerignore and Dockerfile ignore
Docker 1.5 introduced the ability to ignore both the .dockerignore and the Dockerfile (in the .dockerignore file), but this is processed on the server since and only for dealing with the build cache (ie, when someone does ADD., Dockerfile changes can be ignored).

If Dockerfile is in .dockerignore, Dockerfile won't be sent to the server, so the server can't build it

Basically, .dockerignore, and Dockerfile should be treated as special case entries in the .dockerignore so that they still get sent up to the server.

cpuguy83:
Fixes #498
Allowing `Dockerfile` and `.dockerignore` in the exclusion filter
completely breaks the build on docker < 1.5

In Docker 1.5 these entries are treated as special cases when included
in the .dockerignore and are still sent as part of the context. The
daemon ends up excluding them from any `ADD`, `COPY`, and cache
validation.

### DIFF
--- a/docker/client.py
+++ b/docker/client.py
@@ -297,6 +297,12 @@ class Client(requests.Session):
             if os.path.exists(dockerignore):
                 with open(dockerignore, 'r') as f:
                     exclude = list(filter(bool, f.read().split('\n')))
+                    # These are handled by the docker daemon and should not be
+                    # excluded on the client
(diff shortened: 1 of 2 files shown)


shin-:
Thanks!

PR #504, opened 2015-02-24
BUG: Fix errors in Python 3 from JSON-decoding bytes.

ssanderson:
Without this, passing `decode=True` to any methods that delegate to `stream_helper` results in `TypeError` on Python 3.


shin-:
Good catch! Thanks! :)

PR #505, opened 2015-02-25
detect server API version automatically

### LINKED ISSUE #503: RFE: try to figure out version automatically
### Story
I have a docker daemon with version "1.16", my docker-py client has default version "1.17". As soon as I try to do some api calls, it fails with:
```
APIError: 404 Client Error: Not Found ("client and server don't have same version (client : 1.17, server: 1.16)")
```
Nothing new.

### Solution
I would prefer if docker-py could try to change the version so it would match server's api version, if possible.

### Workflow

1. `d = docker.Client()`
2. `d.images()`
3. Exception is thrown
4. docker-py retries request with different version (the one from exception)
 * if it succeeds, warn user that version should be changed explicitely
 * if it fails, it fails

What do you think? If it makes sense, I can create a PR.

TomasTomecek:
Fixes #503 

The reason I've implemented it this way is that Client is **not** making any requests in constructor and I think that this PR should *not* change that behavior. Tried to benchmark how much slower will be the first request and I think that 30-50 ms is not that bad:

```
$ time docker version >/dev/null
docker version > /dev/null  0.02s user 0.02s system 85% cpu 0.047 total
$ time docker version >/dev/null
docker version > /dev/null  0.02s user 0.01s system 106% cpu 0.030 total
$ time docker version >/dev/null
docker version > /dev/null  0.02s user 0.02s system 107% cpu 0.032 total
```

### DIFF
--- a/docker/__init__.py
+++ b/docker/__init__.py
@@ -17,4 +17,4 @@ from .version import version
 __version__ = version
 __title__ = 'docker-py'
 
-from .client import Client # flake8: noqa
+from .client import Client, AutoVersionClient # flake8: noqa
--- a/docker/client.py
+++ b/docker/client.py
@@ -40,7 +40,7 @@ STREAM_HEADER_SIZE_BYTES = 8
 
 
 class Client(requests.Session):
-    def __init__(self, base_url=None, version=DEFAULT_DOCKER_API_VERSION,
+    def __init__(self, base_url=None, version=None,
                  timeout=DEFAULT_TIMEOUT_SECONDS, tls=False):
         super(Client, self).__init__()
         base_url = utils.parse_host(base_url)
@@ -50,15 +50,8 @@ class Client(requests.Session):
             raise errors.TLSParameterError(
                 'If using TLS, the base_url argument must begin with '
                 '"https://".')
-        if not isinstance(version, six.string_types):
-            raise errors.DockerException(
-                'version parameter must be a string. Found {0}'.format(
-                    type(version).__name__
-                )
-            )
         self.base_url = base_url
         self.timeout = timeout
-        self._version = version
         self._auth_configs = auth.load_config()
 
         # Use SSLAdapter for the ability to specify SSL version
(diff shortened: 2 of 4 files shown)


TomasTomecek:
rebased, fixed flake8 errors

shin-:
I'll take some time to look into this today, thank you for your contribution!

shin-:
Looks good from a code standpoint. I have a couple issues with the design, mainly with the idea of having the version be fetched on the first request. I feel like we're hiding what really happens behind the scenes and if issues happen during that phase it will confuse the hell out of users.

Additionally, I'm not sure this should be the default behavior (mainly because I dislike the idea of making that extra request in the general case).

What do you think of the following plan:

1. Add a special value "auto" to the valid `version` values, which will trigger this behavior
2. If `version=None`, use the `DEFAULT_DOCKER_API_VERSION` constant.
3. If `version="auto"`, fetch the API version from the daemon inside the `Client.__init__` method.

Alternatively, if you think having a request inside the `__init__` method is a bad idea, we could have an additional Client method called `retrieve_server_version()` or something similar that fetches the API version and sets the `Client._version` attribute. It's a bit extra work for the user but at least there's no implicit requests happening behind the scenes.

TomasTomecek:
Makes total sense to me. I'm in favor of the `auto` keyword — it's short and super descriptive.

My primary use case is that when I want to try something out with docker-py, I'll open ipython and do:

```python
In [1]: import docker

In [2]: d = docker.Client()

In [3]: d.images()
---------------------------------------------------------------------------
APIError                                  Traceback (most recent call last)
<ipython-input-4-6e3df1620d79> in <module>()
----> 1 d.images()
        global d.images = <bound method Client.images of <docker.client.Client object at 0x7f9a4a2fa510>>

/usr/lib/python2.7/site-packages/docker/client.pyc in images(self=<docker.client.Client object>, name=None, quiet=False, all=False, viz=False, filters=None)
    647             params['filters'] = utils.convert_filters(filters)
    648         res = self._result(self._get(self._url("/images/json"), params=params),
--> 649                            True)
        global True = undefined
    650         if quiet:
    651             return [x['Id'] for x in res]

/usr/lib/python2.7/site-packages/docker/client.pyc in _result(self=<docker.client.Client object>, response=<Response [404]>, json=True, binary=False)
     96     def _result(self, response, json=False, binary=False):
     97         assert not (json and binary)
---> 98         self._raise_for_status(response)
        self._raise_for_status = <bound method Client._raise_for_status of <docker.client.Client object at 0x7f9a4a2fa510>>
        response = <Response [404]>
     99 
    100         if json:

/usr/lib/python2.7/site-packages/docker/client.pyc in _raise_for_status(self=<docker.client.Client object>, response=<Response [404]>, explanation=None)
     92             response.raise_for_status()
     93         except requests.exceptions.HTTPError as e:
---> 94             raise errors.APIError(e, response, explanation=explanation)
        global errors.APIError = <class 'docker.errors.APIError'>
        e = HTTPError('404 Client Error: Not Found',)
        response = <Response [404]>
        explanation = None
     95 
     96     def _result(self, response, json=False, binary=False):

APIError: 404 Client Error: Not Found ("client and server don't have same version (client : 1.18, server: 1.17)")
> /usr/lib/python2.7/site-packages/docker/client.py(94)_raise_for_status()
     93         except requests.exceptions.HTTPError as e:
---> 94             raise errors.APIError(e, response, explanation=explanation)
     95 

ipdb> 
```

And now I have to create client again. Therefore, how about if we created new subclass of client with automatic version detection?

> If version=None, use the DEFAULT_DOCKER_API_VERSION constant.

I totally agree with not changing the current default behavior. We certainly don't want to break existing tooling/scripts written on top of docker-py.

TomasTomecek:
Rewritten, fixed flake errors.

shin-:
Merged, thanks a lot for your help!

TomasTomecek:
Happy to help.

bobrik:
Is there a reason not to set `DEFAULT_DOCKER_API_VERSION` to `auto`?

PR #507, opened 2015-02-26
Update resolve_authconfig to bring it in line with Docker client

aanand:
Instead of expanding the registry name we're looking for to a full URL, strip entries in the authconfig down to just the hostname.

This mirrors how the Docker client does it - see [ResolveAuthConfig in auth.go](https://github.com/docker/docker/blob/8b95ad230e2ee76450ceb9a80aa8e942a56bb397/registry/auth.go#L451-L481) - and should fix https://github.com/docker/compose/issues/75.

### DIFF
--- a/docker/auth/auth.py
+++ b/docker/auth/auth.py
@@ -26,14 +26,6 @@ INDEX_URL = 'https://index.docker.io/v1/'
 DOCKER_CONFIG_FILENAME = '.dockercfg'
 
 
-def swap_protocol(url):
-    if url.startswith('http://'):
-        return url.replace('http://', 'https://', 1)
-    if url.startswith('https://'):
-        return url.replace('https://', 'http://', 1)
-    return url
-
-
 def expand_registry_url(hostname, insecure=False):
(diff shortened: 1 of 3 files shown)


docker/auth/auth.py
```diff
@@ -68,29 +60,29 @@ def resolve_repository_name(repo_name, insecure=False):
...
+        if convert_to_hostname(key) == registry:
+            return config
+
+    return None
+
+
+def convert_to_hostname(url):
+    url = url.replace('http://', '')
+    url = url.replace('https://', '')
+    return url.split('/', 1)[0]
```
dnephin: maybe use  https://docs.python.org/2/library/urlparse.html#urlparse.urlparse instead ? It should to be robust to unexpected parts of the url.

It's available as `six.moves.urllib.parse.urlparse`

aanand: Arguably better, yes, but I wanted to stick as close to the Docker client's behaviour. Not sure what's the stronger principle here!

aanand:
Fixed the tests.

shin-:
Thanks Aanand! Merged :)

PR #508, opened 2015-02-27
Rename test fix to take into account docker/docker#11003

shin-:


tests/integration_test.py
```diff
@@ -358,7 +359,10 @@ def runTest(self):
         self.client.rename(res, name)
         inspect = self.client.inspect_container(res['Id'])
         self.assertIn('Name', inspect)
-        self.assertEqual(name, inspect['Name'])
+        if version == '1.5.0':
```
TomasTomecek: would it make sense to use [compare_version](https://github.com/docker/docker-py/blob/master/docker/utils/utils.py#L99) here? so when 1.6 lands, one does not need to change the test again.

shin-: We won't need to change the tests for 1.6? `compare_version` doesn't work well with the `-dev` suffix, and we want `1.5.0-dev` (and any version after that) to take the `else` path here.

PR #518, opened 2015-03-06
When building, include submoduled .git files as .git subdirectory trees

ianblenke:
Submodules projects with a fig.yml/docker-compose.yml have a build: directive that includes a subdirectory path. If that subdirectory has a .git file in it (that isn't otherwise excluded by a .dockerignore) and check if it starts with gitdir:. If so: include that relative path as a relatively placed .git directory in the tarball that is sent to the docker server.

### DIFF
--- a/docker/utils/utils.py
+++ b/docker/utils/utils.py
@@ -86,7 +86,18 @@ def tar(path, exclude=None):
         dirnames.sort()
         for name in sorted(fnames):
             arcname = os.path.join(relpath, name)
-            t.add(os.path.join(path, arcname), arcname=arcname)
+            fullname = os.path.join(path, arcname)
+            # Deal with .git file/folder for nested submodules
+            if (  os.path.basename(fullname) == '.git' and
+                  os.path.isfile(fullname) ):
+                with open(fullname, 'r') as git_file:
+                    content = git_file.read()
+                    if content.startswith('gitdir: '):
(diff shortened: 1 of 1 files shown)


docker/utils/utils.py
```diff
@@ -86,7 +86,18 @@ def tar(path, exclude=None):
...
+            # Deal with .git file/folder for nested submodules
+            if (  os.path.basename(fullname) == '.git' and
+                  os.path.isfile(fullname) ):
+                with open(fullname, 'r') as git_file:
+                    content = git_file.read()
+                    if content.startswith('gitdir: '):
+                        rellinkpath = string.split(content, 'gitdir: ').pop
+                        linkpath = os.path.join(relpath, rellinkpath)
+                        t.add(os.path.join(path, linkpath), arcname=arcname)
+            else
```
shin-: Missing a colon here...

ianblenke: Heh, yep, the tests caught it. Looks like I also did the split wrong, and have some whitespace before/after parenthesis that the tests are complaining about. Trying to fix those now as well. Thanks for the catch!

dnephin:
Thanks for the contribution! Unfortunately I don't think this is something we want to support.

We'd like `.dockerignore` to work identically to the docker cli implementation, and I don't think it has this special case for `.git`.

PR #520, opened 2015-03-09
Making examples in doc a bit clearer for volumes and ports_binding

kenden:
The examples refer to a variable container_id, but it had not been introduced before.


shin-:
Thanks!

PR #522, opened 2015-03-10
Documentation: removing getting response from docker.client.start()

kenden:
It is confusing that the documentation example gets and prints a response from docker.client.start()
That function returns nothing.
I just removed the part where a response variable is set to the return value of the function.

### DIFF
--- a/docs/volumes.md
+++ b/docs/volumes.md
@@ -4,7 +4,7 @@ Volume declaration is done in two parts. First, you have to provide
 a list of mountpoints to the `Client().create_container()` method.
 
 ```python
-c.create_container('busybox', 'ls', volumes=['/mnt/vol1', '/mnt/vol2'])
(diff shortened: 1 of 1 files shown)


TomasTomecek:
It's not nothing. It's `None`. I think it's pretty important to state that `start` returns `None`. If it's not there, everyone will be curious what `start` returns (and will think that `start` actually returns something more meaningful).

shin-:
Thank you for sharing your concerns! The example was written like this to outline the fact that there is no return value to `client.start`. There might be a better way to get the point across, but removing information definitely isn't the way to go here.

kenden:
Thanks Tomas and shin.

I will create a new pull request adding information about what it returned.
(I did a rebase on am fork, and this pull request got closed).

But I disagree: 
Having an example putting the return to the function in a variable made me believe that there could be a return to the function that could be 'not None'. 
I had to look at the code to check that the function has no 'return' statement.

PR #527, opened 2015-03-16
BUG: Don't barf on Client.port when no port settings provided.

### LINKED ISSUE #526: `Client.port` fails when called on a container with `network_mode=host`
This happens because the response JSON has `None` for ports rather than an empty dictionary, which results in calling `None.get(portname)`, which barfs.

ssanderson:
Fixes #526 .


shin-:
Thanks!

PR #531, opened 2015-03-20
Windows : fix EOL of .dockerignore file

josephpage:
To be able to read `.dockerignore` file edited on Windows, python-py may use the multi-platform split function `str.splitlines()`

Ref : https://docs.python.org/2/library/stdtypes.html#str.splitlines

(tested rudimentarily only)


shin-:
LGTM, thanks!

PR #533, opened 2015-03-21
Fixed stats related typo in api.md

ynachiket:
The new stats api related doc example uses stats instead of stats_obj for looping.
This PR fixes the small typo.


shin-:
Thanks! Nice catch.

PR #537, opened 2015-03-30
IPC sharing implementation

onorua:
it gives possibility to share IPC between containers and use hosts based IPC, which was already for some time available via docker cli tools.


shin-:
Thanks! =]

PR #540, opened 2015-04-03
Add a version_info attribute

terminalmage:
This makes comparing docker-py versions *much* easier. Having to account for
the potential of a "-dev" suffix is a PITA.


shin-:
Thanks!

PR #544, opened 2015-04-11
parse port ranges

uvgroovy:
Added parsing port range utility.
to assist with:
https://github.com/docker/compose/issues/1102
see:
https://github.com/docker/compose/pull/1191#discussion_r27248067
for more details.

### DIFF
--- a/docker/utils/ports/__init__.py
+++ b/docker/utils/ports/__init__.py
@@ -1,0 +1,6 @@
+from .ports import (
+    to_port_range,
+    split_port,
+    add_port_mapping,
+    build_port_bindings
+) # flake8: noqa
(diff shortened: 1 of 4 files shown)


docker/utils/ports/__init__.py
```diff
@@ -0,0 +1,6 @@
+from .ports import (
+    to_port_range,
+    split_port,
+    add_port_mapping,
+    build_port_bindings
```
aanand: I don't think `to_port_range` and `add_port_mapping` should be imported here - they're only used within `ports.py`, and I can't see them being useful outside it

aanand:
Looks good, apart from minor API surface area issue (comment inline).

uvgroovy:
@aanand  - thanks for the feedback - fixed

shin-:
LGTM, thanks for taking the time to do this!

PR #548, opened 2015-04-21
add MAINTAINERS

proppy:
So we can add more later :)

### DIFF
--- a/MAINTAINERS
+++ b/MAINTAINERS
@@ -1,0 +1,1 @@
+Joffrey F <f.joffrey@gmail.com> (@shin-)


shin-:
Should we add @mpetazzoni and maybe @aanand in there? (I know we're kind of in maintainer-limbo, but if we're going to merge this officially let's take the chance to straighten it out.)

mpetazzoni:
FWIW I'm still watching over what's going on here with a lot of attention and can definitely jump in whenever I'm needed. The only thing I can't do I think is push releases to PiPy.

proppy:
@shin- they can always add themselves in follow up PRs, I'd rather not volunteer them :)

Can we go ahead and merge this?

PR #557, opened 2015-04-23
Add /exec/{id}/info support

christopherobin:
Adds a new boolean parameter `exec_id` to `execute` to return the execution id as a second value.
That value can then be fed to `execute_inspect` to get informations about the execution (running status, exit code, etc...).

Usage:

```python
out, exec_id = c.execute(container, ['ls', '/foobar'], exec_id=True)
info = c.execute_inspect(exec_id)
if info.get('ExitCode') == 2:
    raise IOError('Folder /foobar does not exists')
```

### DIFF
--- a/docker/client.py
+++ b/docker/client.py
@@ -505,7 +505,7 @@ class Client(requests.Session):
                                    decode=decode)
 
     def execute(self, container, cmd, detach=False, stdout=True, stderr=True,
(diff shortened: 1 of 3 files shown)


MiLk:
:+1:

shin-:
Thanks! I actually took a different approach in #568, but end result is essentially the same.

PR #563, opened 2015-04-24
Logconfig support

### LINKED ISSUE #547: Add support for logging drivers.
Docker 1.6 has added a log driver option to the host config.

>Logging Drivers
There have been a growing number of proposals for a logging driver API that would allow you to send container logs to other systems such as Syslog or a third-party. This new logging driver follows the exec driver and storage driver concepts already available in Engine today.
There is a new option to docker run,  --log-driver, which has three options:json-file (the default, and same as the old functionality), syslog, and none.

https://blog.docker.com/2015/04/docker-release-1-6/
https://docs.docker.com/reference/api/docker_remote_api_v1.18/#create-a-container

This should be exposed in client.create_container

shin-:
Resolves #547

### DIFF
--- a/docker/utils/__init__.py
+++ b/docker/utils/__init__.py
@@ -5,4 +5,4 @@ from .utils import (
     create_container_config, parse_bytes, ping_registry
 ) # flake8: noqa
 
-from .types import Ulimit # flake8: noqa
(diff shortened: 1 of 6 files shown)


cmoro-deusto:
@shin- you are the man! Thanks!

PR #566, opened 2015-04-26
fix(client.py): unification the container argument for commit

pahaz:
Example:

    container = _client.create_container(..)
    _client.start(container)  # correct!
    _client.attach(container)  # correct!
    _client.wait(container)  # correct !
    _client.commit(container)  # INCORRECT!


shin-:
Thanks! =)

PR #569, opened 2015-04-27
Improved filters documentation

### LINKED ISSUE #416: Documentation on filters and name needs update
The docs for `containers` and `images` could use an entry for the `filters` parameter, which appears to be passed a dictionary that corresponds to the filters available via the docker command line

Also, the `name` parameter for `images(...)` could use some clarification that it is an exact search only e.g. if the image name+tag is `foo/bar:12.04`, then only the string `foo/bar` will return this container, and not any substrings e.g. `foo`, `bar`, or `foo/bar:12.04`

shin-:
Fixes #416

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -135,6 +135,10 @@ non-running ones
 non-running ones
 * limit (int): Show `limit` last created containers, include non-running ones
 * size (bool): Display sizes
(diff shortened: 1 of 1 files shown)


dnephin:
I think this is missing `label`, which is also missing from the docker remote API docs. I have https://github.com/docker/docker/pull/12798 open to try and get it added.

PR #576, opened 2015-04-29
exec_resize is incorrectly passing param as json

ctrlhxj:
as title. change to the same way we did for container resize


ctrlhxj:
Could some one help to review and possibly merge this pull requests. This is a bug fix. Also tested in our prod env as well

shin-:
Nice catch, thanks!

PR #577, opened 2015-04-29
Update client.py

pmoosh:
I think this to be a bug. That can prevent auth with private registries. Tested change in client.py with our private registry and modified code works


shin-:
Indeed. Thanks!

PR #584, opened 2015-05-06
add privilege for exec_create

tonicbupt:
privileged is always `False` in `exec_create` currently.

### DIFF
--- a/docker/client.py
+++ b/docker/client.py
@@ -531,7 +531,8 @@ class Client(requests.Session):
 
         return self.exec_start(create_res, detach, tty, stream)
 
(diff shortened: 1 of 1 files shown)


shin-:
Thanks for contributing!

Does that work? [It's not documented...](http://docs.docker.com/reference/api/docker_remote_api_v1.18/#exec-create) Also, can you look into fixing the pep8 issue (the declaration line spans over 80+ characters)

tonicbupt:
hmm, actually someone has contributed it into master of docker ... see [this PR](https://github.com/docker/docker/pull/12025), I did the advanced thing... so maybe we just let it hang up here and wait for the release of this feature and then merge it... XD

shin-:
Thanks!

Issue #538, opened 2015-03-30
Design proposal: make the docker-py interface more 'data oriented'


moysesb:
TL;DR Make `client.build()` feel more like `client.create_container()`, and then make `create_container()`feel more like the new `build()`.

In its current form, the api exposed by docker-py is a direct one to one mapping over the names of the operations supported by the docker remote API. Since these names refer to operations that can be parametrized with query params, it would be nice if this parametrization were more explicit and dictionary-ish, so that it could be constructed independently of the call to the docker-py method that
implements the operation. 

For example, say I want to send a docker context to the daemon to build. The daemon accepts a certain number of options that control both the interpretation of the contents being sent and the properties of the image that will be generated from it. These options are formally  documented in the Docker Remote API docs, so in principle it's safe to declare them as some kind of version aware data structure in docker-py.

Right now, docker-py does not offer a way for a client application to declare what these options are, except implicitly in the call to `client.build()`. The build options are the named parameters for the `build()` method, which when called looks at the specific combination of arguments that was passed to decide how to construct the request to the daemon.

The proposal is to introduce a way for docker-py to construct that specific combination of parameters in a more explicit way, i.e., as a first class value that can be generated and passed around. 

So, keeping the example of a 'build' operation, suppose I want to send a directory containing a Dockerfile to the daemon. The way it works now is:
```python
    client = Client(**params)
    client.build(path='my/path', tag='some_image', 
    		rm=True, nocache=False, stream=True)
```
The signature for build is
```python
def build(self, path=None, tag=None, quiet=False, fileobj=None,
 		nocache=False, rm=False, stream=False, timeout=None,
		custom_context=False, encoding=None, pull=True,
		forcerm=False, dockerfile=None)
```
One obvious problem is that here are three types of named arguments declared there: 
	 1) some control how the final context object would be built locally.
	 2) some control how the docker daemon will execute the actual build.
	 3) some specify the mechanics of the interaction with the server 	    (timeout, stream).

It would be much better if:
   a) the arguments groups where formally separated so that the distinction is made declaratively.
   b) user code can create each of these groups of parameters independently of the call to `build()` in an "API safe" way.

A sketch implementation:
```python
  # This is the parametrization of the remote build job 
  BuildSpec spec = BuildSpec(api_version='1.17')
  spec['tag'] = 'my_image'
  spec['nocache'] = True
  spec['forcerm'] = False
  spec['dockerfile'] = './mydockerfile'
```
[...]

PR #589, opened 2015-05-07
Add support for building contexts objects

moysesb:
This PR implements #538 and is necessary for docker-compose PR https://github.com/docker/compose/pull/1244


docker/utils/context.py
```diff
@@ -0,0 +1,206 @@
...
+        pass
+
+
+class ContextError(Exception):
+    def __init__(self, msg):
+        self.message = msg
+
+
+INVALID_CONTEXT_FORMAT_LONG_FMT = """
+Build context at %s is not supported by docker-compose\\n
```
aanand: This should probably just say "Docker" :smirk:

docker/utils/context.py
```diff
@@ -0,0 +1,206 @@
...
+REMOTE_CONTEXT_PREFIXES = ["http://",
+                           "https://",
+                           "git://",
+                           "git@",
+                           "github.com/"]
+
+DOCKERFILE_CMD_RG = re.compile(r"[\t\s]*\\n$|([\s\t]*"
+                               r"FROM|MAINTAINER|RUN|CMD|EXPOSE|ENV|ADD|COPY|"
+                               r"ENTRYPOINT|VOLUME|USER|WORKDIR|ONBUILD"
+                               r"[\s\t]+.+)")
```
aanand: What's this constant used for?

moysesb: Removed now. It would eventually be used for validating a Dockerfile thoroughly, but I discarded the idea.

aanand:
1. How much of this logic is actually implemented in the Docker client? For example, validation of the Dockerfile seems to happen on the server side. We don't need to go beyond what the Docker client implements.

2. Documentation would be great.

moysesb:
@aanand 
1.  Scaled back the validation to be just on access rights and Dockerfile name check. The docker go client does roughly that level of validation, but since it supports reading a compressed context from stdin it implicitly rejects a malformed tarball.
2.  Added documentation at [docs/context.md] (https://github.com/moysesb/docker-py/blob/distinct_build_calls/docs/context.md)

aanand:
Thanks. Looking good.

PR #591, opened 2015-05-08
Added proper support for .dockerignore 'folder/' syntax

### LINKED ISSUE #581: .dockerignore ignores trailing slash on directories

```python
>>> fnmatch('src', 'src/')
False
```

shin-:
Fixes #581

### DIFF
--- a/docker/utils/utils.py
+++ b/docker/utils/utils.py
@@ -68,6 +68,35 @@ def fnmatch_any(relpath, patterns):
     return any([fnmatch(relpath, pattern) for pattern in patterns])
 
 
+def _advanced_exclude(exclude):
+    result = {
+        'folders': [],
+        'files': []
+    }
+    for ex in exclude:
+        if ex[-1] == '/':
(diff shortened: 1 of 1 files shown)


shin-:
@mikexstudios would you like to take a look at this before I merge it? Thanks.

mikexstudios:
Hey @shin-. Thanks for the fix and sorry for the late reply. I will test your PR tomorrow and let you know.

mikexstudios:
Sorry, this is taking longer than expected since I want to test your PR rigorously. I'm writing a bunch of test cases, which should help.

mikexstudios:
Hey @shin-. I updated your PR with a unit test for this case. Your PR works for detecting trailing slashes on folders.

However, I am concerned that we need a more general and robust way of handling .dockerignore files. The currently implementation fails on a lot of syntax defined here: https://github.com/docker/docker/blob/master/docs/sources/reference/builder.md#dockerignore-file

I'll see if I can contribute to improving this in docker-py.

PR #594, opened 2015-05-10
Reference the socket in python3

delfick:
It would appear in python3.4 that accessing _sock here is necessary. Otherwise it hangs and produces weird content.


shin-:
Thanks for contributing!

I'm really uncomfortable with touching that part of the code. Do you happen to have a code snippet or integration test that would allow us to reproduce this and test the regression?

delfick:
Fair enough.

I finally found some time to look into this. I'm struggling to write a test that breaks without adding _sock but my harpoon tests are still failing without adding the _sock, so I'll investigate some more.

delfick:
I believe the difference is connecting to the docker daemon over https as opposed to http+docker.

delfick:
hi @shin, I've force pushed to this branch with the following changes:

* A function to encapsulate making the docker context in the integration tests
* A test that only fails when using python3 and connecting to the docker daemon with tls
* The fix for that failing test
* A task in the makefile that starts a docker-machine, activates it and runs the integration tests against it so you can switch between a virtualenv on your machine using python2 or python3 and then test connecting to the docker daemon over tls with it.

:)

delfick:
@shin, hello, I was wondering if you had a chance to look at this pull request since my changes?

Thanks

delfick:
I force pushed again after applying the change on top of the latest master.

delfick:
@shin, I've rebased and force pushed again. Now it's four commits:
1. adding integration-dind-ssl make task
2. Make it more obvious why integration test pauses for so long at the start (pulling busybox)
3. adding a test that fails
4. fixing the test

Without the fixing commit it will fail in the integration-dind-ssl Python3 run and nowhere else.

shin-:
Thanks, and sorry for the delayed response. I'm taking a look now to make sure I understand correctly what's happening.

Dockerfile-dind-certs
```diff
@@ -0,0 +1,20 @@
+FROM python:2.7
```
shin-: Since this is exclusively for testing purposes, I'd rather see this file moved to `tests/`

Makefile
```diff
@@ -22,13 +25,22 @@ unit-test-py3: build-py3
...
+	docker run -d --name dpy-dind -v /tmp --env="DOCKER_HOST=tcp://localhost:2375" --privileged dockerswarm/dind:1.8.1 docker -d -H tcp://0.0.0.0:2375
+	docker exec dpy-dind docker pull busybox
+	docker run --volumes-from dpy-dind --env="DOCKER_HOST=tcp://docker:2375" --link=dpy-dind:docker docker-py py.test -rx -s tests/integration_test.py
+	docker run --volumes-from dpy-dind --env="DOCKER_HOST=tcp://docker:2375" --link=dpy-dind:docker docker-py3 py.test -rx -s tests/integration_test.py
 	docker rm -vf dpy-dind
+
+integration-dind-ssl: build-dind-certs build build-py3
+	docker run -d --name dpy-dind-certs dpy-dind-certs
+	docker run -d --env="DOCKER_HOST=tcp://localhost:2375" --env="DOCKER_TLS_VERIFY=1" --env="DOCKER_CERT_PATH=/certs" --volumes-from dpy-dind-certs --name dpy-dind-ssl -v /tmp --privileged dockerswarm/dind:1.8.1 docker daemon --tlsverify --tlscacert=/certs/ca.pem --tlscert=/certs/server-cert.pem --tlskey=/certs/server-key.pem -H tcp://0.0.0.0:2375
+	docker exec dpy-dind-ssl docker pull busybox
```
shin-: This actually hides a race condition where the client container isn't able to connect to the server if it's started right away. A bit tricky to resolve.

docker/client.py
```diff
@@ -239,7 +241,7 @@ def _multiplexed_response_stream_helper(self, response):
         # Disable timeout on the underlying socket to prevent
         # Read timed out(s) for long running processes
         socket = self._get_raw_response_socket(response)
-        if six.PY3:
+        if six.PY3 and hasattr(socket, "_sock"):
```
shin-: Shouldn't we use the same test in both cases? At first glance it seems checking for existence of the `_sock` attribute would be more robust than basing it upon the value of a publicly accessible high-level attribute like `Client.base_url`.

What was the intent behind making those different?

Dockerfile-dind-certs
delfick: fair enough, I'll move it

Makefile
delfick: It'll still pull down busybox in the tests if this fails...

docker/client.py
delfick: I did this because at this point all that matters is whether it has _sock or not, rather than what causes us to choose _sock inside _get_raw_response_socket.

It's possible though that we don't need the six.PY3 check in this case. I'll have a look at the behavior in Python2 later on today.

Makefile
shin-: Understood - I meant if you remove that exec line, tests will all error out with "Connection refused" when first trying to contact the server.

delfick: ahhh, ok

docker/client.py
delfick: Alternatively, what we could do is settimeout(None) on both the socket and socket._sock.

delfick: Nup, in python3 the socket is SocketIO object and has no settimout.

I'll instead go for

        if hasattr(socket, "settimeout"):
            socket.settimeout(None)
        if hasattr(socket, "_sock") and hasattr(socket._sock, "settimeout"):
            socket._sock.settimeout(None)

I think it's reasonable to expect one of them to always have settimeout and that setting it on both shouldn't have bad effects.

This way we also aren't tying this piece of code to whatever conditions are making this distinction necessary.

shin-: Sorry, to clarify: I'm fine with the way it works now, I'm just wondering about the difference between this line and [that one](https://github.com/docker/docker-py/pull/594/files#diff-5262437d121a30fc85a81c4b1da57e4bR186). I like testing for the existence of `_sock` better than testing the format of `base_url`.

delfick: Fair enough, The problem is that in Python3 _sock still exists even if ssl isn't used, but if we use _sock in that condition then it breaks. Which means we can only use _sock if it's using ssl.

delfick:
meanwhile, I've rebased ontop of origin/master again.

docker/client.py
shin-: Okay, gotcha. This is all sorts of awful...

delfick: yeap, just a bit

Makefile
```diff
@@ -26,10 +29,19 @@ integration-test-py3: build-py3
 	docker run -v /var/run/docker.sock:/var/run/docker.sock docker-py3 py.test -rxs tests/integration_test.py
 
 integration-dind: build build-py3
-	docker run -d --name dpy-dind --privileged dockerswarm/dind:1.8.1 docker -d -H tcp://0.0.0.0:2375
-	docker run --env="DOCKER_HOST=tcp://docker:2375" --link=dpy-dind:docker docker-py py.test -rxs tests/integration_test.py
-	docker run --env="DOCKER_HOST=tcp://docker:2375" --link=dpy-dind:docker docker-py3 py.test -rxs tests/integration_test.py
+	docker run -d --name dpy-dind --env="DOCKER_HOST=tcp://localhost:2375" --privileged dockerswarm/dind:1.8.1 docker -d -H tcp://0.0.0.0:2375
+	docker exec dpy-dind docker pull busybox:buildroot-2014.02
```
shin-: It's going to be tedious to keep track of this if we ever want to change the image we use for testing. This doesn't seem essential, can we get rid of it?

Makefile
```diff
@@ -26,10 +29,19 @@ integration-test-py3: build-py3
...
+	docker run --volumes-from dpy-dind --env="DOCKER_HOST=tcp://docker:2375" --link=dpy-dind:docker docker-py py.test -rx -s tests/integration_test.py
+	docker run --volumes-from dpy-dind --env="DOCKER_HOST=tcp://docker:2375" --link=dpy-dind:docker docker-py3 py.test -rx -s tests/integration_test.py
 	docker rm -vf dpy-dind
 
+integration-dind-ssl: build-dind-certs build build-py3
+	docker run -d --name dpy-dind-certs dpy-dind-certs
+	docker run -d --env="DOCKER_HOST=tcp://localhost:2375" --env="DOCKER_TLS_VERIFY=1" --env="DOCKER_CERT_PATH=/certs" --volumes-from dpy-dind-certs --name dpy-dind-ssl -v /tmp --privileged dockerswarm/dind:1.8.1 docker daemon --tlsverify --tlscacert=/certs/ca.pem --tlscert=/certs/server-cert.pem --tlskey=/certs/server-key.pem -H tcp://0.0.0.0:2375
+	docker exec dpy-dind-ssl docker pull busybox:buildroot-2014.02
```
shin-: ditto above

shin-:
One last small nitpick, otherwise LGTM. Thanks for your help and patience through this!

Makefile
delfick: agreed.

delfick:
No probs.

Thanks for taking the time to review :)

tests/integration_test.py
```diff
@@ -74,8 +76,15 @@ def setup_module():
...
-        c.inspect_image(BUSYBOX)
+        os.write(2, "\npulling busybox\n".encode('utf-8'))
+        for data in c.pull('busybox', stream=True):
+            data = json.loads(data.decode('utf-8'))
+            os.write(2, ("%c[2K\r" % 27).encode('utf-8'))
+            status = data.get("status")
+            progress = data.get("progress")
+            detail = "{0} - {1}".format(status, progress).encode('utf-8')
+            os.write(2, detail)
+        os.write(2, "\npulled busybox\n".encode('utf-8'))
```
mnowster: I think we want to put back in the `c.inspect_image(BUSYBOX)` after we've finished pulling it here.

docker/client.py
```diff
@@ -239,10 +241,14 @@ def _multiplexed_response_stream_helper(self, response):
         # Disable timeout on the underlying socket to prevent
         # Read timed out(s) for long running processes
         socket = self._get_raw_response_socket(response)
-        if six.PY3:
-            socket._sock.settimeout(None)
-        else:
+
+        # Depending on a few factors we might need to access
```
mnowster: I appreciate having a comment here is intended to be helpful, the vagueness of the description of what those factors are that's effecting the decision here isn't helping though. By removing the `if six.PY3` statement it seems less clear to me now what one of those factors are.

I prefer checking `six.PY3` but I also know what the arguments are for doing it with `hasattr`. If you would like to keep the `hasattr` then it needs pulling out into a little function with some specific explanation. Something like `_set_socket_timeout(socket)` that can then include a doc string that explains what the circumstances are for when we have to set the timeout in different places.

tests/integration_test.py
```diff
@@ -1073,6 +1082,64 @@ def runTest(self):
         self.assertTrue(sock.fileno() > -1)
 
 
+class TestRunContainerReadingSocket(BaseTestCase):
```
mnowster: Yeh, this is hard to test isn't it! While I'm not a fan of copying in the dockertpy code, I appreciate it's difficult to write a test without it.

Minor point: please don't abbreviate variable names like `nxt` `nxt_size`, keep it fully expanded and descriptive, `next` `next_size`, readability is important. Thanks.

delfick: yeah, that comes about because next is a special word in python, so I write nxt to not overwrite it.

tests/integration_test.py
delfick: I do this so it's obvious that the tests aren't hanging, otherwise it can stay here for quite a while seemingly not doing anything.

shin-: I believe @mnowster meant adding the line `c.inspect_image(BUSBOX)` at the end here so if there's been an issue during the pull, it will be detected immediately, making it easier to pinpoint why the tests have failed if it occurred. Having the pull logs in addition to that is fine as far as I'm concerned :+1:

delfick: ah, that makes sense :)

tests/integration_test.py
shin-: I have no issue parsing `nxt` into `next` personally, so that doesn't bother me. 

@mnowster Is there a specific issue with those? Considering we're operating in python where `len` and `hasattr` and `def` are common keywords, I believe there's leeway for at least some abbreviation that still preserves clarity and readability.

delfick:
and I've repushed again :)

shin-:
Looks like something's gone awry with your rebase, did you check you were on the latest version of master?

delfick:
that was weird, I totes did that correctly.

oh well, pushed again.

We'll get there eventually :)

docker/client.py
```diff
@@ -240,10 +242,10 @@ def _multiplexed_response_stream_helper(self, response):
         # Disable timeout on the underlying socket to prevent
         # Read timed out(s) for long running processes
```
shin-: Looks like this appears twice now. We probably only want to keep the second occurrence.

shin-:
Much butter, thanks :)

docker/client.py
delfick: oops, fixed.

tests/integration_test.py
mnowster: It's a pep8 style that we avoid abbreviated variable names for the same reason we avoid single letter variables in for loops. Unless there's a very specific reason/context to not have it, this is not one of those times.

I appreciate the `next` conflicts with a reserved keyword, expanding it out to include more information is better than dropping a vowel. Thanks.

tests/integration_test.py
mnowster: Yep, thanks for expanding that @shin- , apologies @delfick that I wasn't clearer.

docker/client.py
```diff
@@ -272,6 +271,19 @@ def _stream_raw_result(self, response):
         for out in response.iter_content(chunk_size=1, decode_unicode=True):
             yield out
 
+    def _disable_socket_timeout(self, socket):
+        """ Depending on the combination of python version and whether we're
```
mnowster: Great. :+1: 

Minor: `settimout`'s have a typo, missing the `e`.

mnowster:
Minor typo to fix then it's a LGTM. 

Thanks for your continued effort on this @delfick :fireworks:

delfick:
Fixed the typo :)

shin-:
Squashed and merged 7f3692ceeda92ca3690394822cbd3c99378c0d7e

Thanks a ton!

delfick:
Excellent

Thank you so much!

PR #598, opened 2015-05-12
Fix the parameter mismatch

HackToday:
Closes-Issue: #597

### DIFF
--- a/docker/client.py
+++ b/docker/client.py
@@ -526,7 +526,7 @@ class Client(requests.Session):
             'exec_start instead', DeprecationWarning
         )
         create_res = self.exec_create(
-            container, cmd, detach, stdout, stderr, tty
+            container, cmd, stdout, stderr, tty
         )
 
         return self.exec_start(create_res, detach, tty, stream)


shin-:
Thanks! [Please note that `execute` will disappear in the next release!](http://docker-py.readthedocs.org/en/1.2.2/change_log/#120)

HackToday:
hi @shin-  If execute disapper, what is the replacement for execute ?

Like cli

docker exec   would disappear?

shin-:
As stated in the link I provided, you should use `exec_create` and `exec_start` instead.

HackToday:
OK. Thanks. Will use Client.exec_start

HackToday:
Hi @shin-  I am little confused for that.

Do you mean for docker exec function to work.

I need to code like this ?

        create_res = self.exec_create(
            container, cmd, detach, stdout, stderr, tty
        )

       self.exec_start(create_res, detach, tty, stream)


I did not understand what's benifit for developers to manual code like this, instead we just call 
execute(...) , execute(..) is more simple, it only have one call, but now I need to call exec_create and exec_start two calls manually.

Maybe I missed your point. Could you explain that to me ? Thanks

shin-:
1. In your example, `create_res` can also be used in `exec_inspect` (very useful to get the result of the exec command) and `exec_resize`. 
2. It's closer to the remote API design
3. It doesn't violate one of the core design principles which is that each method in client should be atomic (i.e. make **at most one** call to the remote API).
4. If none of that matters to you personally, you're still able to literally write a 2-line function that reproduces the behavior of `execute`.

Hopefully that clarifies the reason for the change.

HackToday:
Hi @shin-  OK I got know this history.

So for https://docs.docker.com/reference/commandline/cli/

if we want to support cli like docker exec, it means we have to combine like this

create_res = self.exec_create(
        container, cmd, detach, stdout, stderr, tty
    )

   self.exec_start(create_res, detach, tty, stream)


Right ? it seems have no other option

HackToday:
Hi @shin-  could you give your feedback about above comments ? Thanks

shin-:
Yes, this is correct.

PR #603, opened 2015-05-15
Support for .dockerignore 'folder/' syntax with tests

### LINKED ISSUE #581: .dockerignore ignores trailing slash on directories

mikexstudios:
Fixes #581.

Added unit tests for trailing slash on folders to https://github.com/docker/docker-py/pull/591 (by @shin-).

**NOTE**: There are still issues with how .dockerignore is implemented in docker-py. I'll open up another issue to discuss.


shin-:
Closing in favor of #604

PR #604, opened 2015-05-16
More rigorous support for .dockerignore

### LINKED ISSUE #581: .dockerignore ignores trailing slash on directories

mikexstudios:
Fixes #581. The current implementation for parsing `.dockerignore` files and applying the exclusion to the build path does not account for many common and uncommon syntax and behaviors (see `tests/test.py` for a few of these cases). The problem is that the rules for parsing `.dockerignore` files are complex. Luckily, [pathspec](https://github.com/cpburnz/python-path-specification) provides a robust implementation of `.gitignore` rules, which are almost identical for `.dockerignore`.

This PR uses the `pathspec` library in place of the current `fnmatch_any`. Additional tests for `.dockerignore` have been added. All tests pass for py27.

Thanks @shin-!

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,3 +1,6 @@
 requests==2.5.3
 six>=1.3.0
 websocket-client==0.11.0
+#pathspec >=0.3.3, <0.3.99
+# Currently use HEAD version until new pathspec release (0.3.4)
+-e git://github.com/cpburnz/python-path-specification.git#egg=pathspec
--- a/docker/utils/utils.py
+++ b/docker/utils/utils.py
@@ -25,6 +25,7 @@ from datetime import datetime
 
 import requests
 import six
+import pathspec
 
 from .. import errors
 from .. import tls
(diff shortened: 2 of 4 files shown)


mikexstudios:
This PR supersedes #603 and #591.

shin-:
Thanks, I'll give that a serious look over later this week. Can you look into the test failures in the meantime? https://travis-ci.org/docker/docker-py/builds/62859246

requirements.txt
```diff
@@ -1,3 +1,6 @@
 requests==2.5.3
 six>=1.3.0
 websocket-client==0.11.0
+#pathspec >=0.3.3, <0.3.99
```
shin-: Let's pin dependencies in requirements.txt ; setup.py can have version ranges.

shin-:
You'll need to add `pathspec >=0.3.3, <0.4.0` to the dependencies in `setup.py`.

aanand:
I've continued this in https://github.com/docker/docker-py/pull/721.

PR #605, opened 2015-05-17
Raise ValueError on empty argument to inspect_{container,image}() methods.

### LINKED ISSUE #602: Fixed in #605 - Client.inspect_image("") behaves like Client.images()
```py
>>> from os import environ
>>> from docker import Client
>>> c = Client(environ['DOCKER_HOST'])
>>> type(c.images())
<type 'list'>
>>> >>> len(c.images()) > 0
True
>>> c.inspect_image('') == c.images()
True
```

This isn't hard to see why. `inspect_image()` calls `/images/{image_id}/json`. `images()` calls `/images/json`. Where the `{image_id}` is an empty string, the [resulting URL](https://github.com/docker/docker-py/blob/bf00acf854a3b78335c1157f4a1a83488e611bb8/docker/client.py#L751) for the call to `inspect_image()` is `/images//json`:

```sh
% [ "$( curl --location --post30{1,2,3} --silent 'http://<host>:2375/images//json' -H 'Content-Type: application/json' -H 'Accept: application/json' -X GET )" \
> = "$( curl --location --post30{1,2,3} --silent 'http://<host>:2375/images/json' -H 'Content-Type: application/json' -H 'Accept: application/json' -X GET )" ] \
> && echo 'SAME' || echo 'NOT THE SAME'
SAME
```

By analogy, I suspect the `inspect_container()` and `containers()` methods suffer from a similar problem.

The expected format of the `inspect_*()` methods (a detailed object) is different from the corresponding `*s()` methods (an array of summary objects), so it seems to me calling `inspect_*("")` should probably raise an exception. Maybe something like:

```py
    @check_resource
    def inspect_container(self, container):
        ...
        if not container:
            raise ValueError('container ID required')
        ...

    @check_resource
    def inspect_image(self, image):
        ...
        if not image:
            raise ValueError('image ID required')
        ...
```

I can create a PR if desired.

posita:
Fix #602. Raise `ValueError` on empty argument to `inspect_{container,image}()` methods.

### DIFF
--- a/docker/client.py
+++ b/docker/client.py
@@ -747,6 +747,8 @@ class Client(requests.Session):
 
     @check_resource
     def inspect_image(self, image):
+        if isinstance(image, dict):
+            image = image.get('Id')
         return self._result(
             self._get(self._url("/images/{0}/json".format(image))),
             True
--- a/docker/utils/decorators.py
+++ b/docker/utils/decorators.py
@@ -12,5 +12,7 @@ def check_resource(f):
                 raise errors.NullResource(
                     'image or container param is undefined'
                 )
+        if not resource_id:
+            raise ValueError('image or container param is empty')
         return f(self, resource_id, *args, **kwargs)
     return wrapped
--- a/tests/test.py
+++ b/tests/test.py
@@ -1782,6 +1782,12 @@ class DockerClientTest(Cleanup, base.BaseTestCase):
         except Exception as e:
             self.fail('Command should not raise exception: {0}'.format(e))
 
+        try:
+            self.client.inspect_container('')
+        except ValueError as e:
+            self.assertEqual(e.args[0],
+                             'image or container param is empty')
+
         fake_request.assert_called_with(
             url_prefix + 'containers/3cc2351ab11b/json',
             timeout=DEFAULT_TIMEOUT_SECONDS
@@ -1952,6 +1958,12 @@ class DockerClientTest(Cleanup, base.BaseTestCase):
             self.client.inspect_image(fake_api.FAKE_IMAGE_NAME)
         except Exception as e:
             self.fail('Command should not raise exception: {0}'.format(e))
+
+        try:
+            self.client.inspect_image('')
+        except ValueError as e:
+            self.assertEqual(e.args[0],
+                             'image or container param is empty')
 
         fake_request.assert_called_with(
             url_prefix + 'images/test_image/json',


docker/client.py
```diff
@@ -747,6 +749,10 @@ def inspect_container(self, container):
 
     @check_resource
     def inspect_image(self, image):
+        if not image:
+            raise ValueError('image or ID required')
+        if isinstance(image, dict):
```
posita: This wasn't part of the original issue, but is included to provide consistency with the inspect_container() method.

docker/client.py
```diff
@@ -747,6 +749,10 @@ def inspect_container(self, container):
...
+            image = image.get('Id')
         return self._result(
```
posita: The `isinstance` block wasn't part of the original issue, but is included to provide consistency with the inspect_container() method.

docker/client.py
```diff
@@ -747,6 +749,10 @@ def inspect_container(self, container):
 
     @check_resource
     def inspect_image(self, image):
+        if not image:
+            raise ValueError('image or ID required')
+        if isinstance(image, dict):
+            image = image.get('Id')
```
posita: This `isinstance` block wasn't part of the original issue, but is included to provide consistency with the inspect_container() method.

docker/client.py
```diff
@@ -739,6 +739,8 @@ def insert(self, image, url, path):
 
     @check_resource
     def inspect_container(self, container):
+        if not container:
+            raise ValueError('container or ID required')
         if isinstance(container, dict):
             container = container.get('Id')
```
posita: `if isinstance ...` block added to `inspect_image()`. See note below.

shin-:
This would be better done in the `check_resource` decorator.

posita:
@shin, I've moved the check to `@check_resource` as you suggest, and tests pass. Are you sure this won't have detrimental side effects elsewhere?

docker/client.py
```diff
@@ -747,6 +747,8 @@ def inspect_container(self, container):
 
     @check_resource
     def inspect_image(self, image):
+        if isinstance(image, dict):
+            image = image.get('Id')
         return self._result(
```
posita: This isn't required to fix #602, but it's there to make this method more consistent with `inspect_container`.

docker/client.py
```diff
@@ -747,6 +747,8 @@ def inspect_container(self, container):
 
     @check_resource
     def inspect_image(self, image):
+        if isinstance(image, dict):
+            image = image.get('Id')
```
posita: This isn't required to fix #602, but the change is make this method consistent with `inspect_container`.

@shin, incidentally, this also looks like a good candidate for inclusion in `@check_resource`. Similar checks are littered throughout `client.py` (but still missing from some methods). See my [raise-on-inspect-empty-string-602-with-id-resolution-in-check-resource](https://github.com/posita/docker-py/tree/raise-on-inspect-empty-string-602-with-id-resolution-in-check-resource) branch. (Tests pass.)

posita:
BTW, if this PR looks good for merge as-is, I can propose another that moves code like the following into `@check_resource` (see my line note on `docker/client.py` above):

```py
    if isinstance(..., dict):
        ... = ....get('Id')
```

I think those changes are beyond the scope of this PR though.

shin-:
Thanks!

PR #606, opened 2015-05-20
Remove non-existing argument

bo0ts:
There is no container argument in create_container_config. This is probably a leftover when this was copied from the Client.start documentation


shin-:
Good catch, thanks!

PR #607, opened 2015-05-20
READY FOR REVIEW - Move image/container ID resolution to @check_resource decorator.

posita:
As [promised](https://github.com/docker/docker-py/pull/605#issuecomment-103903948), here's a PR moving code like this...

```py
    if isinstance(..., dict):
        ... = ....get('Id')
```

...into the `@check_resource` decorator.

### DIFF
--- a/docker/client.py
+++ b/docker/client.py
@@ -251,8 +251,6 @@ class Client(requests.Session):
     @check_resource
     def attach(self, container, stdout=True, stderr=True,
                stream=False, logs=False):
-        if isinstance(container, dict):
-            container = container.get('Id')
         params = {
             'logs': logs and 1 or 0,
             'stdout': stdout and 1 or 0,
@@ -296,9 +294,6 @@ class Client(requests.Session):
 
         if ws:
             return self._attach_websocket(container, params)
-
-        if isinstance(container, dict):
-            container = container.get('Id')
 
         u = self._url("/containers/{0}/attach".format(container))
         return self._get_raw_response_socket(self.post(
@@ -410,8 +405,6 @@ class Client(requests.Session):
     @check_resource
     def commit(self, container, repository=None, tag=None, message=None,
                author=None, conf=None):
-        if isinstance(container, dict):
-            container = container.get('Id')
         params = {
             'container': container,
             'repo': repository,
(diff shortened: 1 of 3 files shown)


mpetazzoni:
LGTM. @shin- ?

shin-:
Mmmh, my concerns are:

* if we put it here the condition will also be unnecessarily evaluated in the case of image-oriented functions. Seems like a bad practice even though the overhead is negligible.
* I'd rather have several short, one-purpose decorators than a long, single, multi-purpose one.

posita:
> I'd rather have several short, one-purpose decorators than a long, single, multi-purpose one.

I would tend to agree with this, except that these behaviors *already* seem very closely coupled throughout `client.py`. It seems (at least so far), everywhere that `@check_resource` is used, the code in the OP is repeated (or if it isn't, it was intended to be). Just look at how many places I removed those lines from `client.py`.

> if we put it here the condition will also be unnecessarily evaluated in the case of image-oriented functions. Seems like a bad practice even though the overhead is negligible.

Perhaps, but this is already the case. Well before I came along, `@check_resource` already pulled double duty between containers and images. Perhaps it makes more sense to have nearly identical, but separately named decorators for images and containers (e.g., `@check_image` and `@check_container`). I can do that, but that seems to be a more significant change than this was intended to be.

All I really meant to do here was an incremental improvement towards DRY without disturbing the status quo too much. Ultimately, it's your call whether such a thing is worth merging. Either way, the fact that the code in the OP shows up in so many places is probably indicative that *something* needs to be done.

posita:
> I'd rather have several short, one-purpose decorators than a long, single, multi-purpose one.

I would tend to agree with this, except that these behaviors *already* seem very closely coupled throughout `client.py`. It seems (at least so far), everywhere that `@check_resource` is used, the code in the OP is repeated (or if it isn't, it was intended to be). Just look at how many places I removed those lines from `client.py`.

> if we put it here the condition will also be unnecessarily evaluated in the case of image-oriented functions. Seems like a bad practice even though the overhead is negligible.

Perhaps, but this is already the case. Well before I came along, `@check_resource` already pulled double duty between containers and images. Perhaps it makes more sense to have nearly identical, but separately named decorators for images and containers (e.g., `@check_image` and `@check_container`). I can do that, but that seems to be a more significant change than this was intended to be.

Plus, the overhead may be comparable when stacking the same two decorators everywhere:

```py
@check_container
@check_id
def attach(self, container, ...):
    ...
```

I know that wasn't your original point though.

All I really meant to do here was an incremental improvement towards DRY without disturbing the status quo too much. Ultimately, it's your call whether such a thing is worth merging. Either way, the fact that the code in the OP shows up in so many places is probably indicative that *something* needs to be done.

shin-:
Thanks!

PR #609, opened 2015-05-21
Update docker-py to use a more portable sense of HOME.

mattmoor:
This makes docker-py consistent with Docker's newish way of establishing the path to .dockercfg:
   https://github.com/docker/docker/blob/master/pkg/homedir/homedir.go

NOTE: this is only now relevant as Docker 1.6 was the first release with an official Windows build.


shin-:
Thank you!

PR #611, opened 2015-05-22
Add support for "log_config" parameter in client.start method.

ryanwalls:
PR #563 by @shin- started the support for log driver.  This PR adds `log_config` as a parameter to the client `start` method.


ryanwalls:
I see that passing host config parameters in start() is now deprecated.  Closing this PR.

PR #612, opened 2015-05-26
Typo in ReadonlyRootfs

ibuildthecloud:
read_only parameter does not work because the wrong name is sent to Docker


mpetazzoni:
Thanks!

Confirmed from https://docs.docker.com/reference/api/docker_remote_api_v1.18/#create-a-container

PR #613, opened 2015-05-26
Add log_config field to create

ibuildthecloud:


ibuildthecloud:
Oops, this was the wrong approach

PR #614, opened 2015-05-26
Allow extra_hosts to be a list too

ibuildthecloud:
The current map syntax does not allow the API equivalent of

    --add-host foo:1.1.1.1 --add-host foo:2.2.2.2

The above will map one hostname to two IPs.  The above is valid in Docker.

### DIFF
--- a/docker/utils/utils.py
+++ b/docker/utils/utils.py
@@ -434,7 +434,7 @@ def create_host_config(
                 for k, v in sorted(six.iteritems(extra_hosts))
             ]
 
-            host_config['ExtraHosts'] = extra_hosts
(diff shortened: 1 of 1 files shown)


docker/utils/utils.py
```diff
@@ -436,6 +436,9 @@ def create_host_config(
 
             host_config['ExtraHosts'] = extra_hosts
```
shin-: just extract that line from the inner block instead?

ibuildthecloud:
Updated as per comment

ibuildthecloud:
@shin Can this be merged now?

shin-:
Thanks! :+1:  sorry about the delay.

PR #617, opened 2015-05-28
Fix pinging an unauthenticated v2 registry

mmikulicic:
Hi,

We're building an auth server for docker v2 registry (without a reverse proxy in front of it) and noticed that the docker python client fails to ping the registry because it returns 401 Unauthorized.

I suggest to consider 401 as a sign that the server is alive and kicking. Or at least until the v2 registry is updated to allow some form of unauthenticated pinging.

### DIFF
--- a/docker/utils/utils.py
+++ b/docker/utils/utils.py
@@ -129,7 +129,9 @@ def ping(url):
     except Exception:
         return False
     else:
-        return res.status_code < 400
+        # We don't send yet auth headers
+        # and a v2 registry will respond with status 401
+        return res.status_code == 401 or res.status_code < 400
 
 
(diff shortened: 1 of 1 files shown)


rojer:
let me provide additional clarification.
this bug is only triggered when registry 2.0 is used with token authentication directly with no proxy in front of it.
because we've built a token authentication server (and are about to release it) there is no need for us to run the proxy.

jesper:
Got the same issue - would be grateful if this was merged to master.

shin-:
Thank you!

newhook:
So when can get a new version of docker-py? This issue is causing some major pain for me.

shin-:
End of June is the current plan.

PR #625, opened 2015-06-02
Add timeout to client pull method

jongiddy:
The `build` (#209) and `wait` (#301) methods have recently had a `timeout` parameter added to allow callers to set a timeout.  Like these calls, the default timeout is not suitable for the `pull` call (#100), but it seems reasonable to allow a caller to set some timeout explictly.


ilonajulczuk:
Hi, I have a similar issue, I would like to have a timeout parameter in `exec_start`.

wallrj:
Hey @jongiddy 

This branch needs merging forward.

I'm upgrading to docker-py 1.3.1  and noted that the `pull` method still doesn't have a timeout method so this is still needed.

PR #629, opened 2015-06-05
Allow any mode string to be passed into a volume bind

aanand:
- Volume binds now take a `mode` key, whose value can be any string.
- `ro` is still supported. It is an error to specify both `ro` and `mode`.

The rationale for this: Docker has added support for a [variety of volume modes](https://github.com/docker/docker/blob/53d9609de472f2fc9e6aaef2b713f526605f4ab8/daemon/volumes.go#L115-L132). I reckon it's best to leave it to the Docker daemon to validate them.


shin-:
Thanks :)

Issue #630, opened 2015-06-07
Attaching and streaming logs from containers started with tty=true does not work properly


dano:
Currently, when docker-py attaches to a container, it will always try to read the output stream as if it was multiplexed, unless the API version is below 1.6:
```python
        # Stream multi-plexing was only introduced in API v1.6. Anything before
        # that needs old-style streaming.
         
        if utils.compare_version('1.6', self._version) < 0:
            def stream_result():
                self._raise_for_status(response)
                for line in response.iter_lines(chunk_size=1,
                                                decode_unicode=True):
                    # filter out keep-alive new lines
                    if line:
                        yield line

            return stream_result() if stream else \
                self._result(response, binary=True)

        sep = bytes() if six.PY3 else str()

        if stream:
            return self._multiplexed_response_stream_helper(response)
        else:
            return sep.join(
                [x for x in self._multiplexed_buffer_helper(response)]
            )
```
`Client.logs` similarly assumes a multiplexed response stream after calling the `container/<id>/logs` API, unless the API is below version 1.11, in which cases it just calls `attach` with the `logs=True` keyword argument.

However, the docker API docs note that if a container is opened with the tty=true option, [you just get the a raw stream, not a multiplexed one](http://docs.docker.com/reference/api/docker_remote_api_v1.18/#attach-to-a-container):

>Stream details:
>
>When using the TTY setting is enabled in POST /containers/create , the stream is the raw data from >the process PTY and client's stdin. When the TTY is disabled, then the stream is multiplexed to >separate stdout and stderr.

The code for both `attach` and `logs` need to handle cases where the container that is being attached to/streamed from has tty=true. Currently, you'll end up not seeing any output, or seeing delayed and incompleted output from such containers.

shin-:
Closing via #669

PR #631, opened 2015-06-07
Fix handling output from tty-enabled containers.

dano:
As described in issue #630, getting the stdout/stderr from tty-enabled containers doesn't work properly, because docker-py treats the output as multiplexed, when docker is actually sending a raw stream. This patch attempts to address this by determining if a container to is tty-enabled before trying to read its output, and if it is, read the raw stream, rather than reading at a multiplexed stream.

One piece here I'm not 100% sure about is what to do with the logic that was in `attach` to handle old API versions. The logic being used there - read the stream line by line, and skip empty new-lines, doesn't work if the container ever writes blank lines. I also haven't seen any keep-alive new lines being written in my testing, so I wonder if this check isn't really needed?

I suppose I could split `_stream_raw_result` into two methods - one for TTY-enabled containers, and one for versions below 1.6. The one for versions below 1.6 could keep the old, line-by-line behavior.

Also please let me know if any additional unit tests are needed.


docker/client.py
```diff
@@ -591,15 +590,19 @@ def exec_start(self, exec_id, detach=False, tty=False, stream=False):
...
                               data=data, stream=stream)
+        # Stream multi-plexing was only introduced in API v1.6. Anything before
+        # that needs old-style streaming. We also should use old-style
+        # streaming if we're dealing with a tty-enabled container.
+        if utils.compare_version('1.6', self._version) < 0 or tty:
+            return self._stream_raw_result(res) if stream else \
+                self._result(res, binary=True)
+
         self._raise_for_status(res)
+        sep = bytes() if six.PY3 else str()
```
aanand: This can be replaced with just `six.binary_type()`, I believe.

docker/client.py
```diff
@@ -824,14 +827,21 @@ def logs(self, container, stdout=True, stderr=True, stream=False,
                 params['tail'] = tail
             url = self._url("/containers/{0}/logs".format(container))
             res = self._get(url, params=params, stream=stream)
```
aanand: Can we reduce the duplication here? This is very similar to the same code in `attach()`.

aanand:
> I suppose I could split `_stream_raw_result` into two methods - one for TTY-enabled containers, and one for versions below 1.6. The one for versions below 1.6 could keep the old, line-by-line behavior.

Sounds sensible to me.

> Also please let me know if any additional unit tests are needed.

I'm not sure I fully understand your changes to the tests. Are they now exclusively testing the `Tty=True` case? If so, we should make sure to also test the `Tty=False` case.

dano:
@aanand The tests currently don't actually specify a tty setting - the changes I made add a `tty=False` setting to the mocked container object used in each test, just to keep the tests from failing because `container['Config']['Tty'] is missing. I'll look into adding a test for the `tty=True` case in addition to just keeping the existing tests working.

dano:
Ugh, it looks like my attempt to squash my commits ended up screwing up the branch completely. I'll open a new request...

dano:
Ok, I've opened https://github.com/docker/docker-py/pull/669 to replace this.

PR #632, opened 2015-06-10
Add exception check to push before returning stream

bcicen:
As underlying exceptions(such as push already in progress) will be hidden otherwise. I haven't come across them yet, but I presume that exceptions are being hidden by other methods that use streaming output as well.


bcicen:
Closing in favor of #634

PR #633, opened 2015-06-10
Docs: Update boot2docker shellinit example to use 'eval'

edmorley:
The boot2docker documentation has since changed the recommended way to use shellinit - see boot2docker/boot2docker#786. The command also no longer prints the export lines to the console, so have updated the example output.


shin-:
Thanks!

PR #634, opened 2015-06-12
Add response check to stream helper

bcicen:
as underlying exceptions(such as push already in progress) will be hidden in the
returned generator otherwise.


bcicen:
closing in favor of #638

PR #637, opened 2015-06-12
Pass log_config to start

fin09pcap:
* pass log_config to allow for different log_config options by container


shin-:
See #611 and [docs](https://github.com/docker/docker-py/blob/master/docs/api.md#start)

PR #639, opened 2015-06-15
Add and document a decode parameter for build

bo0ts:
This makes the build method consistent with the events method and adds
some convenience.


shin-:
Thanks! [Merged](https://github.com/docker/docker-py/commit/3a32be79d310982b598ca2c9a391180c87099dae)

PR #643, opened 2015-06-18
Persist docker authentification

olsaki:
- Persist docker authentication
- Compatible with CLI client's docker login

### DIFF
--- a/docker/auth/auth.py
+++ b/docker/auth/auth.py
@@ -79,6 +79,31 @@ def resolve_authconfig(authconfig, registry=None):
     return None
 
 
+def write_authconfig(authconfig, config_path=None):
+    """
+    Write the authentification data from a given auth configuration to a file.
+    """
+    docker_conf = {}
+
+    config_file = config_path or os.path.join(os.environ.get('HOME', '.'),
+                                              DOCKER_CONFIG_FILENAME)
+
+    # if config path doesn't exist return False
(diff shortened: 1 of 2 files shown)


olsaki:
Can someone please review this?

aanand:
Thanks for the contribution!

First of all: @shin-, would you consider this functionality within scope for docker-py? I'm a bit nervous about writing out auth files - if the format or location ever changes, software which uses docker-py could end up breaking things, unless both docker-py and the Docker client are very careful about backwards-compatibility.

I have a few comments/questions regarding the code:

- "authentification" should be "authentication"
- The return value of `write_authconfig` is unused. Why is it there?
- We don't need both `config_path` and `config_file` variables.
- If the path doesn't exist, should we create it, rather than silently continuing?
- If the path does exist, are we *sure* that we're not overwriting any auth data that's already there? i.e have we definitely loaded it?

olsaki:
@aanand @shin- ping. Any update on the idea of this in general? Please let me know if you think we should go ahead with this and I will continue to look into / work on the points mentioned above.

dnephin:
I'd be ok with the idea as long as we make it opt-in. Adding a new kwarg that defaults to False, and when it's true we can write the config.

olsaki:
@dnephin good idea. Shouldn't be a problem to do that. Waiting for input from @aanand and @shin- to proceed.

PR #644, opened 2015-06-19
memory / memswap in hostconfig

### LINKED ISSUE #623: memory_limit and memswap_limit support in host_config
Unable to set memory_limit and memswap_limit in host_config, both are missing from supported parameters in utils.py.

Containers created using python client,

	[{
	        "Config":  {
	        ...
	        "Memory": 2147483648,
	        "MemorySwap": -1,
	       ...
	       "HostConfig": {
	        ...
	        "Memory": 0,
	        "MemorySwap": 0,
	        ...

Containers created using commandline with --memory=2048MB --memory_swap=-1

	[{
	        "Config":  {
	        ...
	        "Memory": 2147483648,
	        "MemorySwap": -1,
	       ...
	       "HostConfig": {
	        ...
	        "Memory": 2147483648,
	        "MemorySwap": -1,
	        ...

As results, limits are only configured in Config and not Host_Config and therefore make no effect.

shin-:
Fixes #623

### DIFF
--- a/docker/client.py
+++ b/docker/client.py
@@ -464,11 +464,11 @@ class Client(requests.Session):
 
     def create_container(self, image, command=None, hostname=None, user=None,
                          detach=False, stdin_open=False, tty=False,
-                         mem_limit=0, ports=None, environment=None, dns=None,
-                         volumes=None, volumes_from=None,
+                         mem_limit=None, ports=None, environment=None,
+                         dns=None, volumes=None, volumes_from=None,
                          network_disabled=False, name=None, entrypoint=None,
                          cpu_shares=None, working_dir=None, domainname=None,
-                         memswap_limit=0, cpuset=None, host_config=None,
+                         memswap_limit=None, cpuset=None, host_config=None,
                          mac_address=None, labels=None, volume_driver=None):
 
         if isinstance(volumes, six.string_types):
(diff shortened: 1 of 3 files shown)


mpetazzoni:
:+1:

As a side node, it'd be nice if the `InvalidVersion` exception actually took the minimum required version as an argument so that we can emit, in all cases, a more structured error message.

shin-:
Created #664 to keep track of that separately - I agree, but we have to keep in mind it's used when the version is too low as well as too high.

veggiemonk:
Hi everyone,

This fix breaks ansible docker module. Do you know it can be fixed? :cry: 
Thanks in advance.

shin-:
Fix the fix? ~

Jokes aside, the onus is on ansible to support the new way of passing those arguments (otherwise it will break with newer versions of Docker anyway). In the meantime, you can:
- pin the docker-py version to 1.2.3 
- somehow tell ansible to use a Docker API version < 1.19 (the `version` parameter in `docker.Client()`)

veggiemonk:
Thanks a lot! Will give it a try... :smile:

veggiemonk:
OK. It took me a while to figure out that it was on the server that the version needed to be pinned. 
Thanks a lot for your help!
Cheers

PR #645, opened 2015-06-19
Update websocket client dependency

shin-:
Bump to latest (0.32.0) - which also supports Python 3, allowing us to remove some obsolete code.


aanand:
Nice.

LGTM

PR #652, opened 2015-06-24
ClientBase extraction

shin-:


mpetazzoni:
Unless you expect to have other subclasses of `ClientBase`, I don't really see the point.

If readability of the `Client` class is a concern, then yes some utility methods need to be extract, but in a way that doesn't depend on instance state (aka by taking what they need as arguments). Then you can extract those methods into one or more side helper modules.

shin-:
I agree that the end goal should be to have most of those outside of the class, but I was considering this as a first step in separating API code from internal code. I strongly believe it's still better than the current state of the code and it's relatively low effort.

mpetazzoni:
Alright, let's work from there.

Issue #648, opened 2015-06-24
Use ~/.docker/config.json for registry authentication


juhasz:
Docker now stores authentication data in `~/.docker/config.json`, but docker-py still relies on `~/.dockercfg`, so application based on docker-py can't access private repositories.

See docker/docker#12009 and docker/compose#1590

PR #654, opened 2015-06-25
Prefer new Docker config location and format.

Melraidin:
This tries to load Docker authentication info from
~/.docker/config.json before falling back to its legacy location and
format at ~/.dockercfg.

Resolves https://github.com/docker/docker-py/issues/648

### DIFF
--- a/docker/auth/auth.py
+++ b/docker/auth/auth.py
@@ -23,7 +23,8 @@ from ..utils import utils
 from .. import errors
 
 INDEX_URL = 'https://index.docker.io/v1/'
-DOCKER_CONFIG_FILENAME = '.dockercfg'
+DOCKER_CONFIG_FILENAME = os.path.join('.docker', 'config.json')
+LEGACY_DOCKER_CONFIG_FILENAME = '.dockercfg'
 
 
 def expand_registry_url(hostname, insecure=False):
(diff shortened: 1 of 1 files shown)


Melraidin:
@mpetazzoni Refactored a bit of that to avoid some duplication. Also allowed the provided config_path to override the default both for the old and new config locations.

Melraidin:
On a side note it'd be nice if we could give some feedback to the caller on whether any authentication was sent. Without this docker-compose is left to simply say that a repo wasn't found with no indication that it could be a simple authentication issue.

docker/auth/auth.py
```diff
@@ -115,9 +139,25 @@ def load_config(config_path=None):
...
                                               DOCKER_CONFIG_FILENAME)
 
+    if os.path.exists(config_file):
+        try:
+            with open(config_file) as f:
+                for section, data in six.iteritems(json.load(f)):
+                    if section != 'auths':
+                        continue
+                    return parse_auth(data)
+        except:
```
aanand: Could this be scoped to the specific type(s) of exception we're expecting?

Melraidin:
@aanand I was aiming to follow the style later in the code for the legacy location where we're ignoring all exceptions but your point's totally valid. Will update.

aanand:
Just saw that line, yes - we should fix that too (in a separate PR).

mpetazzoni:
:+1:  LGTM. Might be worth squashing those together.

shin-:
LGTM. Thanks!

Issue #656, opened 2015-06-26
client.logs(stream=True) is broken when using docker 1.7


paetling:
@shin- 

The streaming functionality of client.logs is broken when using docker 1.7.  I am working on a fix and have found the reason that it is now broken:
Line line 130 in the file daemon/logs.go
```
// write an empty chunk of data (this is to ensure that the
// HTTP Response is sent immediatly, even if the container has
// not yet produced any data)
outStream.Write(nil)
```

This messes up how `_multiplexed_response_stream_helper` works (will break if it sees a response like this).  

The thing I am really wondering though, is why the test to catch this in integration_test.py has been commented out since this commit 8a63e70b in June 2014? This was a commit by you, @shin-.  It seems like this is the exact reason you have integration tests, to catch bugs.   I am not sure of the the best way to proceed given that I cannot show you a failing test.   I would really appreciate it if you could explain:

1. Why the test was removed
2. Steps I need to take to put the tests back and merge my fix. 

Alex

jasonamyers:
I'm also having this issue

shin-:
#660 was merged but I have to look into the test removal question still, so I'm keeping this ticket open for now.

spinus:
As a workaround I am using ``client.attach(logs=True, stream=True)`` - works ok.

shin-:
Test re-added in #848

spinus:
yey :-)

PR #660, opened 2015-06-29
fix premature EOF detection in streams

a-ba:
The docker engine may send empty chunks of data in the stream
(especially since https://github.com/docker/docker/pull/13033)
They should not be taken for EOF

Signed-off-by: Anthony Baire <Anthony.Baire@irisa.fr>

### DIFF
--- a/docker/client.py
+++ b/docker/client.py
@@ -233,7 +233,7 @@ class Client(requests.Session):
                 break
             _, length = struct.unpack('>BxxxL', header)
             if not length:
-                break
+                continue
             data = response.raw.read(length)
             if not data:
                 break


shin-:
Is this similar to #631?

a-ba:
It is a different issue. #631 deals with choosing the right reader 
('multiplexed' or 'raw')

This PR fixes a bug in the 'multiplexed' reader.

shin-:
Fixes #656

yank1:
+1

aanand:
LGTM

thieman:
The existence of this bug suggests a hole in integration testing, can we fix that?

aanand:
@thieman +1, actually

thieman:
Can we get this fix merged for 1.3.0 one way or the other? This is a serious problem for us.

shin-:
Sorry, it was too late to make the cut, but I'll look into having it in a patch version.

PR #662, opened 2015-06-30
Exec create user

stevenewey:
The user parameter has been present but unimplemented in the code for a little while, I'm not sure if there's a good reason for that? Hopefully this is an good enough implementation of it.

### DIFF
--- a/docker/client.py
+++ b/docker/client.py
@@ -521,7 +521,7 @@ class Client(requests.Session):
         return self.exec_start(create_res, detach, tty, stream)
 
     def exec_create(self, container, cmd, stdout=True, stderr=True, tty=False,
-                    privileged=False, user=None):
+                    privileged=False, user=''):
         if utils.compare_version('1.15', self._version) < 0:
             raise errors.InvalidVersion('Exec is not supported in API < 1.15')
         if privileged and utils.compare_version('1.19', self._version) < 0:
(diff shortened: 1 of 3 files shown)


tests/integration_test.py
```diff
@@ -830,6 +830,23 @@ def runTest(self):
...
+        id = container['Id']
+        self.client.start(id)
+        self.tmp_containers.append(id)
+
+        res = self.client.exec_create(id, 'whoami', user='default')
+        self.assertIn('Id', res)
+
+        exec_log = self.client.exec_start(res)
+        expected = b'default' if six.PY3 else 'default\n'
+        self.assertEqual(exec_log, expected)
```
aanand: Could we also test that `whoami` returns `root` if `user` isn't passed in?

stevenewey: Added.

aanand:
LGTM

schu:
+1

docker/client.py
```diff
@@ -521,19 +521,26 @@ def execute(self, container, cmd, detach=False, stdout=True, stderr=True,
...
             )
+        if user and utils.compare_version('1.19', self._version) < 0:
+            raise errors.InvalidVersion(
+                'User-specific exec is not supported in API < 1.19'
+            )
         if isinstance(cmd, six.string_types):
             cmd = shlex.split(str(cmd))
 
+        if user is None:
+            user = ''
```
shin-: Can you explain the reasoning here? Seems unnecessary.

stevenewey: You're right, it's not necessary, I'll remove it.

shin-:
Awesome, thanks!

PR #669, opened 2015-07-06
Fix handling output from tty-enabled containers.

dano:
As described in issue #630, getting the stdout/stderr from tty-enabled containers doesn't work properly, because docker-py treats the output as multiplexed, when docker is actually sending a raw stream. This patch attempts to address this by determining if a container to is tty-enabled before trying to read its output, and if it is, read the raw stream, rather than reading at a multiplexed stream.

I previously submitted a pull request for this issue [here](https://github.com/docker/docker-py/pull/631), but ended up screwing up the revision history in that branch, so I've opened a new one. Please see that request for comments on an earlier version of the diff.

### DIFF
--- a/docker/clientbase.py
+++ b/docker/clientbase.py
@@ -221,6 +221,46 @@ class ClientBase(requests.Session):
                 break
             yield data
 
+    def _stream_raw_result_old(self, response):
+        ''' Stream raw output for API versions below 1.6 '''
+        self._raise_for_status(response)
+        for line in response.iter_lines(chunk_size=1,
+                                        decode_unicode=True):
+            # filter out keep-alive new lines
+            if line:
+                yield line
+
+    def _stream_raw_result(self, response):
+        ''' Stream result for TTY-enabled container above API 1.6 '''
+        self._raise_for_status(response)
+        for out in response.iter_content(chunk_size=1, decode_unicode=True):
+            yield out
+
+    def _get_result(self, container, stream, res):
+        cont = self.inspect_container(container)
+        return self._get_result_tty(stream, res, cont['Config']['Tty'])
+
+    def _get_result_tty(self, stream, res, is_tty):
+        # Stream multi-plexing was only introduced in API v1.6. Anything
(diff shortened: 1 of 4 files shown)


aanand:
LGTM

For future reference, you can always `git push --force` to a PR's branch, rather than close it and open a new one.

docker/clientbase.py
```diff
@@ -221,6 +221,46 @@ def _multiplexed_response_stream_helper(self, response):
...
+                yield line
+
+    def _stream_raw_result(self, response):
+        ''' Stream result for TTY-enabled container above API 1.6 '''
+        self._raise_for_status(response)
+        for out in response.iter_content(chunk_size=1, decode_unicode=True):
+            yield out
+
+    def _get_result(self, container, stream, res):
+        cont = self.inspect_container(container)
```
shin-: I'm a bit miffed by that. It doesn't seem right that we would have to make additional API calls when streaming content. I understand why it's done obviously, but I'd like us to try to see if there's an alternative that doesn't incure additional calls.

dano: @shin- I share your concern - I just don't know what other option there is to determine if you're dealing with a TTY-enabled container or not. FWIW, [it looks like the Docker CLI clientuses the same technique](https://github.com/docker/docker/blob/master/api/client/attach.go#L29) (assuming I'm reading the source correctly).

aanand: As an optimisation, `attach` (and `_get_result`) could take an optional argument to force TTY or non-TTY behaviour. Syntax to be determined, but e.g.:

```python
client.attach(force_tty=True)
client.attach(force_no_tty=True)
```

The `inspect_container` call could then be skipped.

This wouldn't need to be implemented as part of this PR - the pressing concern is to fix the bug.

shin-: If that's how the docker client does it, I guess we'll have to do it as well. Let's merge for now - the fix is definitely needed. We can revisit later as @aanand said.

PR #674, opened 2015-07-14
Update all client instances in integration tests to use DOCKER_HOST

estesp:
In testing Docker master with the latest docker-py release, a few new
tests exist which don't set the `base_url` on client creation, which is not
the default unix socket location when Docker's integration tests are running.

Also, there is no guarantee syslog can be reached from a test container
so I wrapped that test with the `NOT_ON_HOST` skip logic.

Docker-DCO-1.1-Signed-off-by: Phil Estes <estesp@linux.vnet.ibm.com> (github: estesp)

### DIFF
--- a/tests/integration_test.py
+++ b/tests/integration_test.py
@@ -242,6 +242,7 @@ class TestCreateContainerWithRoBinds(BaseTestCase):
         self.assertFalse(inspect_data['VolumesRW'][mount_dest])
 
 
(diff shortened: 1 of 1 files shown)


estesp:
See docker/docker#14628 for my attempt to update Docker's `Dockerfile` to use the docker-py 1.3.0 release.

shin-:
LGTM!

aanand:
LGTM

PR #677, opened 2015-07-17
Switch to send full AuthConfig object for build action

moutten:
In order to support the docker API for version 1.7+, this command
changes the way the `X-Registry-Config` header is sent when attempting
to build an image.

This fixes an issue I've experienced when using `docker-compose` and building
an image which is based off of a private image hosted at Docker Hub. I don't believe
this change is backwards compatible, and I'm not familiar enough with python to
have a solution for that. I'd appreciate some help if that is necessary.

[Here's the commit](https://github.com/docker/docker/commit/02c7bbefb8841e5e86a4b9d467defa668e3ea613) on the Docker project where they changed the way the AuthConfig
objects were being parsed for the build API endpoint.

### DIFF
--- a/docker/auth/auth.py
+++ b/docker/auth/auth.py
@@ -102,12 +102,6 @@ def encode_header(auth):
     return base64.b64encode(auth_json)
 
 
-def encode_full_header(auth):
-    """ Returns the given auth block encoded for the X-Registry-Config header.
-    """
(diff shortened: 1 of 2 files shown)


shin-:
I created #682 based off of this, with a change to support older versions of the API / Docker daemon. Feel free to review it! I will close this PR in favor of the new one as a result,

PR #680, opened 2015-07-20
Enforce UTC datetimes in arguments to `events()`

aanand:
I can't make the test suite pass on a machine that's not set to UTC, because the `datetime_to_timestamp()` utility function performs date arithmetic under that assumption, causing one of the `events()` tests to fail.

I've changed it so that it always expects a UTC datetime, and made sure we pass one in in the test.

I also removed the default value for its argument, because (a) it wasn't used anywhere and (b) it was being set to to `datetime.now()` at file load time, which almost certainly isn't what the user wants.

### DIFF
--- a/docker/utils/utils.py
+++ b/docker/utils/utils.py
@@ -333,9 +333,9 @@ def convert_filters(filters):
     return json.dumps(result)
 
 
-def datetime_to_timestamp(dt=datetime.now()):
-    """Convert a datetime in local timezone to a unix timestamp"""
-    delta = dt - datetime.fromtimestamp(0)
+def datetime_to_timestamp(dt):
+    """Convert a UTC datetime to a Unix timestamp"""
+    delta = dt - datetime.utcfromtimestamp(0)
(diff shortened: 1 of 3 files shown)


docker/utils/utils.py
```diff
@@ -333,9 +333,9 @@ def convert_filters(filters):
     return json.dumps(result)
 
 
-def datetime_to_timestamp(dt=datetime.now()):
-    """Convert a datetime in local timezone to a unix timestamp"""
-    delta = dt - datetime.fromtimestamp(0)
+def datetime_to_timestamp(dt=None):
+    """Convert a UTC datetime to a Unix timestamp"""
+    delta = dt - datetime.utcfromtimestamp(0)
```
shin-: This is going to break if the `dt` parameter is unspecified. Not that it ever should - I say we remove the default value altogether, what do you say?

aanand: For some reason I thought that's what I did, but it clearly isn't. Updated.

shin-:
Thanks, lgtm!

PR #681, opened 2015-07-20
Stop pinging registries from the client

aanand:
The daemon already pings the registry, so doing it on our end is redundant and error-prone.

- The `insecure_registry` argument to `push()`, `pull()` and `login()` has been deprecated - in the latter case, it wasn't being used anyway.

- The `insecure` argument to `docker.auth.resolve_repository_name()` has also been deprecated.

- `docker.utils.ping_registry()` has been deprecated.

- `docker.auth.expand_registry_url()` has been removed.

- `docker.auth.resolve_repository_name()` now returns a hostname as the first part of the tuple, rather than a full URL.

- When encountering a legacy `~/.dockercfg` with only authentication information for the Hub, `docker.auth.load_config()` now returns a dict with `index.docker.io` as its sole key, not `https://index.docker.io/v1/`.

### DIFF
--- a/docker/auth/auth.py
+++ b/docker/auth/auth.py
@@ -16,38 +16,33 @@ import base64
 import fileinput
 import json
 import os
+import warnings
 
 import six
 
-from ..utils import utils
 from .. import errors
 
-INDEX_URL = 'https://index.docker.io/v1/'
+INDEX_NAME = 'index.docker.io'
+INDEX_URL = 'https://{0}/v1/'.format(INDEX_NAME)
 DOCKER_CONFIG_FILENAME = os.path.join('.docker', 'config.json')
 LEGACY_DOCKER_CONFIG_FILENAME = '.dockercfg'
 
 
-def expand_registry_url(hostname, insecure=False):
-    if hostname.startswith('http:') or hostname.startswith('https:'):
-        return hostname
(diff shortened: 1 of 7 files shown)


docker/auth/auth.py
```diff
@@ -16,38 +16,33 @@
...
+import warnings
 
 import six
 
-from ..utils import utils
 from .. import errors
 
-INDEX_URL = 'https://index.docker.io/v1/'
+INDEX_NAME = 'index.docker.io'
+INDEX_URL = 'https://%s/v1/' % INDEX_NAME
```
shin-: Let's stay consistent across the board and use `format()` instead :)

docker/client.py
```diff
@@ -499,6 +499,13 @@ def load_image(self, data):
 
     def login(self, username, password=None, email=None, registry=None,
               reauth=False, insecure_registry=False, dockercfg_path=None):
+        if insecure_registry:
+            warnings.warn(
+                'The `insecure_registry` argument to login() '
+                'is deprecated and non-functional. Please remove it.',
```
shin-: Maybe store this message in `constants.py`

aanand:
Updated:

- Use `format` instead of `%`
- Store the deprecation warning in `constants.py` to avoid reuse

shin-:
Thanks!

PR #682, opened 2015-07-20
Switch to send full AuthConfig object for build action

shin-:
Follow-up to #677

### DIFF
--- a/docker/client.py
+++ b/docker/client.py
@@ -140,7 +140,7 @@ class Client(clientbase.ClientBase):
             if self._auth_configs:
                 if headers is None:
                     headers = {}
-                headers['X-Registry-Config'] = auth.encode_full_header(
+                headers['X-Registry-Config'] = auth.encode_header(
                     self._auth_configs
                 )
 
--- a/docker/client.py
+++ b/docker/client.py
@@ -140,9 +140,14 @@ class Client(clientbase.ClientBase):
             if self._auth_configs:
                 if headers is None:
                     headers = {}
-                headers['X-Registry-Config'] = auth.encode_header(
-                    self._auth_configs
-                )
+                if utils.compare_version('1.19', self._version) >= 0:
+                    headers['X-Registry-Config'] = auth.encode_header(
+                        self._auth_configs
+                    )
+                else:
+                    headers['X-Registry-Config'] = auth.encode_header({
+                        'configs': self._auth_configs
+                    })
 
         response = self._post(
             u,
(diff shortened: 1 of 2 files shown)


docker/client.py
```diff
@@ -140,9 +140,14 @@ def build(self, path=None, tag=None, quiet=False, fileobj=None,
             if self._auth_configs:
                 if headers is None:
                     headers = {}
-                headers['X-Registry-Config'] = auth.encode_full_header(
-                    self._auth_configs
-                )
+                if utils.compare_version('1.19', self._version) >= 0:
```
moutten: The only issue that this causes, for my particular situation, is that `docker-compose` would still be broke since it registers the [client API version as 1.18](https://github.com/docker/compose/blob/master/compose/cli/docker_client.py#L35). I don't know really how to detect what version the docker server is expecting. I'm not sure that an API version was changed when this authentication header was changed.

shin-: ping @aanand :3

moutten:
By the way, thanks @shin-, your solution is much better.

docker/client.py
aanand: Looks like the change went out in Docker 1.7.1, according to the milestone on this PR: https://github.com/docker/docker/pull/14139

I don't see any switching being done on the API version, so it's possible that backwards compatibility is broken. I've asked for clarification: https://github.com/docker/docker/pull/14139#issuecomment-123246739

shin-: From Josh's answer, seems like the change only affects 1.7.0 and was reverted in 1.7.1 - making this PR unnecessary. Users of 1.7.0 should upgrade to 1.7.1.

@moutten Can you confirm that this change is unnecessary with Docker 1.7.1 ?

moutten: This change was still necessary with 1.7.1. You'll see that the previous code was including a top level key of `configs`. Removing the top level key of `configs` works with 1.7.1. Based on Josh's comment, I'm not sure that was previously necessary. I haven't tried testing to see if removing the `configs` still works against older versions of docker (prior to 1.7).

moutten: I tested my setup using docker version 1.6.2, docker-compose 1.3.3, along with a patched version of docker-py, and I can verify it did work correctly with the change in this pull request. This looks like the cleanest way for docker-py to implement this change. It looks like I'm going to have to wait for docker-compose to change the supported API to 1.19.

shin-:
Thanks for following up so closely @moutten !

PR #686, opened 2015-07-21
Default to ~/.docker if DOCKER_CERT_PATH is empty

aanand:


thaJeztah:
Looks like this needs a rebase

docker/utils/utils.py
```diff
@@ -314,9 +314,15 @@ def kwargs_from_env(ssl_version=None, assert_hostname=None):
...
 
     params = {}
+
     if host:
         params['base_url'] = (host.replace('tcp://', 'https://')
                               if tls_verify else host)
+
+    if tls_verify and not cert_path:
+        if 'HOME' in os.environ:
+            cert_path = os.path.join(os.environ['HOME'], '.docker')
```
shin-: We should use the cross-platform compliant alternative `os.path.expanduser('~')` instead.

PR #691, opened 2015-07-23
Add netmode to tests

jhowardmsft:
Signed-off-by: John Howard <jhoward@microsoft.com>

@swernli @cpuguy83 

This PR relates to https://github.com/docker/docker/issues/10662 which is for the docker daemon port to Windows. Part of the daemon port requires the checking of the network mode setting to be moved from the CLI to the daemon, as network modes are platform specific. That PR in docker is https://github.com/docker/docker/pull/14530.

However, for CI on docker-py to pass in docker, it requires updates to docker-py to pass the netmode (which the docker CLI itself would normally do - just the tests here are generally written with the smallest amount of config settings necessary for validation). 

Hence this PR adds the network mode where required to the tests. After this is merged, I can update the Dockerfile in 14530 to point to the right docker-py commit.

Validation performed: I have validated docker-py with these changes passes "make binary test-docker-py" against docker/docker master as of this morning. I have also validated that docker-py running against docker with the 14530 changes passes Jenkins CI in docker.

### DIFF
--- a/docker/utils/utils.py
+++ b/docker/utils/utils.py
@@ -615,6 +615,16 @@ def create_container_config(
         if volumes_from is not None:
             raise errors.InvalidVersion(message.format('volumes_from'))
 
+    # NetworkMode must be present and valid in host config from 1.20 onwards
+    if compare_version('1.20', version) >= 0:
+        if host_config is None:
+            host_config = {'NetworkMode': 'default'}
+        else:
+            if 'NetworkMode' not in host_config:
+                host_config['NetworkMode'] = 'default'
+            elif host_config['NetworkMode'] == '':
+                host_config['NetworkMode'] = 'default'
+
     return {
         'Hostname': hostname,
         'Domainname': domainname,
--- a/tests/integration_test.py
+++ b/tests/integration_test.py
@@ -181,7 +181,8 @@ class TestCreateContainerWithBinds(BaseTestCase):
             container = self.client.create_container(
                 'busybox',
                 ['ls', mount_dest], volumes={mount_dest: {}},
-                host_config=create_host_config(binds=binds)
+                host_config=create_host_config(
+                    binds=binds, network_mode='none')
             )
             container_id = container['Id']
             self.client.start(container_id)
@@ -221,7 +222,8 @@ class TestCreateContainerWithRoBinds(BaseTestCase):
             container = self.client.create_container(
                 'busybox',
                 ['ls', mount_dest], volumes={mount_dest: {}},
-                host_config=create_host_config(binds=binds)
+                host_config=create_host_config(
+                    binds=binds, network_mode='none')
             )
             container_id = container['Id']
             self.client.start(container_id)
@@ -273,7 +275,8 @@ class TestCreateContainerReadOnlyFs(BaseTestCase):
     def runTest(self):
         ctnr = self.client.create_container(
             'busybox', ['mkdir', '/shrine'],
-            host_config=create_host_config(read_only=True)
+            host_config=create_host_config(
+                read_only=True, network_mode='none')
         )
         self.assertIn('Id', ctnr)
         self.tmp_containers.append(ctnr['Id'])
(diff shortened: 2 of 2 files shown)


jhowardmsft:
@shin- @mpetazzoni @aanand - Guys, any chance of getting a review on this? It's needed as a pre-req for a docker PR which we need for the Windows daemon to support the upcoming TP3 docker release. Thanks :)

docker/utils/utils.py
```diff
@@ -615,6 +615,16 @@ def create_container_config(
         if volumes_from is not None:
             raise errors.InvalidVersion(message.format('volumes_from'))
 
+    # NetworkMode must be present and valid in host config from 1.20 onwards
+    if compare_version('1.20', version) >= 0:
+        if host_config is None:
+            host_config = {'NetworkMode': 'none'}
```
shin-: Shouldn't it be `default` instead to maintain backwards compatibility?

tests/integration_test.py
```diff
@@ -181,7 +181,8 @@ def runTest(self):
             container = self.client.create_container(
                 'busybox',
                 ['ls', mount_dest], volumes={mount_dest: {}},
-                host_config=create_host_config(binds=binds)
+                host_config=create_host_config(
+                    binds=binds, network_mode='none')
```
shin-: Why are we setting this explicitly here if it's already handled in `create_container_config`?

docker/utils/utils.py
```diff
@@ -615,6 +615,16 @@ def create_container_config(
         if volumes_from is not None:
             raise errors.InvalidVersion(message.format('volumes_from'))
 
+    # NetworkMode must be present and valid in host config from 1.20 onwards
```
shin-: It feels like that snippet of code belongs to `create_host_config` instead. What's the rationale for having it here?

shin-:
Can you address those few comments? Thanks!

docker/utils/utils.py
jhowardmsft: Just looked at v1.6.0 docker\docker runconfig\parse.go, func parseNetMode L445. (Sorry, can't work out how to get a link directly to a previous version). The valid cases are bridge, none, host and container. Looks like default didn't arrive until later.

docker/utils/utils.py
jhowardmsft: Yup, you may well be right. I'd need to look at the flow more closely, but I believe there are cases where create_container_config is called without create_host_config. I'm not sure calling create_host_config from create_container_config is the right thing to do.

docker/utils/utils.py
shin-: But since the pre-condition here is that API version is >= 1.20, Docker 1.6 shouldn't be a concern.

docker/utils/utils.py
shin-: Well, the user can decide to provide their own dictionary for `host_config` instead of using `create_host_config`, but at that point they are making the conscious choice of not perusing the library's validation and transformation features the library provides, which I believe docker-py should let them do without tampering. AFAIK the tests always use `create_host_config`.

The only drawback is that `create_host_config` doesn't have access to the client's API version. Not sure we have a way to work around that.

docker/utils/utils.py
jhowardmsft: Using none is (IMO) cleaner as it pulls in "less" for the test on the daemon side (bearing in mind networking isn't needed for these tests).

docker/utils/utils.py
jhowardmsft: Hmm, I'm not sure what I should do here in that case :)

docker/utils/utils.py
shin-: Yeah, but that's the default value you're assigning not just for the tests but every other project using docker-py. It should be `default`.

jhowardmsft: Sure. I should have used none in the others lol. I'll update :)

tests/integration_test.py
jhowardmsft: Yes, I think you're right. Removing.

jhowardmsft:
@shin- I have it passing again, but I'm not sure how to address the remainder of your comments. (Python is new to me entirely).

shin-:
Thanks, that's helpful - I'll take it from here.

icecrime:
Ping @shin-: can we move this forward please?

shin-:
See #698

PR #692, opened 2015-07-24
Add cgroup_parent option.

### LINKED ISSUE #684: CgroupParent option not available in docker.utils.create_host_config() api
Docker has added capability to provide custom cgroup parent.
Currently the docker-py doesn't provide an option in docker.utils.create_host_config() api to specify the cgroup_parent.This would be required for those who want to customize the cgroup implementation.

francois2metz:
To:
 - utils.create_host_config
 - client.start

Should fix #684


shin-:
Thanks! Fixed via #716 :)

PR #693, opened 2015-07-24
[109] A verified fake Docker Client

wallrj:
I started writing this branch as material for a talk at EuroPython2015.
- https://ep2015.europython.eu/conference/talks/faking-it-the-art-of-testing-using-verified-fakes

I'd like to have submitted a PR before giving the talk but ran out of time.

Anyway, here's what I've written so far. 
Lot's more to do but probably enough to begin a discussion about the design of the verified fake.

The gist of it is:
- Add verified fake and associated test tools to `docker-py` and run tests against both the real and fake implementations.
- An additional benefit of having a `IDockerClient(zope.interface)` is that it can be an authoritative source for the docker-py API documentation. I'm sure, but haven't yet checked, that there's a sphinx plugin that can extract the RST docs from the interface and render them nicely.
- I'm not going to be able to submit a completely accurate fake in the first place;  In order to keep the branch manageable (and easily reviewable) I need to figure out what will be the minimum viable FakeDockerClient that will actually be useful. I'm thinking:
  - `containers(quiet=True, all=True)`
  - `create_container(image, name)`
  - `remove_container(container)`
  - `inspect_container(container)`

What do people think?

See also https://github.com/stackforge/nova-docker/blob/master/novadocker/tests/virt/docker/mock_client.py which seems to be an incomplete and unverified fake of `docker.client.Client`.


rmb938:
Is there any updates on getting this branch merged or something like it implemented? @shin- @wallrj

wallrj:
> Is there any updates on getting this branch merged or something like it implemented?

Sorry @rmb938, I'm useless and never finished it off. Not sure when I'm going to get around to it.

bfirsh:
There's now a mock fake client: https://github.com/docker/docker-py/blob/master/tests/unit/fake_api_client.py

Do you think something more elaborate than this is required?

PR #708, opened 2015-08-07
Split the entrypoint string to shell-like syntax.

ColinHuang:
If not doing shlex for ```entrypoint``` parameter. 
Jenkin image cannot start successful.


aanand:
Thanks. The tests need fixing, and it'd be good to have a test that explicitly checks the new behaviour.

ColinHuang:
Fixed test case for ```entrypoint``` new  behaviour.

aanand:
LGTM

linlinlinlin:
LGTM

shin-:
There's a lot of commits that shouldn't be here - can you try and isolate your changes and submit a clean PR? Thanks :)

PR #713, opened 2015-08-13
Add GELF logging support

dkerwin:
Trivial change to add gelf logging support (docker 1.8).

Cheers!
Daniel


shin-:
Thanks! #724

PR #716, opened 2015-08-16
Added support for cgroup parent

mohitsoni:
This change adds cgroup-parent support to docker-py, which will allow us to add cgroup-parent support inside compose.

### DIFF
--- a/docker/client.py
+++ b/docker/client.py
@@ -225,7 +225,8 @@ class Client(clientbase.ClientBase):
                          network_disabled=False, name=None, entrypoint=None,
                          cpu_shares=None, working_dir=None, domainname=None,
                          memswap_limit=None, cpuset=None, host_config=None,
-                         mac_address=None, labels=None, volume_driver=None):
+                         mac_address=None, labels=None, volume_driver=None,
+                         cgroup_parent=None):
 
         if isinstance(volumes, six.string_types):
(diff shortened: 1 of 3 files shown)


aanand:
LGTM

docker/client.py
```diff
@@ -737,7 +738,7 @@ def start(self, container, binds=None, port_bindings=None, lxc_conf=None,
               dns=None, dns_search=None, volumes_from=None, network_mode=None,
               restart_policy=None, cap_add=None, cap_drop=None, devices=None,
               extra_hosts=None, read_only=None, pid_mode=None, ipc_mode=None,
-              security_opt=None, ulimits=None):
+              security_opt=None, ulimits=None, cgroup_parent=None):
```
shin-: I'd argue against adding this - we've deprecated passing host config arguments in `start`, and will probably remove it entirely in a future version.

mohitsoni: @shin- Good to know. Can you please tell how ```cgroup_parent``` and other host config arguments be passed ?

shin-: Gladly! You can see the [documentation](http://docker-py.readthedocs.org/en/latest/hostconfig/) on that topic.

shin-:
Applied a few fixes and merged manually. Thanks!

PR #718, opened 2015-08-17
Remove hard coded list of log driver types

mnowster:
This removes the hardcoded list of log drivers we allow as valid. Instead, let those values get passed to the daemon that already does it's own validation of whether a log driver is supported/valid or not.

This removes a bottleneck to supporting new/more log drivers as they get added into docker and the duplication of validation.

This PR will allow for this issue https://github.com/docker/compose/issues/1869 to be resolved.

### DIFF
--- a/tests/integration_test.py
+++ b/tests/integration_test.py
@@ -245,30 +245,25 @@ class TestCreateContainerWithRoBinds(BaseTestCase):
         self.assertFalse(inspect_data['VolumesRW'][mount_dest])
 
 
-@unittest.skipIf(NOT_ON_HOST, 'Tests running inside a container; no syslog')
-class TestCreateContainerWithLogConfig(BaseTestCase):
-    def runTest(self):
-        config = docker.utils.LogConfig(
-            type=docker.utils.LogConfig.types.SYSLOG,
-            config={'key1': 'val1'}
+class CreateContainerWithLogConfigTest(BaseTestCase):
+    def test_valid_log_driver_and_log_opt(self):
+        log_config = docker.utils.LogConfig(
+            type='json-file',
+            config={'max-file': '100'}
         )
-        ctnr = self.client.create_container(
+
+        container = self.client.create_container(
             'busybox', ['true'],
             host_config=self.client.create_host_config(log_config=config)
         )
-        self.assertIn('Id', ctnr)
-        self.tmp_containers.append(ctnr['Id'])
-        self.client.start(ctnr)
-        info = self.client.inspect_container(ctnr)
-        self.assertIn('HostConfig', info)
-        host_config = info['HostConfig']
-        self.assertIn('LogConfig', host_config)
-        log_config = host_config['LogConfig']
-        self.assertIn('Type', log_config)
-        self.assertEqual(log_config['Type'], config.type)
-        self.assertIn('Config', log_config)
-        self.assertEqual(type(log_config['Config']), dict)
-        self.assertEqual(log_config['Config'], config.config)
+        self.tmp_containers.append(container['Id'])
+        self.client.start(container)
+
+        info = self.client.inspect_container(container)
+        container_log_config = info['HostConfig']['LogConfig']
+
+        self.assertEqual(container_log_config['Type'], log_config.type)
+        self.assertEqual(container_log_config['Config'], log_config.config)
 
 
 @unittest.skipIf(not EXEC_DRIVER_IS_NATIVE, 'Exec driver not native')
--- a/tests/integration_test.py
+++ b/tests/integration_test.py
@@ -255,7 +255,7 @@ class CreateContainerWithLogConfigTest(BaseTestCase):
 
         container = self.client.create_container(
             'busybox', ['true'],
-            host_config=self.client.create_host_config(log_config=config)
+            host_config=self.client.create_host_config(log_config=log_config)
         )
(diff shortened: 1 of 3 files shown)


tests/integration_test.py
```diff
@@ -246,30 +247,41 @@ def runTest(self):
...
-        self.assertEqual(log_config['Config'], config.config)
+        self.tmp_containers.append(container['Id'])
+        self.client.start(container)
+
+        info = self.client.inspect_container(container)
+        container_log_config = info['HostConfig']['LogConfig']
+
+        self.assertEqual(container_log_config['Type'], log_config.type)
+        self.assertEqual(container_log_config['Config'], log_config.config)
+
```
aanand: Perhaps we should also test the "no type arg" and "no config arg" cases.

mnowster: Included extra tests.

shin-:
I feel there's value in having the user be able to use LogConfig.types.FLUENTD instead of providing an error-prone string. Maybe we can do away with the validation...

aanand:
For the sake of backwards compatibility, we should arguably at least keep the existing constants around; there's a case to be made for adding GELF and FLUENTD for consistency too.

However, validating it on the client side is a huge pain. It's too much of a blocker to have to add a new logging driver to docker-py's whitelist, then update Compose's docker-py version (bringing in stuff we might not want to deal with yet), then release a new version, just so Compose users can use it.

mnowster:
From a design stand point I also don't think it's docker-py's responsibility to do the duplicated effort of maintaining the list of supported drivers. Separation of concerns, I view that as docker daemon's responsibility.

That's a good point @aanand about keeping the types for backwards compatibility, I'll update to include them.

aanand:
LGTM

tests/integration_test.py
```diff
@@ -246,30 +247,80 @@ def runTest(self):
         self.assertFalse(inspect_data['VolumesRW'][mount_dest])
 
 
-@unittest.skipIf(NOT_ON_HOST, 'Tests running inside a container; no syslog')
-class TestCreateContainerWithLogConfig(BaseTestCase):
-    def runTest(self):
-        config = docker.utils.LogConfig(
-            type=docker.utils.LogConfig.types.SYSLOG,
-            config={'key1': 'val1'}
+class CreateContainerWithLogConfigTest(BaseTestCase):
```
shin-: Let's stay consistent with the other test class names, please.

tests/integration_test.py
```diff
@@ -246,30 +247,80 @@ def runTest(self):
         self.assertFalse(inspect_data['VolumesRW'][mount_dest])
 
 
-@unittest.skipIf(NOT_ON_HOST, 'Tests running inside a container; no syslog')
```
shin-: Is it safe to remove this? Have you tried running `make integration-test` ?

tests/utils_test.py
```diff
@@ -185,25 +185,19 @@ def test_ulimit_invalid_type(self):
...
-        obj = LogConfig(type=LogConfig.types.SYSLOG, config={'key1': 'val1'})
+        obj = LogConfig(type='syslog', config={'key1': 'val1'})
         config = create_host_config(log_config=obj)
         self.assertIn('LogConfig', config)
         self.assertTrue(isinstance(config['LogConfig'], LogConfig))
         self.assertEqual(obj, config['LogConfig'])
 
-    def test_logconfig_invalid_type(self):
-        self.assertRaises(ValueError, lambda: LogConfig(type='xxx', config={}))
-        self.assertRaises(ValueError, lambda: LogConfig(
```
shin-: It doesn't hurt to continue testing that one IMO

shin-:
Just a few comments, looks pretty good overall :+1:

tests/utils_test.py
mnowster: The validation code that would have raised a ValueError has been removed, thus making this test redundant. The tests added in integration_tests cover the use case of invalid config types and check that we receive an error back from the API.

tests/integration_test.py
mnowster: Yep, I think it is. I ran this test on it's own. I did also run make integration-test but get some time out errors which I believe aren't due to this test. Did you see the commit that goes with this code change? It describes the why of doing this. By using `json-file` as the use case we can run this without needing to be on a machine with syslog installed.

tests/integration_test.py
mnowster: This test is written in a convention that echoes that of docker-compose, https://github.com/docker/compose/blob/master/tests/unit/config_test.py, which is a good guide of how tests are to be structured and written.

Test case names are easier to read without Test at the beginning, we know it's a test, so we can put it at the end. Naming our tests individually rather than runTest provides helpful feedback when a test fails or we get a stack trace error. This is a way of improving the code while I was there and a demonstration of what the tests can be refactored towards. 

Consistency for it's own sake forsaking that of improved code is not as beneficial.

aanand: While it's not consistent, I think @mnowster's changes are improvements, so I'm fine with it. Would be good to get the rest of the suite updated along similar lines.

mnowster:
Ok, I've rebased and responded to comments. 

Anything else or we good for merging? @shin- @aanand

aanand:
I like it :+1:

shin-:
Thanks.

PR #723, opened 2015-08-18
create_host_config now needs version

### LINKED ISSUE #714: docker-py version "1.4.0-dev" network_mode bug with docker 1.7.1
With the introduction of [this code](https://github.com/docker/docker-py/blame/master/docker/utils/utils.py#L432) to cater for docker 1.8's `network_mode` option of `default`, it now breaks the integration tests if you are on docker 1.7 with:

`NotFound: 404 Client Error: Not Found ("Cannot start container 1a0b16410775af68692e174462b94c8ebd9e79be31da27d3b870b63dd83d2625: error locating network with name default: network default not found")`

I propose that we add in a `compare_version` around the setting of the `network_mode` and if the API version < 1.20 we don't set it to `default`, so that we are backwards compatible for docker 1.7.

mnowster:
When network_mode was introduced to have a default of 'default'
it broke backwards compatibility for API versions < 1.20

To fix this, we need to check a version, to do that, we need the
version passed in. To ensure this fix doesn't break backwards
compatibility itself, we use the DEFAULT_DOCKER_API_VERSION as
a default if version isn't passed in.

Fixes #714

### DIFF
--- a/docker/client.py
+++ b/docker/client.py
@@ -794,6 +794,7 @@ class Client(clientbase.ClientBase):
                     DeprecationWarning
                 )
             start_config = utils.create_host_config(**start_config_kwargs)
+            start_config['version'] = self._version
 
         url = self._url("/containers/{0}/start".format(container))
         res = self._post_json(url, data=start_config)
--- a/docker/utils/utils.py
+++ b/docker/utils/utils.py
@@ -27,6 +27,7 @@ from datetime import datetime
 import requests
 import six
 
+from .. constants import DEFAULT_DOCKER_API_VERSION
 from .. import errors
 from .. import tls
 from .types import Ulimit, LogConfig
@@ -389,6 +390,7 @@ def parse_bytes(s):
 
 
 def create_host_config(
+    version=DEFAULT_DOCKER_API_VERSION,
     binds=None, port_bindings=None, lxc_conf=None,
     publish_all_ports=False, links=None, privileged=False,
     dns=None, dns_search=None, volumes_from=None, network_mode=None,
@@ -434,7 +436,8 @@ def create_host_config(
     if network_mode:
         host_config['NetworkMode'] = network_mode
     elif network_mode is None:
-        host_config['NetworkMode'] = 'default'
+        if compare_version('1.19', version) > 0:
+            host_config['NetworkMode'] = 'default'
 
     if restart_policy:
         host_config['RestartPolicy'] = restart_policy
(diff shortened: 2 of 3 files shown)


shin-:
The problem I have with that is the fact that in most cases, this method will be called directly by the API consumer, which makes the version detection process very error-prone. I don't disagree that it needs a solution, but I'm unconvinced this is the right approach to it.

mnowster:
I appreciate I'm not as familiar with this codebase, this was the only approach I could see how to do and saw that version checking was used elsewhere in the codebase. 

Is there an alternative approach I've missed? Do you have an idea how else this could be done?

If not, is this good enough for now as it's at least some help towards backwards compatibility for users which is an improvement on not having anything which gives the error of `default` not found for `network_mode`

aanand:
For consistency, I think an ideal scenario is that `create_host_config` *requires* a `version` arg, much like `create_container_config`.

However, this will break things for many current users of docker-py, who are probably doing things like this:

```python
from docker import Client
from docker.utils import create_host_config

client = Client()
container = client.create_container(
    'busybox',
    'true',
    host_config=create_host_config(privileged=True),
)
```

So how about this: we define new `create_container_config` and `create_host_config` methods on `Client` (which can always set the version correctly via `self._version`), and deprecate the direct use of their namesakes in `docker.utils`.

The above snippet would continue to work exactly as before, but a deprecation warning would be printed. Something like "docker.utils.create_host_config() is deprecated. Please use client.create_host_config() instead".

This would be the new idiomatic approach:

```python
from docker import Client

client = Client()
container = client.create_container(
    'busybox',
    'true',
    host_config=client.create_host_config(privileged=True),
)
```

As part of doing this, we have two additional opportunities for improvement:

1. Move the methods to a module with a more descriptive name - say `docker.config`. We can then replace the ones in `docker.utils` with wrappers that print the warning and then pass through.

2. Move all the API version checks currently done in `start` into `client.create_host_config()`, and eventually into `docker.config.create_host_config()` after a sufficient deprecation period.

Thoughts?

mnowster:
@aanand this approach sounds good to me. I like the 2 step approach so we can display deprecation warnings then remove it, that'll be helpful to not screw over users who are using the direct approach.

I think this will also help bring some clarity to the intent of some of the code. There's a lot in the utils file which isn't utility functions it's core logic, so to get to move this around would be a positive step in my view.

aanand:
@shin- would you be happy with the above approach? This bug's a bit of a showstopper.

shin-:
Sorry - yeah, that seems to be the right solution for this. It rubs me the wrong way that we're adding still more stuff to the `Client` object that already feels overcrowded, but that's a minor thing, it shouldn't block us.

mnowster:
Closing as this has been resolved in https://github.com/docker/docker-py/pull/732

PR #724, opened 2015-08-18
Update types.py to support fluentd logging driver

bedwards:
docker 1.8 now supports a fluentd logging driver - https://docs.docker.com/reference/logging/fluentd/?utm_content=buffer61fdd&utm_medium=social&utm_source=twitter.com&utm_campaign=buffer

### DIFF
--- a/docker/utils/types.py
+++ b/docker/utils/types.py
@@ -6,9 +6,11 @@ class LogConfigTypesEnum(object):
         'json-file',
         'syslog',
         'journald',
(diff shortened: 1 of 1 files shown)


shin-:
Thanks, could you squash those 2 commits?

bedwards:
added gelf too - https://docs.docker.com/reference/logging/overview/

shin-:
Thanks!

PR #728, opened 2015-08-20
Add logging for auth loading/resolution

aanand:
Some Compose users still report authentication problems, and they're very hard to debug (because *any* problem locating/resolving auth credentials results in silent failure).

To make it easier, I've added a bunch of debug logging to the `auth` module, and to `client.build()` and `client.pull()`, which can be enabled by any user of docker-py with a log handler set to the DEBUG level.

One thing has changed about the semantics of `load_config`: there were a couple of `try/except` blocks with an empty `except` clause. These would have caught `KeyboardInterrupt` and `SystemExit`, which is generally a bad idea. They now explicitly catch `Exception`, meaning those two exception types will still propagate. That's arguably still too broad, but it's an improvement.

### DIFF
--- a/docker/auth/auth.py
+++ b/docker/auth/auth.py
@@ -15,6 +15,7 @@
 import base64
 import fileinput
 import json
+import logging
 import os
 import warnings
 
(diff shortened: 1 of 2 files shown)


shin-:
LGTM!

PR #729, opened 2015-08-20
WIP: Networking support

aanand:
A whole bunch of new `Client` methods which bind to the [experimental networking API](https://github.com/docker/docker/blob/master/experimental/networking_api.md) (here's an [overview of the functionality it offers](https://github.com/docker/docker/blob/master/experimental/networking.md)).

### DIFF
--- a/docker/api/network.py
+++ b/docker/api/network.py
@@ -1,0 +1,49 @@
+import json
+
+from ..utils import check_resource
+
+
+class NetworkApiMixin(object):
+    def networks(self, names=None, ids=None):
+        filters = {}
+        if names:
+            filters['name'] = names
+        if ids:
+            filters['id'] = ids
+
+        params = {'filters': json.dumps(filters)}
+
+        url = self._url("/networks")
+        res = self._get(url, params=params)
+        return self._result(res, json=True)
+
+    def create_network(self, name, driver=None):
+        data = {
+            'name': name,
+            'driver': driver,
+        }
+        url = self._url("/networks/create")
+        res = self._post_json(url, data=data)
+        return self._result(res, json=True)
+
+    def remove_network(self, net_id):
+        url = self._url("/networks/{0}", net_id)
+        res = self._delete(url)
+        self._raise_for_status(res)
+
+    def inspect_network(self, net_id):
+        url = self._url("/networks/{0}", net_id)
+        res = self._get(url)
+        return self._result(res, json=True)
+
+    @check_resource
+    def connect_container_to_network(self, container, net_id):
+        data = {"container": container}
+        url = self._url("/networks/{0}/connect", net_id)
+        self._post_json(url, data=data)
+
+    @check_resource
+    def disconnect_container_from_network(self, container, net_id):
+        data = {"container": container}
+        url = self._url("/networks/{0}/disconnect", net_id)
+        self._post_json(url, data=data)
--- a/docker/api/__init__.py
+++ b/docker/api/__init__.py
@@ -5,3 +5,4 @@ from .daemon import DaemonApiMixin
 from .exec_api import ExecApiMixin
 from .image import ImageApiMixin
 from .volume import VolumeApiMixin
+from .network import NetworkApiMixin
--- a/docker/client.py
+++ b/docker/client.py
@@ -39,7 +39,8 @@ class Client(
         api.DaemonApiMixin,
         api.ExecApiMixin,
         api.ImageApiMixin,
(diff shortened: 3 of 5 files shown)


aanand:
This is ready for more eyes.

The networking API doesn't appear to be present in Docker master (which is strange - @mavenugo, @dave-tucker, is that expected? When should it land?), so to test it you'll need to install experimental Docker. If not *straightforward* per se, it's at least fairly doable with Docker Machine:

    docker-machine create -d virtualbox experimental
    docker-machine ssh experimental 'sudo sh -c "set -ex; /etc/init.d/docker stop || true; sleep 2; curl https://experimental.docker.com/builds/Linux/x86_64/docker-1.9.0-dev > /usr/local/bin/docker; /etc/init.d/docker start"'
    eval "$(docker-machine env experimental)"

Then update `DEFAULT_DOCKER_API_VERSION` to `'1.21'` in `docker/constants.py`.

Finally, run the tests:

    make unit-test
    make integration-test

There are unrelated failures in both test suites, caused by the version update.

mavenugo:
@aanand yes. the API has not landed into docker proper yet. It is still under experimental. We are waiting on a few fixes to be merged before moving the API/UX out of experimental.

docker/api/network.py
```diff
@@ -0,0 +1,29 @@
...
+            'name': name,
+            'network_type': '',
+            'options': {},
+        }
+        url = self._url("/networks")
+        res = self._post_json(url, data=data)
+        return self._result(res, json=True)
+
+    def remove_network(self, net_id):
+        url = self._url("/networks/{0}".format(net_id))
```
shin-: use `self._url('/networks/{0}', resource_id=net_id)` instead to avoid URL tampering

docker/api/network.py
```diff
@@ -0,0 +1,29 @@
...
+        res = self._delete(url)
+        self._raise_for_status(res)
+
+    def inspect_network(self, net_id):
+        url = self._url("/networks/{0}".format(net_id))
```
shin-: ditto

docker/api/service.py
```diff
@@ -0,0 +1,62 @@
...
+        data = {
+            'name': name,
+            'network_name': network_name,
+        }
+        url = self._url("/services")
+        res = self._post_json(url, data=data)
+        return self._result(res, json=True)
+
+    def unpublish_service(self, service_id):
+        url = self._url("/services/{0}".format(service_id))
```
shin-: same thing about the use of format for this and the few other occurrences below

docker/api/service.py
```diff
@@ -0,0 +1,62 @@
...
+
+    def service_backends(self, service_id):
+        url = self._url("/services/{0}/backend".format(service_id))
+        res = self._get(url)
+        return self._result(res, json=True)
+
+    def sandboxes(self, partial_container_id=None):
+        url = self._url("/sandboxes")
+        params = {}
+        if partial_container_id:
```
shin-: `params = { 'partial-container-id': partial_container_id }` would do the same thing here in practice.

docker/api/service.py
```diff
@@ -0,0 +1,62 @@
...
+            params['partial-container-id'] = partial_container_id
+        res = self._get(url, params=params)
+        return self._result(res, json=True) or []
+
+    @check_resource
+    def attach_container_to_service(self, container, service_id):
+        sandbox = self.sandboxes(partial_container_id=container)[0]
```
shin-: That's a bit odd... Shouldn't the API do that kind of lookup rather than the client?

shin-: The other aspect is having 2 API calls in one method might break user expectations.

aanand: ping @mavenugo - I'm not familiar enough with the sandbox/container distinction

I can see a case for splitting it out in docker-py, yes - the combined logic would then live in Compose

mavenugo: the apis are being simplified and as @shin- indicated, there will be a /connect api with container as the resource of interest. sandbox is a backend detail and will not be exposed in the API.

This API that @aanand is using is based on experimental version and will be changed soon.
I will add an update here once the docker/docker API/UX change PR is pushed.

aanand: OK, in that case I'll hold off on changing anything as it sounds like docker-py's API won't change.

aanand:
This is now working against the latest iteration of the networking API, as implemented in https://github.com/docker/docker/pull/16645.

shin-:
:+1:

PR #732, opened 2015-08-24
Version dependent hostconfig

shin-:
@aanand Following the discussion in #723


aanand:
LGTM

bboreham:
Getting a bunch of failures from the integration test now - doesn't it need to be updated to cope with this change?

PR #734, opened 2015-08-25
Split the entrypoint string to shell-like syntax.

ColinHuang:
If not doing shlex for  ```entrypoint```  parameter. 
 Jenkin image cannot start successful.


aanand:
Thanks for the cleanup!

LGTM

PR #735, opened 2015-08-25
Explicitly test NetworkMode logic in create_host_config

aanand:
Ensure that we're testing both sides of the version check.


aanand:
Pushed another commit to stop people passing `version` into `create_host_config`, since it'll be overridden (and it makes no sense).

shin-:
Thanks!

PR #736, opened 2015-08-25
Stop catching Exception in tests where it isn't expected

aanand:
Unless I've misunderstood something fundamental, the test runner will report on the exception anyway, so catching it and explicitly calling `self.fail` is redundant.


shin-:
LGTM!

PR #739, opened 2015-08-25
Use pytest in unit tests

aanand:
Sits on top of #736.

Among other things, this makes asserting about exceptions and deprecation warnings much less verbose.


shin-:
:+1:

PR #721, opened 2015-08-18
Better support for .dockerignore

aanand:
Builds on @mikexstudios' work in #604, with fixes for Python 2.6 and some refactoring.

Still waiting on version 0.3.4 of pathspec, though - any word on that, @cpburnz?

PR #741, opened 2015-08-25
Run dockerignore integration test with nocache enabled

### LINKED ISSUE #712: Integration test fails on second run with Docker 1.8
Specifically, `TestBuildWithDockerignore`.

The second time docker is using the cache, so the output does not include names of the added files. I believe this may be because the docker 1.8 build cache ignores mtime.

Could either clear the cache beforehand, or run the container and check it's output (instead of the build output).

shin-:
Fixes #712

### DIFF
--- a/tests/integration_test.py
+++ b/tests/integration_test.py
@@ -1322,7 +1322,7 @@ class TestBuildWithDockerignore(Cleanup, BaseTestCase):
         with open(os.path.join(subdir, 'grunt'), 'w') as f:
             f.write("grunt")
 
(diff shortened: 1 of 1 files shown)


aanand:
I'm a bit afraid of making assertions about the whole of the build stream output. I took a different approach in 380d279bf04ffcb5c414bc5893f4aa5f27f91fad, which I rolled into #721 - what do you think? I could submit it as an isolated PR.

shin-:
Looks like you removed it from #721 - But your solution looks good. Feel free to merge it in.

PR #743, opened 2015-08-27
WIP: Run integration tests on Travis

aanand:
Hopefully we can make use of their Docker support. If not, we should consider moving to Docker's CI infrastructure.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,10 @@
-sudo: false
+sudo: required
 language: python
+services:
+    - docker
 python:
-  - "2.7"
+    - "2.7"
+
 env:
     - TOX_ENV=py26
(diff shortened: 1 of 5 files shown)


aanand:
Getting a lot of `APIError: 500 Server Error: Internal Server Error ("Cannot kill container <long id string>: [8] System error: permission denied")`

Only related issue I can find is https://github.com/docker/docker/issues/15101, which implies it's something to do with AppArmor and suggests that something's up with the Ubuntu installation. Maybe Travis isn't going to work for us.

aanand:
I've separated the integration test improvements out into https://github.com/docker/docker-py/pull/747.

aanand:
Closing as we're going to move to Jenkins.

PR #746, opened 2015-08-27
Put the pre-test code in a pytest fixture

### LINKED ISSUE #745: integration_test.py assumes busybox image is present
It used to do a `pull` near the bottom of the file, but now you've changed the way the script gets run, that code is no longer executed.

bboreham:
I used the pattern suggested at http://stackoverflow.com/questions/19225279/pytest-init-setup-for-few-modules; I am no Python expert so apologies if this is wrong.

Fixes #745

### DIFF
--- a/tests/integration_test.py
+++ b/tests/integration_test.py
@@ -27,6 +27,7 @@ import time
 import unittest
 import warnings
 
+import pytest
 import docker
 import six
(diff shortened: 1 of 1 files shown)


bboreham:
Obsoleted by #747

PR #747, opened 2015-08-27
Integration test improvements

aanand:
- Fix the setup step (the process has changed now that we're using pytest)

- Skip tests that are unstable

- Make better use of build caching

### DIFF
--- a/tests/integration_test.py
+++ b/tests/integration_test.py
@@ -28,6 +28,7 @@ import unittest
 import warnings
 
 import docker
+from docker.utils import kwargs_from_env
 import six
 
 from six.moves import BaseHTTPServer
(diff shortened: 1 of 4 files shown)


tests/integration_test.py
```diff
@@ -36,15 +37,40 @@
...
 warnings.simplefilter('error')
 compare_version = docker.utils.compare_version
 
+_exec_driver = []
+
+
+def exec_driver_is_native():
+    if not _exec_driver:
+        c = docker_client()
+        _exec_driver.append(c.info()['ExecutionDriver'])
```
shin-: Why are we using a list to store a single value? Am I missing something?

tests/integration_test.py
```diff
@@ -322,9 +354,11 @@ def test_valid_no_config_specified(self):
         self.assertEqual(container_log_config['Config'], {})
 
 
-@unittest.skipIf(not EXEC_DRIVER_IS_NATIVE, 'Exec driver not native')
 class TestCreateContainerReadOnlyFs(BaseTestCase):
     def runTest(self):
+        if not exec_driver_is_native():
+            pytest.skip('Exec driver not native')
```
shin-: Is there a reason we don't use the `@pytest.mark.skipif` decorator here as well?

aanand: Yes - we can't evaluate `exec_driver_is_native()` at script execution time, because the check is done in `setup_module()` (as per pytest convention), which hasn't run yet.

tests/integration_test.py
aanand: A hack caused by my misunderstanding of what features existed in Python 2 and 3. Updated to use `global`.

PR #749, opened 2015-08-28
PyPy support

funkyfuture:
this adds tests against PyPy2 and PyPy3.

i couldn't test locally due to networking problems, but once already tested successfully against both interpreters. let's see what travis says.

sits on top of #747


funkyfuture:
`make test` also passes.

shin-:
So, merging that PR comes with the expectation that we'll support Pypy specific issues in the future. I'm not sure I want to commit to that.

Of course, that doesn't mean that we'll intentionally break it, or that we won't make a reasonable effort to help fix it if it's broken.

funkyfuture:
of course one could explicitly formulate in the README that there's no guarantee that PyPy will be supported, but efforts are made (and thus it should be tested).

i wouldn't anticipate any changes to a rest-adapter that still could break compatibility. you'll mainly extend the interfaces according to Docker's evolvement, right?

funkyfuture:
afaik, compatibility issues with PyPy are expected when using C-extensions and the `inspect`-module.

funkyfuture:
rebased on master and the tests still pass with PyPy, i added a config that allows them to fail on travis anyway.

PR #751, opened 2015-08-31
Support to pass optional args to adapter

wtakase:
This patch enables to pass optional args to UnixAdapter and SSLAdapter.
For example, increasing `maxsize` in UnixHTTPConnectionPool leads to
suppress `Connection pool is full, discarding connection` warnings.

Signed-off-by: Wataru Takase wataru.takase@kek.jp


aanand:
I see the use case, but catching every unrecognised kwarg to `Client()` and passing it to the adapter seems extreme - it'll make it hard to add kwargs in future, because of potential name collisions.

I think we should either:

1. Just add a `max_connections` argument for now, which passes through to the adapter
2. Add a named `adapter_kwargs` argument
3. Allow the user to construct their own adapter and pass it in

PR #753, opened 2015-08-31
Always send a string for the user param in create_container

### LINKED ISSUE #715: Slightly incorrect documentation for user parameter in create_container?
The documentation for `user` parameter in `create_container` says: 

`user (str or int): Username or UID`

However, supplying it as python's int(`client.create_container(user=1000, ...)`) gives 
```docker.errors.APIError: 500 Server Error: Internal Server Error ("json: cannot unmarshal number into Go value of type string")```

Supplying it as string works. So this code `client.create_container(user="1000", ...)` works fine. 

I guess it is a minor issue, though only because Celery printed the exception as "500 Server Error: Internal Server Error", so the critical piece of information was missing(`json: cannot unmarshal number into Go value of type string`). So I guess it is a minor issue, but I figured I should report it.

### PR DESCRIPTION
**Pull request by shin-**
Fixes #715

### DIFF
--- a/docker/utils/utils.py
+++ b/docker/utils/utils.py
@@ -715,7 +715,7 @@ def create_container_config(
         'Hostname': hostname,
         'Domainname': domainname,
         'ExposedPorts': ports,
-        'User': '{0}'.format(user) if user else None,
+        'User': six.text_type(user) if user else None,
         'Tty': tty,
         'OpenStdin': stdin_open,
         'StdinOnce': stdin_once,
--- a/docker/utils/utils.py
+++ b/docker/utils/utils.py
@@ -715,7 +715,7 @@ def create_container_config(
         'Hostname': hostname,
         'Domainname': domainname,
         'ExposedPorts': ports,
-        'User': user,
+        'User': '{0}'.format(user) if user else None,
         'Tty': tty,
         'OpenStdin': stdin_open,
         'StdinOnce': stdin_once,
--- a/tests/integration_test.py
+++ b/tests/integration_test.py
@@ -1626,4 +1626,7 @@ class TestRegressions(BaseTestCase):
         self.client.stop(ctnr)
 
     def test_715(self):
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

File: docker/utils/utils.py [hunk 1]
```diff
@@ -715,7 +715,7 @@ def create_container_config(
         'Hostname': hostname,
         'Domainname': domainname,
         'ExposedPorts': ports,
-        'User': user,
+        'User': '{0}'.format(user) if user else None,
```
aanand: Feels a bit hacky - could we just say `six.text_type(user)`?

shin-: Fair enough, I didn't want to use `str()` and didn't bother to look what was available in `six` :)

tests/integration_test.py
```diff
@@ -1624,3 +1624,6 @@ def test_649(self):
         ctnr = self.client.create_container('busybox', ['sleep', '2'])
         self.client.start(ctnr)
         self.client.stop(ctnr)
+
+    def test_715(self):
```
aanand: Could we give this test a more descriptive name? Also, we should assert that the container came out with the right uid.

shin-: It's part of the TestRegressions class, which all simply reference the issue number they're testing for. But yeah, I'll add the test for the correctness of the user ID.

aanand: OK... I'm not sure about that as a general thing. It places a rather unnecessary demand on people reading the code. Out of scope for this PR, but maybe we could rename them with descriptive names and include the URL to the GitHub issue in a docstring?

shin-: Sure. My reasoning was that those tests won't be read unless there's an issue (regression) happening, in which case the person examining it will probably go to the original issue anyway. If you think there's value in renaming them that's fine too.

dnephin: I'm also a fan of having a descriptive name and linking to the issue in the git commit

dnephin:
LGTM, agree we can adjust test names in followup

PR #754, opened 2015-09-01
Default to 127.0.0.1:2375 on Windows

aanand:
Following the logic of the Docker client.

### DIFF
--- a/docker/utils/utils.py
+++ b/docker/utils/utils.py
@@ -272,11 +272,15 @@ def parse_repository_tag(repo):
 # fd:// protocol unsupported (for obvious reasons)
 # Added support for http and https
 # Protocol translation: tcp -> http, unix -> http+unix
-def parse_host(addr):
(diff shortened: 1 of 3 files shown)


docker/utils/utils.py
```diff
@@ -272,11 +272,15 @@ def parse_repository_tag(repo):
 # fd:// protocol unsupported (for obvious reasons)
 # Added support for http and https
 # Protocol translation: tcp -> http, unix -> http+unix
-def parse_host(addr):
+def parse_host(addr, platform):
```
shin-: Couldn't we make `platform` optional?

aanand: Ah, I didn't realise it was public API, so I guess we have to. Will fix.

aanand: Done.

mpetazzoni:
:+1:

PR #755, opened 2015-09-03
Add @dnephin and @mnowster as maintainers

aanand:
@dnephin I'm not actually sure what the right email to use for you is. I've gone with the same one that's in your commits, which is consistent with [MAINTAINERS in docker/compose](https://github.com/docker/compose/blob/master/MAINTAINERS).


dnephin:
Either that one or `dnephin@docker.com` works for me. I guess I should update my gitconfig

shin-:
Perfect - I was gonna ask yesterday but I forgot. LGTM :+1:

Issue #756, opened 2015-09-08
Integration tests should run on CI


dmp42:

shin-:
That's happening now - some improvements can and will be made in the near future so that they hopefully run faster than they are currently.

PR #757, opened 2015-09-09
Containerized Integration

shin-:
Needed for #756 - @aanand @dnephin Please take a look!


tests/Dockerfile
```diff
@@ -0,0 +1,19 @@
+FROM dockerswarm/dind:1.8.1
+MAINTAINER Joffrey F <joffrey@docker.com>
+
+RUN mkdir /home/docker-py
+WORKDIR /home/docker-py
+
+RUN apt-get update
+RUN apt-get install -y python python-setuptools && easy_install pip
```
dnephin: Couple minor suggestions here:

* [best practices](https://docs.docker.com/articles/dockerfile_best-practices/#run) says to make these a single line, otherwise you can run into a situation where the update is cached without the available package you need to install
* instead of `easy_install pip` why not add `python-pip`  to the apt-get install line? If you want a newer version https://pip.pypa.io/en/latest/installing.html#install-pip suggests their install script.

Makefile
```diff
@@ -23,3 +23,7 @@ integration-test: build
 
 integration-test-py3: build-py3
 	docker run -v `$(HOST_TMPDIR)`:/tmp -v /var/run/docker.sock:/var/run/docker.sock docker-py3 py.test -rxs tests/integration_test.py
+
+integration-ci:
+	docker build -t dpy-tests -f ./tests/Dockerfile .
+	docker run --privileged -t dpy-tests
```
dnephin: Instead of  running both processes in the same container, what do you think about doing this:

* change the base of this `Dockerfile` to be something like `debian:jessie`
* change the docker run link to the `dind` image instead

```
docker run -d --name dpy-dind --privileged dockerswarm/dind:1.8.1 docker -d  -H tcp://0.0.0.0:2375
docker run --rm --link=dpy-dind:docker --env="DOCKER_HOST=tcp://docker:2375"
docker rm -vf dpy-dind
```

(which would allow the `CMD` in the `Dockerfile` to change to just `py.test -rxs tests/integration_test.py`

`dpy-dind` could be a variable even, which would allow for jenkins to set a unique value for each build.

This is the approach aanand used in https://github.com/docker/compose/pull/1783

tests/Dockerfile
shin-: > best practices says to make these a single line, otherwise you can run into a situation where the update is cached without the available package you need to install

WIll update!

> instead of easy_install pip why not add python-pip to the apt-get install line? If you want a newer version https://pip.pypa.io/en/latest/installing.html#install-pip suggests their install script.

"Surprisingly" (har har), python-pip [is outdated/broken](http://stackoverflow.com/questions/27341064/how-do-i-fix-importerror-cannot-import-name-incompleteread), so I'm using `easy_install` to get the latest version. I guess I could fetch their install script and use that instead, but that seems really unnecessary for what we're doing.

Makefile
shin-: What's the upside to that method? Seems unnecessarily complicated for docker-py.

dnephin: I think one advantage is that you can use the `tests/Dockerfile` without `dind` if you want to, which can make it easier to debug problems.

It also makes it easier to see errors from the daemon (since the current setup sends them to `/dev/null`).

aanand: Yeah, it also means you can run the tests against any arbitrary daemon by passing in the DOCKER_HOST environment variable (so down the line we can e.g. run it against a Swarm cluster).

It's also nice to have the Docker-in-Docker stuff isolated to a container that's not getting messed with.

shin-:
@aanand @dnephin PTAL!

aanand:
I still don't like that those tests require access to the daemon's filesystem - we'll have to change them if we want to test against Swarm (which we should do). Still, that's arguably out of scope for this.

I think it could be squashed to fewer commits, then LGTM.

shin-:
How else would you test volume binds though?

PR #760, opened 2015-09-09
CI fixes (wip)

shin-:


Makefile
```diff
@@ -28,7 +28,6 @@ integration-test-py3: build-py3
 	docker run -v `$(HOST_TMPDIR)`:/tmp -v /var/run/docker.sock:/var/run/docker.sock docker-py3 py.test -rxs tests/integration_test.py
 
 integration-dind: build build-py3
-	docker build -t dpy-tests -f ./tests/Dockerfile .
```
dnephin: Is this being done as part of the jenkins config? Maybe just change this to a comment so that users can run it locally ?

shin-: No, it's just unneeded, and I forgot to remove it - we already have Dockerfiles for building docker-py at the root.

dnephin: Ah I see, so should the `integration-dind` target depend on `build` and `build-py3` ?

shin-: yep! Already does!

dnephin: oops, right!

dnephin:
LGTM

PR #763, opened 2015-09-10
API splitup

shin-:
Split client methods into different mixins for clarity / cleanliness.


aanand:
I'm not the biggest fan of mixins, but it's better than one huge class, and I never liked that ClientBase superclass.

Needs squashing into logical commits, then LGTM.

mpetazzoni:
LGTM. Like @aanand not a huge fan of mixins either, but this was getting out of hand so this will definitely be more manageable!

PR #764, opened 2015-09-11
Volume API support

### LINKED ISSUE #744: add support for volume api of docker engine 1.9
https://github.com/docker/docker/pull/14242

shin-:
Fixes #744


tests/test.py
```diff
@@ -2021,6 +2103,100 @@ def test_build_container_invalid_container_limits(self):
             })
         )
 
+    ###################
+    #  VOLUMES TESTS  #
+    ###################
+
+    @pytest.mark.skipif(docker.utils.compare_version(
+        '1.21', docker.constants.DEFAULT_DOCKER_API_VERSION
+    ) < 0, reason="API version too low")
```
dnephin: This very check is rather verbose (and needs to be repeated in a few places).  You can actually do something like this to re-use the mark (https://pytest.org/latest/skipping.html#skipif):

```python

def requires_api_version(version):
    return pytest.mark.skipif(
        docker.utils.compare_version(
            version,
            docker.constants.DEFAULT_DOCKER_API_VERSION
        ) < 0,
        reason="API version too low"
    )


@requires_api_version('1.21')
def test...
```

docs/api.md
```diff
@@ -870,6 +919,23 @@ Nearly identical to the `docker version` command.
 }
 ```
 
+## volumes
+
+List volumes currentlu registered by the docker daemon. Similar to the `docker volume ls` command.
```
dnephin: "currently"

docker/api/volume.py
```diff
@@ -0,0 +1,50 @@
+import functools
+
+from .. import errors
+from ..utils import utils
+
+
+def check_api_version(f):
```
dnephin: I'd suggest parameterizing this decorator with the version, so it can be used anywhere.

shin-:
Included a fix for #769 as well.

docker/api/container.py
```diff
@@ -312,16 +309,13 @@ def start(self, container, binds=None, port_bindings=None, lxc_conf=None,
         res = self._post_json(url, data=start_config)
         self._raise_for_status(res)
 
-    @check_resource
+    @utils.minimum_version('1.17')
```
dnephin: This new decorator is great!

docker/api/exec_api.py
```diff
@@ -38,16 +37,14 @@ def exec_create(self, container, cmd, stdout=True, stderr=True, tty=False,
...
-            )
         if isinstance(exec_id, dict):
             exec_id = exec_id.get('Id')
         res = self._get(self._url("/exec/{0}/json", exec_id))
         return self._result(res, True)
 
+    @utils.minimum_version('1.15')
     def exec_resize(self, exec_id, height=None, width=None):
         if utils.compare_version('1.15', self._version) < 0:
             raise errors.InvalidVersion('Exec is not supported in API < 1.15')
```
dnephin: Should we delete these 2 lines as well?

docker/api/exec_api.py
```diff
@@ -59,6 +56,7 @@ def exec_resize(self, exec_id, height=None, width=None):
         res = self._post(url, params=params)
         self._raise_for_status(res)
 
+    @utils.minimum_version('1.15')
     def exec_start(self, exec_id, detach=False, tty=False, stream=False):
         if utils.compare_version('1.15', self._version) < 0:
             raise errors.InvalidVersion('Exec is not supported in API < 1.15')
```
dnephin: delete these two as well?

docs/api.md
```diff
@@ -695,6 +733,16 @@ Remove an image. Similar to the `docker rmi` command.
...
 
+## remove_volume
+
+Remove a volume. Similar to the `docker volume rm` command.
+
+**Params**:
+
+* name (str): The volume's name
+
+**Returns** (bool): True on successful removal. Failure will raise an exception.
```
dnephin: It would be good to document which exception.

Using `rst` and sphinx you can do ``` `:class:docker.errors.APIError` ``` and get it all linked up :smiling_imp:

tests/integration_test.py
```diff
@@ -101,6 +103,13 @@ def tearDown(self):
                 pass
         for folder in self.tmp_folders:
             shutil.rmtree(folder)
+
+        for volume in self.tmp_volumes:
+            try:
+                self.client.remove_volume(volume)
+            except docker.errors.APIError:
+                pass
```
dnephin: This is one place where pytest can really improve the test code. Instead of having every test attempt this, you can use a https://pytest.org/latest/yieldfixture.html#yieldfixture (or https://pytest.org/latest/fixture.html#finalization) in just the tests that use a volume to do the proper cleanup.

We don't have to make that change now, but I think we should move in that direction eventually.

tests/integration_test.py
```diff
@@ -1370,9 +1384,63 @@ def runTest(self):
...
-# BUILDER TESTS #
+# VOLUMES TESTS #
 #################
 
+@requires_api_version('1.21')
+class TestVolumes(BaseTestCase):
+    def test_create_volume(self):
+        name = 'perfectcherryblossom'
+        result = self.client.create_volume(name)
+        self.tmp_volumes.append(name)
```
dnephin: Minor, but if the tests crash or are killed between line 1394 and 1395 the volume won't be cleaned up. Same problem if the client call creates the volume, but then hits a timeout.

I think this needs to be:

```python
name = 'perfectcherryblossom'
self.tmp_volumes.append(name)
result = self.client.create_volume(name)
```

Same for the tests below

tests/integration_test.py
shin-: Isn't an iteration on a list of size 0 almost a no-op? It feels to me like we would make the code more complex without actually gaining anything from it.

dnephin: I don't think the improvement is in performance, it's in clarify.

With a single base class for all tests, it's not clear which tests actually make use of this feature.  Refactoring and moving the test suite around becomes harder because everything needs a single base.

With a yield fixture it's explicit that the test requires volume_cleanup. 

Not sure how it would make the code more complex.

dnephin:
LGTM (once janky is done and happy)

I think the docs for filter format could be improved (currently undefined format), but we can do that as a followup

PR #765, opened 2015-09-12
Support API version 1.20

shin-:


aanand:
Those snippets look near-identical. Maybe there should be a test utility method?

```python
def get_volume_info(container, destination, version):
    """
    Takes a container dictionary and volume destination path
    and returns a tuple of (rw, origin), using the appropriate
    fields for the specified API version
    """
    if docker.utils.compare_version('1.20', version) < 0:
        # old logic
    else:
        # new logic
```

tests/integration_test.py
```diff
@@ -285,13 +279,29 @@ def test_ro(self):
             logs = logs.decode('utf-8')
         self.assertIn(self.filename, logs)
 
-        # FIXME: format changes in API version >= 1.20
+        self.assertIn(self.filename, logs)
```
aanand: This seems to be a duplicate of the line above - merge conflict resolution mishap?

aanand:
Duplicate line aside, LGTM

PR #767, opened 2015-09-14
Support group_add param in host config

shin-:
Related to #765

### DIFF
--- a/docker/utils/utils.py
+++ b/docker/utils/utils.py
@@ -449,7 +449,7 @@ def create_host_config(
     restart_policy=None, cap_add=None, cap_drop=None, devices=None,
     extra_hosts=None, read_only=None, pid_mode=None, ipc_mode=None,
     security_opt=None, ulimits=None, log_config=None, mem_limit=None,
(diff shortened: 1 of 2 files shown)


dnephin:
LGTM

docker/utils/utils.py
```diff
@@ -509,6 +509,13 @@ def create_host_config(
     if devices:
         host_config['Devices'] = parse_devices(devices)
 
+    if group_add:
+        if compare_version(version, '1.20') < 0:
```
aanand: I think this should have been `>`, not `<` - if I set the DEFAULT_DOCKER_API_VERSION to '1.21', then this error gets raised.

shin-: yup, I think I fixed it afterwards but it may be in one of the pending PRs. Let me find it.

shin-: https://github.com/docker/docker-py/commit/6cf7758f25de72a3f53528de0a376eadba902ae9#diff-193b0cbc51c3ae87519912dd6aeadec4R522

PR #768, opened 2015-09-15
Remove not exists file in MANIFEST.in

ColinHuang:
Should be remove in MANIFEST.in
Or it will shows warning message when doing ``pip install docker-py``

    warning: no files found matching 'requirements3.txt'


shin-:
Thanks!

PR #770, opened 2015-09-15
Support for cpu_quota and cpu_period

### LINKED ISSUE #766: Add support for --cpu-quota & --cpu-period run flags
Any chance we could add support for the above run flags?
https://docs.docker.com/reference/commandline/run/

shin-:
Fixes #766

### DIFF
--- a/docker/utils/utils.py
+++ b/docker/utils/utils.py
@@ -457,7 +457,8 @@ def create_host_config(
     restart_policy=None, cap_add=None, cap_drop=None, devices=None,
     extra_hosts=None, read_only=None, pid_mode=None, ipc_mode=None,
     security_opt=None, ulimits=None, log_config=None, mem_limit=None,
-    memswap_limit=None, cgroup_parent=None, group_add=None, version=None
+    memswap_limit=None, cgroup_parent=None, group_add=None, cpu_quota=None,
+    cpu_period=None, version=None
 ):
     host_config = {}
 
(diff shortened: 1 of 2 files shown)


shin-:
#764 needs to be merged for the tests to pass.

docker/utils/utils.py
```diff
@@ -601,6 +602,30 @@ def create_host_config(
...
+        raise TypeError(
+            'Invalid type for cpu_quota param: expected int but'
+            ' found {0}'.format(type(cpu_quota))
+        )
+    elif cpu_quota:
+        if compare_version(version, '1.19') > 0:
+            raise errors.InvalidVersion(
+                'cpu_quota param not supported for API version < 1.19'
+            )
+        host_config['CpuQuota'] = cpu_quota
```
aanand: Minor thing, but this style of conditional makes me do a double-take - plus we're immediately doing another check in the second branch. I think a single outer check and two inner checks would be more normal.

Also, could we use `version_lt`? I find it much easier to read.

```python
if cpu_quota:
    if not isinstance(cpu_quota, int):
        raise TypeError(...)
    if version_lt(version, '1.19'):
        raise errors.InvalidVersion(...)
    host_config['CpuQuota'] = cpu_quota
```

Same for the `cpu_period` conditional.

shin-:
Should be good now.

aanand:
LGTM

PR #778, opened 2015-09-17
Fixes python tests for Issue 15566. PR for actual issue is 15567

shishir-a412ed:
Signed-off-by: Shishir Mahajan <shishir.mahajan@redhat.com>


shin-:
As mentioned in the docker/docker#15567, this seems to be the wrong approach (we should maintain backwards compatibility at every level of the stack). 

As a result, I am closing this for now until the issue is properly resolved upstream.

PR #780, opened 2015-09-19
[WIP] Contributing guidelines

shin-:
@docker/docker-py-maintainers 
Thoughts, ideas, corrections, comments... Most of what I wrote so far is based on experience, but maybe we can take this opportunity to improve the things that need improving.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -15,6 +15,7 @@ the following information:
 ```
 pip freeze | grep docker-py && python --version && docker version
 ```
+- OS, distribution and OS version
 - The issue you're encountering including a stacktrace if applicable
 - If possible, steps or a code snippet to reproduce the issue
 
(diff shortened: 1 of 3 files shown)


dnephin:
overall LGTM

CONTRIBUTING.md
```diff
@@ -0,0 +1,74 @@
+# Contributing guidelines
+
+Thank you for your interest in the project. We look forward to your
```
mnowster: It'd be good to refer to docker's contributing guidelines here, in the same way we do for docker-compose. https://github.com/docker/compose/blob/master/CONTRIBUTING.md#contributing-to-compose

CONTRIBUTING.md
```diff
@@ -0,0 +1,74 @@
...
+[search function](https://github.com/docker/docker-py/search) to find out if
+it has already been addressed, or is currently being looked at.
+
+## Submitting pull requests
+
+Do you have a fix for an existing issue, or want to add a new functionality
+to docker-py? We happily welcome pull requests. Here are a few tips to make
+the review process easier on both the maintainers and yourself.
+
+### 1. Sign your commits
```
mnowster: Having commits be a series of logical well written commits is beneficial to talk about here. In compose we link to Government Digital Service(where I used to work) styleguide for commits, https://github.com/alphagov/styleguides/blob/master/git.md

CONTRIBUTING.md
```diff
@@ -0,0 +1,74 @@
...
+
+Please refer to the ["Sign your work"](https://github.com/docker/docker/blob/master/CONTRIBUTING.md#sign-your-work)
+paragraph in the Docker contribution guidelines.
+
+### 2. Make sure tests pass
```
mnowster: Can we have a step that details how to set up a development environment to enable you to work on docker-py? It'll help ensure that our projects are accessible to people of all levels of experience.

CONTRIBUTING.md
```diff
@@ -0,0 +1,74 @@
...
+Whether you're adding new functionality to the project or fixing a bug, please
+add relevant tests to ensure the code you added continues to work as the
+project evolves.
+
+### 7. Add docs
+
+This usually applies to new features rather than bug fixes, but new behavior
+should always be documented.
+
+### 8. Ask questions
```
mnowster: I like this positive re-enforcement of questions being welcomed :smile_cat:

CONTRIBUTING.md
mnowster: Re: Tests, it'd also be helpful to have an example command of how to run an individual test.

mnowster:
I'm really happy to see docker-py get a contributing guideline :+1:

CONTRIBUTING.md
shin-: Mh, we can mention it. It's never been an issue in docker-py's case from memory.

CONTRIBUTING.md
shin-: I guess that's fair. I considered it, but since we're not really in the spotlight when contributor events happen I thought it wasn't necessary. 

But it would surely make the project more welcoming to a broader range of people.

CONTRIBUTING.md
shin-: I actually decided against it when writing this document because I feel like the major part is irrelevant to what we're doing here. I'd rather refer to specific subsections like we do in the "sign your commits" part. Do you feel like there's one part of the docker/docker document that should be explicitly mentioned but isn't?

CONTRIBUTING.md
```diff
@@ -0,0 +1,74 @@
...
+here is a set of guidelines we recommend you follow.
+
+## Reporting issues
+
+We do our best to ensure bugs don't creep up in our releases, but some may
+still slip through. If you encounter one while using docker-py, please create
+an issue [in the tracker](https://github.com/docker/docker-py/issues/new) with
+the following information:
+
+- docker-py version, docker version and python version
```
TomasTomecek: I think it could make sense to mention OS/distribution and whether docker-py comes from PyPI.

shin-: Good point!

shin-:
I'm going to merge this, we can always continue to improve it as we go if need be. Thanks everyone for your feedback!

PR #783, opened 2015-09-22
Integration test fixes

aanand:
- Only pull the busybox image if we don't already have it (cuts down on setup time)
- Remove volume binds tests' need to share the `/tmp` directory with the Docker host (longer explanation in the commit message)

### DIFF
--- a/tests/integration_test.py
+++ b/tests/integration_test.py
@@ -33,7 +33,7 @@ from six.moves import BaseHTTPServer
 from six.moves import socketserver
 
 import docker
-from docker.errors import APIError
+from docker.errors import APIError, NotFound
 from docker.utils import kwargs_from_env
 
 from .base import requires_api_version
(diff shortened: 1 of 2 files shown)


tests/integration_test.py
```diff
@@ -114,6 +118,21 @@ def tearDown(self):
 
         self.client.close()
 
+    def run_container(self, *args, **kwargs):
+        container = self.client.create_container(*args, **kwargs)
+        self.client.start(container)
+        self.tmp_containers.append(container)
```
dnephin: I think these two lines need (123, 124) to be swapped, otherwise container start failures will prevent a container from being removed.

aanand: True. Updated.

aanand:
Build's green.

dnephin:
LGTM

shin-:
Merged! Thanks.

PR #786, opened 2015-09-22
Add docker.api to setup packages

ibuildthecloud:
`docker.api` package was missing from setup.py

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -24,8 +24,8 @@ setup(
     description="Python client for Docker.",
     url='https://github.com/docker/docker-py/',
     packages=[
-        'docker', 'docker.auth', 'docker.unixconn', 'docker.utils',
-        'docker.utils.ports', 'docker.ssladapter'
+        'docker', 'docker.api', 'docker.auth', 'docker.unixconn',
+        'docker.utils', 'docker.utils.ports', 'docker.ssladapter'
     ],
     install_requires=requirements,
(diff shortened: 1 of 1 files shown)


shin-:
What would be an example of a use-case where accessing this package directly is necessary?

ibuildthecloud:
Great question... our stuff fails on master because it can't find api.  Are you saying this package does not need to be installed?

ibuildthecloud:
@shin- Seems to be referenced from here: https://github.com/docker/docker-py/blob/4475310eeac905e0def4e0eeec6f3573a65bb028/docker/client.py#L25

shin-:
Did you install the package from pypi? Which version of Python are you using? I'm not seeing any issues with a fresh install on 2.7 or 3.4 :/

ibuildthecloud:
The issue is with master only, so git clone and `setup.py install`.  1.4.0 is fine because `api` didn't exist.

ibuildthecloud:
```python
.tox/py27/local/lib/python2.7/site-packages/docker/__init__.py:20: in <module>
    from .client import Client, AutoVersionClient # flake8: noqa
.tox/py27/local/lib/python2.7/site-packages/docker/client.py:25: in <module>
    from . import api
E   ImportError: cannot import name api
```

shin-:
Ah, you're right. I was confused, sorry.

PR #787, opened 2015-09-23
Don't break when volume binds contain unicode characters

### LINKED ISSUE #782: Create container bind volume with Unicode folder name
If bind volume folder name is Unicode, for example Japanese, It will raise exception:
Host volume and container volume both should handle Unicode.

```
  File "/home/vagrant/.local/share/virtualenvs/qnap/local/lib/python2.7/site-packages/docker/utils/utils.py", line 461, in create_host_config
    host_config['Binds'] = convert_volume_binds(binds)
  File "/home/vagrant/.local/share/virtualenvs/qnap/local/lib/python2.7/site-packages/docker/utils/utils.py", line 208, in convert_volume_binds
    k, v['bind'], mode
```

shin-:
Also includes a few unit tests for utils.convert_volume_binds

Fixes #782

### DIFF
--- a/tests/utils_test.py
+++ b/tests/utils_test.py
@@ -1,7 +1,12 @@
+# -*- coding: utf-8 -*-
+
 import os
 import os.path
 import shutil
 import tempfile
+
+import pytest
+import six
 
 from docker.client import Client
 from docker.constants import DEFAULT_DOCKER_API_VERSION
@@ -9,7 +14,7 @@ from docker.errors import DockerException
 from docker.utils import (
     parse_repository_tag, parse_host, convert_filters, kwargs_from_env,
     create_host_config, Ulimit, LogConfig, parse_bytes, parse_env_file,
-    exclude_paths,
+    exclude_paths, convert_volume_binds,
 )
 from docker.utils.ports import build_port_bindings, split_port
 from docker.auth import resolve_repository_name, resolve_authconfig
@@ -17,7 +22,6 @@ from docker.auth import resolve_repository_name, resolve_authconfig
 from . import base
 from .helpers import make_tree
 
-import pytest
 
 TEST_CERT_DIR = os.path.join(
     os.path.dirname(__file__),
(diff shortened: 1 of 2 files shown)


tests/utils_test.py
```diff
@@ -45,6 +49,53 @@ def setUp(self):
...
+                'bind': '/data'
+            }
+        }
+        self.assertEqual(convert_volume_binds(data), ['/mnt/vol1:/data:rw'])
+
+    def test_convert_volume_binds_unicode(self):
+        expected = ['/mnt/지연:/unicode/박:rw']
+
+        if six.PY2:
+            expected[0] = expected[0].decode('utf-8')
```
aanand: Might be able to avoid this step if we go:

```python
expected = [six.u('/mnt/지연:/unicode/박:rw')]
```

tests/utils_test.py
```diff
@@ -45,6 +49,53 @@ def setUp(self):
...
+
+        data = {
+            '/mnt/지연': {
+                'bind': '/unicode/박',
+                'mode': 'rw'
+            }
+        }
```
aanand: These will be byte strings in Python 2 - is that correct? Do we care what happens when users pass in `unicode` strings in Python 2?

shin-:
@aanand Thanks, didn't know about `six.u`. PTAL?

aanand:
Cool. Perhaps the first test should be rewritten so it tests byte string input on all Python versions:

```python
    def test_convert_volume_bindings_binary_input(self):
        expected = [six.u('/mnt/지연:/unicode/박:rw')]

        data = {
            b'/mnt/지연': {
                'bind': b'/unicode/박',
                'mode': 'rw'
            }
        }
        self.assertEqual(
            convert_volume_binds(data), expected
        )
```

shin-:
Okay, I've tried all sorts of combination, I've settled on just separating py2 and py3 tests completely (so it's 2+2 now). On the bright side, results were consistent all along (and we have tests to prove it)!

aanand:
OK, I see the rationale, but instead of entirely separate test methods, could we just switch on the Python version *inside* the method?

```python
def test_convert_volume_binds_unicode_bytes_input(self):
    if six.PY2:
        expected = [unicode('/mnt/지연:/unicode/박:rw', 'utf-8')]
        data = {
            '/mnt/지연': {
                'bind': '/unicode/박',
                'mode': 'rw'
            }
        }
    else:
        expected = ['/mnt/지연:/unicode/박:rw']
        data = {
            bytes('/mnt/지연', 'utf-8'): {
                'bind': bytes('/unicode/박', 'utf-8'),
                'mode': 'rw'
            }
        }

    self.assertEqual(convert_volume_binds(data), expected)

def test_convert_volume_binds_unicode_unicode_input(self):
    if six.PY2:
        expected = [unicode('/mnt/지연:/unicode/박:rw', 'utf-8')]
        data = {
            unicode('/mnt/지연', 'utf-8'): {
                'bind': unicode('/unicode/박', 'utf-8'),
                'mode': 'rw'
            }
        }
    else:
        expected = ['/mnt/지연:/unicode/박:rw']
        data = {
            '/mnt/지연': {
                'bind': '/unicode/박',
                'mode': 'rw'
            }
        }

    self.assertEqual(convert_volume_binds(data), expected)
```

Issue #417, opened 2014-12-10
Add helper API on top of core API


hamiltont:
It seems that the current architecture of `docker-py` is to exactly mirror the API of docker, and not add any "helper functions" on top. I totally agree that the docker-py core should mirror the API of docker as closely as possible. However, I keep finding myself re-writing common helpers over and over when integrating docker-py with other code. 

Would the maintainers be opposed to adding a `helper` API on top of the existing core API that adds  common operations, such as checking if a container or image exists, converting raw strings into more pythonic types, etc? I'm happy to provide some initial PRs to build a base for this new API if someone gives me a green light

gtaylor:
We had to figure out how to handle this very issue with boto a few years ago. Initially, I felt that a separate package with the more sane API would make the most sense. However, that would mean juggling two projects with two potentially different release cadences, with a really confusing backwards compatibility story (boto evolves/evolved fast, like Docker).

We eventually settled on a layered approach. The official Amazon Java SDK was imitated closely in ``layer1``, and a more humane API took form in ``layer2``. You can see this in practice in the [dynamodb2 module](https://github.com/boto/boto/tree/develop/boto/dynamodb). Amazon's DynamoDB API was particularly difficult to deal with at first, so this second layer brought some much-needed sanity and usability to the table.

However, some developers really do need access to the "native" level in order to specially optimize or handle some more complex cases. Most end up using layer2, because it's easier, but this setup allows everyone to be happy.

I don't know what this would look like in docker-py, but figured I'd share experiences from another Python module that also had to track a rapidly changing upstream API. It sounds like you are leaning that way, anyway.

FWIW, I'd gravitate towards a module name that suggests its primary purpose (helpers, conveniences, friendliness). Maybe something like ``docker.humane``.

Julian:
FWIW I am +1 on this being a separate package, and I've already started assembling some helpers that I've used -- to add to your list, `temporary_image` and `temporary_container`, which create images and containers and delete them in a contextmanager. Possibly by the time I'm done I'll have shoved them into a separate library myself anyhow.

hamiltont:
@Julian great ideas! FWIW I've got some basic code but it's a bit stalled, if anyone wants to take point on a PR for this I'm happy to deposit my current code into a gist in case it's useful

Julian:
Nice! I've got some as well, although it's a bit stuck on not having an
easy way to write tests for it unfortunately, but maybe starting to put
some stuff in the same place would get this kick started, yeah.

[...]

PR #788, opened 2015-09-24
[WIP] Efficiency

shin-:
# For context, see
- #417 "Add helper API on top of core API"
- [tools-project label](https://github.com/docker/docker-py/labels/tools-project)
# What is it?
1. Compilation of helper functions that help perform common tasks outside of the API proper. Example:

``` python
In [1]: from docker import efficiency
In [2]: efficiency.start_auto_remove?
Signature: efficiency.start_auto_remove(container, *args, **kwargs)
Docstring:
Start a container and try to remove it when it's finished running,
similar to using --autorm in the docker CLI.

**Params:**
    container: ID of the container to be started
    args, kwargs: `Client.start()` arguments
File:      ~/work/pydocker/docker/efficiency/efficiency.py
Type:      function
```
1. "Smart objects" providing an object-oriented interface on top of the client layer

``` python
In [29]: import docker

In [30]: c = docker.Client()

In [31]: docker.efficiency.init(c)

In [32]: docker.efficiency.get_image('busybox:latest')
Out[32]: <DockerImage(id=d7057cb020844f24)>

In [33]: img = _

In [34]: img.os
Out[34]: u'linux'

In [36]: img.architecture
Out[36]: u'amd64'

In [37]: img.create_container(command='true')
Out[37]: <DockerContainer(id=58bf1374a4116b38)>

In [38]: ctnr = _

In [39]: ctnr.status
Out[39]: 'stopped'

In [40]: ctnr.start()

In [41]: ctnr.wait()
Out[41]: 0 
```
# TODO
- [ ] Test coverage
- [ ] Documentation (lots of docstring already available)
- [ ] Cleanup (smart objects in their own files, etc.)


docker/efficiency/builds.py
```diff
@@ -0,0 +1,221 @@
+import json
```
dnephin: `efficiency` seems a bit strange to me for a module name.  What do you think about just making this `docker.build` ?

docker/efficiency/builds.py
```diff
@@ -0,0 +1,221 @@
...
+                parsed_lines.append(parsed_line)
+            match = re.match(build_success_re, line)
+            if match:
+                image_id = match.group(1)
+        except ValueError:
+            # sometimes all the data is sent on a single line
+            # This ONLY works because every line is formatted as
+            # {"stream": STRING}
+            lines = re.findall('{\s*"stream"\s*:\s*"[^"]*"\s*}', line)
+            return get_build_id(lines, discard_logs)
```
dnephin: This sounds like a bug, do you know we can reproduce it?

If there is an open issue on engine, maybe add the url to the issue here

docker/efficiency/builds.py
```diff
@@ -0,0 +1,221 @@
...
+    )
+
+
+def make_context_from_directory(path, dockerfile='Dockerfile'):
+    return BuildContext(
+        'directory',
+        path,
+        dockerfile=dockerfile,
+        job_params={}
+    )
```
dnephin: What do you think about making these factory classmethods on `BuildContext` instead of separate functions?

```python

class BuildContext(BuildCtxTuple):

    @classmethod
    def from_tarball(cls, path, dockerfile='Dockerfile'):
       return cls(...)

    @classmethod
    def from_dockerfile(cls, ...):
        ...
```

docker/efficiency/commons.py
```diff
@@ -0,0 +1,36 @@
...
+
+    @property
+    def id(self):
+        return self._id
+
+
+class Interactive(Identifiable):
+    def __init__(self, client, id):
+        super(Interactive, self).__init__(id)
+        self._client = client
```
dnephin: Base classes like this are a common pattern, but I always avoid them at all costs.

The code re-use is pretty minimal here. Instead of the `@property` `_id` , you could just assign to `self.id`, right ?

So instead of having two base classes you have to do two assignments in a couple constructors, but it means that developers don't have to worry about things that might be hidden away in a base class.

docker/efficiency/containers.py
```diff
@@ -0,0 +1,178 @@
...
+        self._client.exec_resize(self.id, height, width)
+
+    def __str__(self):
+        return '<DockerExec(id={0})>'.format(self.id[:16])
+
+    def __repr__(self):
+        return str(self)
+
+
+class Container(commons.Interactive):
```
dnephin: I think we should to make this class compatible with `compose.container.Container` so that we can replace it with this one.

docker/efficiency/efficiency.py
```diff
@@ -0,0 +1,113 @@
+from . import builds
+from . import containers
+from . import images
+
+_client = None
```
dnephin: I think this is a bad idea. Globals make libraries much more difficult to use.

I would pass in a client instead of registering a global. It's going to make this API much more re-usable, easier to test, and easier to reason about.

bfirsh:
Carried and merged in https://github.com/docker/docker-py/issues/1186. :tada:

PR #789, opened 2015-09-25
Allow to specify ps_args when listing processes

PierreF:
Docker remote API allow to use custom option to ps for endpoint "top" (ps_args). Unfortunately docker-py does not allow to specify ps_args.

This pull request fix this issue by adding an optional ps_args argument

### DIFF
--- a/docker/api/container.py
+++ b/docker/api/container.py
@@ -325,9 +325,12 @@ class ContainerApiMixin(object):
         self._raise_for_status(res)
 
     @utils.check_resource
-    def top(self, container):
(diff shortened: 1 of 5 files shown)


shin-:
Thanks!

LGTM - as a bonus, adding a small test for that feature would be nice.

PierreF:
Added test.
Test test using the fake API does not check returned value. I don't know how to do that (response that change depending on params).

Anyway, integration test does check the output.

shin-:
Cool! LGTM :+1:

dnephin:
LGTM

PR #790, opened 2015-09-25
Shorter stack traces in tests

aanand:
We don't need the full-blown "here's the code of the massive library method that your error came from" report that py.test gives you by default.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -20,15 +20,15 @@ unit-test-py3: build-py3
 	docker run docker-py3 py.test tests/test.py tests/utils_test.py
 
 integration-test: build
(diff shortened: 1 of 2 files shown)


dnephin:
If you want to just make this default, you could do this https://pytest.org/latest/customize.html#adding-default-options

aanand:
@dnephin nice, I'll update it

shin-:
Actually, if something fails, the more verbose the better I think... It's not always useful but I'll take too much information over not enough of it. :x

dnephin:
I personally always use `--tb=short`, you still get the full traceback (filename and line number) for the relevant stack frames, it just doesn't show you the full source for every function.

https://pytest.org/latest/usage.html#modifying-python-traceback-printing

LGTM

mnowster:
LGTM

PR #791, opened 2015-09-25
_url can take arbitrarily many arguments

aanand:
This is needed by the networking PR.

I removed a test case for when `None` is explicitly passed in. I don't think we'd ever do that, but am I mistaken?


shin-:
Fixed the flake8 issues and merged :)

PR #793, opened 2015-09-28
Support explicit protocol in Client.port

### LINKED ISSUE #792: Client.port() only reports the UDP exposed port mapping if the same port is configured to be mapped on UDP and TCP.
E.g.:
`docker ps`
```
CONTAINER ID  IMAGE         COMMAND        CREATED        STATUS        PORTS                                         NAMES
6348dd5bc7c2  ubuntu:14.04  /usr/bin/bash  8 seconds ago  Up 8 seconds  0.0.0.0:49153->80/udp, 0.0.0.0:49159->80/tcp  expose_ports
```

`In ipython`
```
In [1]: docker_client.port('6348dd5bc7c2', 80)
Out[1]: [{u'HostIp': u'0.0.0.0', u'HostPort': u'49153'}]
```

The reason seems to lie in `/docker/api/container.py`, line 208-210
```python
def port(self, container, private_port):

...

h_ports = port_settings.get(s_port + '/udp')
if h_ports is None:
    h_ports = port_settings.get(s_port + '/tcp')
```

Instead of basically only accepting a port number for `private_port` and trying for udp and tcp exposed ports, I'd suggest that the method rather accepts an explicit `private_port` input like `1234/tcp`.

shin-:
User can now specify a protocol in Client.port using the port/proto syntax. Default port is now TCP instead of UDP (to match the behavior of `docker port`) when no proto is specified.

FIxes #792

### DIFF
--- a/docker/api/container.py
+++ b/docker/api/container.py
@@ -196,7 +196,7 @@ class ContainerApiMixin(object):
         res = self._get(self._url("/containers/{0}/json", container))
         self._raise_for_status(res)
         json_ = res.json()
-        s_port = str(private_port)
+        private_port = str(private_port)
         h_ports = None
 
         # Port settings is None when the container is running with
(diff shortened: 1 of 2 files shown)


dnephin:
LGTM

PR #794, opened 2015-09-30
Allow a host to be passed for the IP portion of a port binding.

tbeadle:
Previously, if a hostname were provided in place of an IP address in a port
binding definition, the docker daemon would simply ignore it and bind to all
interfaces instead of a specific one.  This change makes it so that a hostname
can be used in place of the IP.  The address for that host will be resolved
to an IP and then the IP is passed to the docker daemon.  This still requires
that the IP that it resolves to exists on the docker host.

This is particularly useful in situations where it is common to move containers
from one docker host to another.  One could have a docker-compose.yml, for
instance, that does not need to change (using the hostname in the 'ports'
definition).  When the container is to be moved to a new host, DNS can just be
updated to reflect the address on the new host to be used.  Then docker-compose
will use the right address when it's brought up on that host.

Signed-off-by: Tommy Beadle tbeadle@gmail.com

### DIFF
--- a/docker/utils/utils.py
+++ b/docker/utils/utils.py
@@ -18,6 +18,7 @@ import os
 import os.path
 import json
 import shlex
+import socket
 import tarfile
 import tempfile
 import warnings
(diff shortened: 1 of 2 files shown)


ijc:
PR grooming: Thanks for your PR and sorry it took so long before anyone looked at it.

I think this would make the behaviour of `docker-py` different to the docker cli:
```
$ docker run -p localhost:8080:8080 nginx
docker: invalid publish opts format (should be name=value but got 'localhost:8080:8080').
See 'docker run --help'.
```
which I think is undesirable. As you say the IP needs to be available on the host, so if anything I think this should be an engine feature to resolve things at that end (NB: I haven't fully thought that through, that may have its own issue).

So I'm going to close now, thanks (and sorry) again.

PR #796, opened 2015-10-01
Support the 'since' option in the 'containers/<id>/logs' endpoint (API upgrade)

v-boyko:
Added 'since' timestamp parameter to the 'logs' command

(as in https://docs.docker.com/reference/api/docker_remote_api_v1.19/#get-container-logs)


docker/api/container.py
```diff
@@ -174,6 +175,14 @@ def logs(self, container, stdout=True, stderr=True, stream=False,
                 if tail != 'all' and (not isinstance(tail, int) or tail <= 0):
                     tail = 'all'
                 params['tail'] = tail
+            if utils.compare_version('1.19', self._version) >= 0:
+                if isinstance(since, datetime):
```
dnephin: I think normally we want to raise an error if someone is using an option that isn't supported by the API version.  It would be nice if the server did this, but I don't think that's the case most of the time.

v-boyko:
@dnephin Thanks for the tip, I've added the error in case the API doesn't support the flag.

shin-:
LGTM :+1:

dnephin:
LGTM

PR #798, opened 2015-10-01
Remove skip test

mnowster:
It has been reported that the bug, https://github.com/docker/docker/issues/15633, in docker has now been fixed so we can re-instate this test.


dnephin:
That fix isn't in a release yet though, so we'll have to wait for 1.9 before we can merge this.

mnowster:
@dnephin Good point. I made mistake of thinking it had been included in the 1.8.2 release.

shin-:
Updated version #850 - closing.

PR #801, opened 2015-10-08
1.5.0 release

shin-:


shin-:
I wonder if we should release that with the default API version set to `1.21`? Given there's a good chance this goes out before Docker 1.9.0 releases, this might be very confusing to users.

dnephin:
Maybe just add that to the release notes (that the new default version is 1.9.0) ?

PR #802, opened 2015-10-09
Use DOCKER_CONFIG environment variable to look up auth config

shin-:
When provided, default paths (`~/.docker/config.json` and `~/.dockercfg`) are ignored.

### DIFF
--- a/docker/auth/auth.py
+++ b/docker/auth/auth.py
@@ -13,7 +13,6 @@
 #    limitations under the License.
 
 import base64
-import fileinput
 import json
 import logging
 import os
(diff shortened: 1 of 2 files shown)


docker/auth/auth.py
```diff
@@ -136,10 +136,19 @@ def load_config(config_path=None):
...
 
+    if not config_path and os.environ.get('DOCKER_CONFIG'):
+        config_path = os.path.join(
+            os.environ.get('DOCKER_CONFIG'),
+            os.path.basename(DOCKER_CONFIG_FILENAME)
+        )
+
     # Prefer ~/.docker/config.json.
     config_file = config_path or os.path.join(os.path.expanduser('~'),
                                               DOCKER_CONFIG_FILENAME)
```
dnephin: I'd like to make a suggestion here that I think will make it easier to test.  I think the logic for "find config file" can be separated from "read contents of config file".

```python

def find_config(config_path=None):
    for path in (
        config_path, 
        os.path.join(os.environ.get('DOCKER_CONFIG'), os.path.basename(DOCKER_CONFIG_FILENAME)),
        os.path.join(os.path.expanduser('~'), DOCKER_CONFIG_FILENAME),
        os.path.exists(os.path.join(os.path.expanduser('~'), LEGACY_DOCKER_CONFIG_FILENAME))
    ):
        if path and os.path.exists(path):
            return path

    return None
```

tests/test.py
```diff
@@ -2423,6 +2423,37 @@ def test_load_config_with_random_name(self):
...
+                'email': 'sakuya@scarlet.net'
+            }
+        }
+
+        with open(dockercfg_path, 'w') as f:
+            json.dump(config, f)
+
+        environ_mem = os.environ.copy()
+        try:
+            os.environ['DOCKER_CONFIG'] = folder
```
dnephin: This could be done with 

```python

with mock.patch.dict(os.environ):
    ...

```

shin-:
Implemented @dnephin 's suggestions and rebased.

dnephin:
LGTM

PR #806, opened 2015-10-09
Use url-safe base64 when encoding auth header

### LINKED ISSUE #803: Auth fails with long passwords
See https://github.com/docker/docker/issues/16840

docker-py is encoding `X-Registry-Auth` with regular base64 and not the url safe version of base64 that jwt tokens use.

shin-:
Fixes #803

### DIFF
--- a/docker/auth/auth.py
+++ b/docker/auth/auth.py
@@ -102,7 +102,7 @@ def decode_auth(auth):
 
 def encode_header(auth):
     auth_json = json.dumps(auth).encode('ascii')
-    return base64.b64encode(auth_json)
+    return base64.urlsafe_b64encode(auth_json)
 
 
 def parse_auth(entries):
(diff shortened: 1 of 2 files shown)


tests/base.py
```diff
@@ -13,6 +13,11 @@ def assertIn(self, object, collection):
             return self.assertTrue(object in collection)
         return super(BaseTestCase, self).assertIn(object, collection)
 
+    def assertNotIn(self, object, collection):
+        if six.PY2 and sys.version_info[1] <= 6:
+            return self.assertTrue(object not in collection)
+        return super(BaseTestCase, self).assertNotIn(object, collection)
+
```
dnephin: Now that the tests use py.test, you could just use `assert x not in y` and get a really nice error message when it fails! (instead of having this wrapper)

dnephin:
some CI failures, otherwise looks good

PR #808, opened 2015-10-14
Docs examples

alasdairnicol:
A couple of minor tidy ups to the examples in the docs, which make them more consistent, and avoid deprecation warnings.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -4,8 +4,8 @@ To instantiate a `Client` class that will allow you to communicate with a
 Docker daemon, simply do:
 
 ```python
-from docker import Client
-c = Client(base_url='unix://var/run/docker.sock')
+>>> from docker import Client
+>>> cli = Client(base_url='unix://var/run/docker.sock')
 ```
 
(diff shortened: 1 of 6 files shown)


alasdairnicol:
My initial patch was wrong, because `create_host_config` is a regular method, not a classmethod. I've fixed the patch and pushed it again.

shin-:
Thanks! Can you squash those into one single commit?

On a tangential note, I'm not the biggest fan of calling the client instance `cli` because of the "Command Line Interface" acronym. But we can take another pass at this later - it being consistent means it's easier to `sed` later down the line.

alasdairnicol:
Thanks for the quick review! I've squashed into a single commit and rebased.

> I'm not the biggest fan of calling the client instance cli because of the "Command Line Interface" acronym. 

Sure, I understand. I chose `cli` because it appeared to be the most common variable name, not for any other reason.

shin-:
LGTM!

dnephin:
LGTM

PR #809, opened 2015-10-16
Reorganize tests

shin-:
Big cleanup / reorganization of our tests into several files


dnephin:
I didn't look over everything, but I assume it's all just moving stuff around.

I think the structure looks great, but there is a test failure on janky. I think something needs to be updated for the new structure.

shin-:
Yup, I was trying to import the integration/ module in the integration_test.py file to help transition but that doesn't work as I expected to. Will update soon. :)

PR #812, opened 2015-10-19
Fix Unix adapter bug with newer versions of requests

### LINKED ISSUE #811: requests 2.8.0 re-introduces bug #647

shin-:
The select_proxy utility in requests errors out when the provided URL
doesn't have a hostname, like is the case when using a UNIX socket.
Since proxies are an irrelevant notion in the case of UNIX sockets
anyway, we simply return the path URL directly.

Fixes #811


dnephin:
LGTM

PR #814, opened 2015-10-21
Add support for MemorySwappiness to create_host_config

lots0logs:


shin-:
Thanks!

The tests are failing because you have trailing whitespace and an indentation error in your patch. You can run `make flake8` in the project's root to detect those locally.

Apart from the code style nitpicking:

- It seems the option was added in API version 1.20 - can we add a client-side check for that? Just like we do for group_add [here](https://github.com/docker/docker-py/blob/master/docker/utils/utils.py#L541-L544)
- The parameter should be added to the HostConfig documentation (`docs/host_config.md`).
- It might also be valuable to add an integration test that verifies the value we pass is reflected in the `inspect_container` dict.

Issue #815, opened 2015-10-21
When creating a network allow more options


rmb938:
When creating a network allow more options that are not currently documented.

The current docker API allows for these options:

```
{
  "name": str,
  "driver": str,
  "ipam": {
    "driver": str.
    "config": [{
      "subnet": str,
      "iprange": str,
      "gateway": str,
      "auxaddresses": map
    }]
  },
  "options": map
}
```

shin-:
Closing via #818

PR #818, opened 2015-10-21
allow custom ipam options when creating networks

rmb938:
This addresses issue #815 

These options are currently not documented but are still useful when creating networks.

### DIFF
--- a/docker/api/network.py
+++ b/docker/api/network.py
@@ -19,14 +19,15 @@ class NetworkApiMixin(object):
         return self._result(res, json=True)
 
     @minimum_version('1.21')
-    def create_network(self, name, driver=None, options=None):
+    def create_network(self, name, driver=None, options=None, ipam=None):
         if options is not None and not isinstance(options, dict):
             raise TypeError('options must be a dictionary')
 
         data = {
             'name': name,
             'driver': driver,
-            'options': options
+            'options': options,
+            'ipam': ipam,
         }
         url = self._url("/networks/create")
         res = self._post_json(url, data=data)
--- a/docker/utils/utils.py
+++ b/docker/utils/utils.py
@@ -44,6 +44,23 @@ BYTE_UNITS = {
 }
 
 
+def create_ipam_pool(subnet=None, iprange=None, gateway=None,
+                     aux_addresses=None):
+    return {
+        'subnet': subnet,
+        'iprange': iprange,
+        'gateway': gateway,
+        'auxaddresses': aux_addresses
+    }
+
+
+def create_ipam_config(driver='default', pool_configs=None):
+    return {
+        'driver': driver,
+        'config': pool_configs or []
+    }
+
+
 def mkbuildcontext(dockerfile):
     f = tempfile.NamedTemporaryFile()
     t = tarfile.open(mode='w', fileobj=f)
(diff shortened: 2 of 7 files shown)


docker/api/network.py
```diff
@@ -19,16 +19,30 @@ def networks(self, names=None, ids=None):
...
             'name': name,
             'driver': driver,
+            'ipam': ipam,
         }
         url = self._url("/networks/create")
         res = self._post_json(url, data=data)
         return self._result(res, json=True)
 
     @minimum_version('1.21')
+    def generate_ipam_config(self, driver='default', subnet=None,
```
shin-: Since this doesn't need to exist inside the `Client` (doesn't use `self`), I think it should be moved to a function inside the `docker.utils` module instead.

shin-:
Thanks for contributing!

I'm going to wait for the 1.9.0 release to merge this to make sure things don't change last minute.

In the meantime, having some tests and mentioning the added methods in the documentation would be a valuable improvement.

docker/api/network.py
rmb938: That makes sense, moving it now.

rmb938:
Sounds good. Will start working on those.

docker/api/network.py
shin-: Also, it would be nice to have it renamed to `create_ipam_config` to be consistent with the currently existing `create_host_config` and `create_container_config`. :smile:

shin-:
Hi @rmb938 , do you mind rebasing this? We can take a closer look now that 1.9 is out.

rmb938:
@shin- Yup no problem. Sorry for the late reply I will get it done this weekend.

rmb938:
@shin- sorry for the delay. I got everything rebased. if it all looks good I can squash all the commits for you.

docker/utils/utils.py
```diff
@@ -44,6 +44,25 @@
...
+        'auxaddresses': aux_addresses
+    }
+
+
+def create_ipam_config(driver='default', pool_configs=None):
+    if not pool_configs:
+        pool_configs = []
+    return {
+        'driver': driver,
+        'config': pool_configs
```
shin-: `pool_configs or []` would allow us to get rid of the first two lines here. Small readability improvement.

docs/networks.md
```diff
@@ -0,0 +1,20 @@
+# Using Networks
```
shin-: This file needs to be added to the pages list in `mkdocs.yml`

shin-: Also, the other documentation files are written using "you" instead of "we". This needs to use the same perspective.

docs/networks.md
```diff
@@ -0,0 +1,20 @@
...
+With the release of Docker 1.9 we can now manage custom networks.
+
+
+Here we can see how to create a network named ```network1``` using the ```bridge``` driver
+
+```python
+docker_client.create_network("network1", driver="bdige")
+```
+
+We can also create more advanced networks with custom ipam configuration were we set the subnet
```
shin-: Can we reformulate that sentence? I would go with:

> You can also create more advanced networks with custom IPAM configurations. For example, setting the subnet to `192.168.52.0/24` and gateway to `192.168.52.254`:

shin-:
Thanks for updating. A few things to change in the docs and a small nitpick in the `create_ipam_config` function, then we should be good.

Feel free to squash it as well :)

rmb938:
@shin- how does that look

shin-:
Thanks, looking great - LGTM!

aanand:
LGTM

PR #822, opened 2015-10-21
Remove support for host_config in Client.start

### LINKED ISSUE #750: Passing host_config parameters in start() overrides the host_config that was passed in create()?
I had a `host_config` with `extra_hosts` defined. I used this `host_config` to create a container. When starting container, I passed extra volumes_from parameter. However, this lead to `extra_hosts` seemingly not doing their job. 

After I moved `volumes_from` from `start()` to `create_host_config` everything started working OK. 

I figured, I should report this, since I spent a couple of hours figuring this out. IMO, start() should override the necessary parts of host_config(and not the whole thing). Or raise an exception.

shin-:
Any additional arguments passed to start will raise a `DeprecatedMethod` (`DockerException`) exception.

It is time!
- The `DeprecationWarning` has been present in docker-py since version `1.2.3` which was released on June 18.
- On the docker engine, this has been soft deprecated since API version `1.15`, which was released along docker `1.3.0`. This version of the engine was released on October 16th, 2014.
- WIth the hub dropping support for clients < 1.6, it seems reasonable for us to drop support for versions older than `1.3.0`. People stuck on older versions of docker for some reason can use `1.5.x` which is stable

This might warrant bumping the version to `2.0.0`

Closes #750

### DIFF
--- a/docker/api/container.py
+++ b/docker/api/container.py
@@ -997,19 +997,16 @@ class ContainerApiMixin(object):
         self._raise_for_status(res)
 
     @utils.check_resource
-    def start(self, container, binds=None, port_bindings=None, lxc_conf=None,
-              publish_all_ports=None, links=None, privileged=None,
-              dns=None, dns_search=None, volumes_from=None, network_mode=None,
-              restart_policy=None, cap_add=None, cap_drop=None, devices=None,
-              extra_hosts=None, read_only=None, pid_mode=None, ipc_mode=None,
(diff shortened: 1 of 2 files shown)


dnephin:
Might be worth thinking about any other breaking changes you'd like to make before doing a 2.0 release.  Sounds reasonable to me.

bfirsh:
LGTM

bfirsh:
rebased!

PR #824, opened 2015-10-26
Support unicode commands

dnephin:
While debugging https://github.com/docker/compose/issues/2255 I can across this error on python2.

The docs claim that python >= 2.7.3 will support unicode in `shlex.split()`, but that is a lie in many cases because it attempts to import and use `cStringIO` which does not support unicode.


shin-:
LGTM!

PR #825, opened 2015-10-27
Fix auth.load_config bug calling parse_auth

aebm:
When load_config found and auths section it didn't call parse_auth only with
the auths section. Instead it called parse_auth with all the configuration.

There is also a test to check this case

Signed-off-by: Alejandro Brito Monedero <alejandro.monedero@gmail.com>


shin-:
Thanks! LGTM.

dnephin:
LGTM

PR #826, opened 2015-10-27
Fix #627

### LINKED ISSUE #627: decode_auth function does not handle utf-8 logins or password
HI

I have found that the function **decode_auth** (line 96, [file](https://github.com/docker/docker-py/blob/master/docker/auth/auth.py)) fails when decoding UTF-8 passwords from the .dockercfg file, and **load_config** returning an empty config.

I have checked and docker hub can handle UTF-8 passwords, this code proves that:
```python
# coding=utf-8
from docker import Client
cred = { 'username': <user>, 'password': <utf-8 password> }
c = Client(base_url='unix://var/run/docker.sock')
res = c.pull(repository='<private container>', tag='latest', auth_config=cred)
print(res)
```

Thank you

aebm:
Docker-py couldn't pull private images if your account have non ascii chars.
Instead docker pull command, didn't have that issue.

Also add a test to check when you have a login or password with an non ascii
char.

Signed-off-by: Alejandro Brito Monedero <alejandro.monedero@gmail.com>

### DIFF
--- a/docker/auth/auth.py
+++ b/docker/auth/auth.py
@@ -96,7 +96,7 @@ def decode_auth(auth):
         auth = auth.encode('ascii')
     s = base64.b64decode(auth)
     login, pwd = s.split(b':', 1)
-    return login.decode('ascii'), pwd.decode('ascii')
+    return login.decode('utf8'), pwd.decode('utf8')
 
 
 def encode_header(auth):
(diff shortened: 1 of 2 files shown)


dnephin:
Very interesting.  How would this fail if there was unicode? Would it crash or just fail to authenticate?

aebm:
Hi,

It would throw an decoding exception that is logged at the debug level  in load_config, ending with load_config returning a empty dict. When pulling the image you don't have auth and the error shown is that the image doesn't exists

aebm:
Hi,

I will close it to open it again to fix the tests

PR #827, opened 2015-10-27
Move container volume bind tests to their own class

coreypobrien:
The ```setUp``` in CreateContainerTest was entirely for the two binding tests (```test_create_with_binds_rw``` and ```test_create_with_binds_ro```). Moved those tests to their own class along with ```setUp``` and their two helper functions.


dnephin:
LGTM 

less setup is faster tests!

shin-:
Thanks!

PR #830, opened 2015-10-28
Use network_mode=bridge when testing links

shin-:
Links are only compatible with network mode bridge. New versions of docker only create env entries when network mode is properly set.


shin-:
RIP janky :angel: :rose:

dnephin:
LGTM

PR #831, opened 2015-10-28
Fix test issues brought on by changes in inspect results

shin-:
docker/docker#17393 changes the network settings struct a bit, which impacts tests checking values in inspect results.

### DIFF
--- a/tests/integration/container_test.py
+++ b/tests/integration/container_test.py
@@ -263,8 +263,16 @@ class CreateContainerTest(api_test.BaseTestCase):
 
         self.client.start(container)
         res = self.client.inspect_container(container['Id'])
-        self.assertEqual(mac_address_expected,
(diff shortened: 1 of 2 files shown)


tests/integration/container_test.py
```diff
@@ -263,8 +263,16 @@ def test_create_with_mac_address(self):
...
         self.client.start(container)
         res = self.client.inspect_container(container['Id'])
-        self.assertEqual(mac_address_expected,
-                         res['NetworkSettings']['MacAddress'])
+        try:
+            self.assertEqual(
+                mac_address_expected,
+                res['NetworkSettings']['Networks']['bridge']['MacAddress']
+            )
+        except KeyError:
```
dnephin: I think a comment about why this needs to exist would be great (so we don't forget). Something about which API version would hit which branch.

That way we know when we can remove it as well.

Issue #837, opened 2015-11-04
docker build: support --build-arg argument


leonty:
Docker 1.9 offers a new command line parameter to customize the Dockerfile at the build time:

The ARG instruction defines a variable that users can pass at build-time to the builder with the docker build command using the --build-arg = flag.
https://docs.docker.com/engine/reference/builder/#arg

leonty:
Hi shin, I have a ready PR for this... just have stuck with the signing...

leonty:
Done!

shin-:
The PR for this has been merged - closing.

PR #838, opened 2015-11-04
Added support for build arguments.

leonty:
This implements the feature #837

### DIFF
--- a/tests/integration/build_test.py
+++ b/tests/integration/build_test.py
@@ -96,3 +96,18 @@ class BuildTest(api_test.BaseTestCase):
             list(filter(None, logs.split('\n'))),
             ['not-ignored'],
         )
+
+    @requires_api_version('1.21')
+    def test_build_with_buildargs(self):
+        script = io.BytesIO('\n'.join([
+            'FROM scratch',
+            'ARG test',
+            'USER $test'
+        ]).encode('ascii'))
+
+        self.client.build(fileobj=script, tag='buildargs',
+                          buildargs={'test': 'OK'}, stream=True)
+        self.tmp_imgs.append('buildargs')
+
+        info = self.client.inspect_image('buildargs')
+        self.assertEqual(info['Config']['User'], 'OK')
--- a/docker/api/build.py
+++ b/docker/api/build.py
@@ -1,6 +1,7 @@
 import logging
 import os
 import re
+import json
 
 from .. import constants
 from .. import errors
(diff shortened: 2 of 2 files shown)


leonty:
Done!

leonty:
@shin what's the due date for the 1.6.0 milestone?

dnephin:
Thanks for the PR!

A test for this case would be great.

I don't think we have a set release date for the 1.6.0 milestone.

leonty:
@dnephin voilà

tests/integration/build_test.py
```diff
@@ -96,3 +96,16 @@ def test_build_with_dockerignore(self):
...
+    @requires_api_version('1.9')
+    def test_build_with_buildargs(self):
+        script = io.BytesIO('\n'.join([
+            'FROM scratch',
+            'ARG test',
+            'USER $test'
+        ]).encode('ascii'))
+
+        self.client.build(fileobj=script, tag='buildargs',
+                          buildargs={'test': 'OK'}, stream=True)
```
shin-: The created image needs to be cleaned up after the test has finished running. `self.tmp_imgs.append('buildargs')` should do the trick.

shin-:
Thanks!

One small comment on the test. Additionally, can you squash those commits into one? Once that's done, LGTM.

leonty:
@shin- no problem. But I used the existing tests as a reference and they don't clean anything after (all integration build tests)...

leonty:
@shin- done. Though I'm not sure the squash was correct, have never done it before

leonty:
Hi, is this accepeted? I'm about to push another PR for Ansible to use this functionality and just waiting to know for sure that this API will be in 1.6.

shin-:
Thanks! Looking back there's a few things still.

1. Your commit is missing the signoff at the bottom.
2. The test should require version `1.21` (the latest API version) instead of `1.9` (the engine version).
3. In `Client.build`, if `buildargs` is provided, we should check the version is greater or equal `1.21` (`utils.version_gte` is a good way to do that) and raise an `errors.InvalidVersion` if it isn't.

leonty:
Done. But I don't know why the integration test has failed, for me locally it works:
```
>>> import docker, io
>>> c = docker.Client('127.0.0.1:4243')
>>> script = io.BytesIO('\n'.join([
            'FROM scratch',
            'ARG test',
            'USER $test'
        ]).encode('ascii'))
>>> c.build(fileobj=script, tag='buildargs', buildargs={'test': 'OK'}, stream=True)
>>> info = c.inspect_image('buildargs')
>>> info
{'Os': 'linux', 'Parent': 'bc2b3187f9a63a6bc427c2b6a80675ec4885a4262d8f8ed7b04ba14c9b848ce2', 'RepoTags': ['buildargs:latest'], 'Config': {'Hostname': '4ee80013
...
```

shin-:
There's a race condition where your test is progressing before the build operation is finished on the server side. I've updated the PR and submitted it as #846 . Let's follow up there.

leonty:
Ok, thanks.

shin-:
Merged into master :+1:

leonty:
Great! Thanks again.

PR #841, opened 2015-11-08
Ensure assert_hostname is set on the pool connection

delfick:
If you have assert_hostname turned off and are using a proxy on your
computer then assert_hostname wasn't being properly set on the pool
connection and it was failing to connect to docker

### DIFF
--- a/docker/ssladapter/ssladapter.py
+++ b/docker/ssladapter/ssladapter.py
@@ -46,6 +46,19 @@ class SSLAdapter(HTTPAdapter):
 
         self.poolmanager = PoolManager(**kwargs)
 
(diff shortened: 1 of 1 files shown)


delfick:
https://github.com/kennethreitz/requests/blob/master/requests/adapters.py#L253 is the block that uses something other than self.poolmanager to create the connection

shin-:
LGTM :+1:

dnephin:
Code change looks ok to me.  

Do you think this is a bug in `requests` ? I'm not sure I understand why we have to handle this

delfick:
I agree that it would be nice if requests actually passed assert_hostname around.

I'm not sure how that change would look though.

PR #843, opened 2015-11-09
create_host_config: allow setting oom_kill_disable

schu:


docker/utils/utils.py
```diff
@@ -511,6 +511,9 @@ def create_host_config(
     if privileged:
         host_config['Privileged'] = privileged
 
+    if oom_kill_disable:
```
shin-: Can we test that `version_gte(version, '1.19') is True and raise an `InvalidVersion` otherwise?

shin-:
Thanks, looks good! Just a small request to make it perfect :+1:

docker/utils/utils.py
schu: Done.

shin-:
Thanks!

LGTM.

schu:
Thanks!

PR #848, opened 2015-11-11
Client.logs(tail=0) now shows no past logs (expected behavior)

### LINKED ISSUE #845: docker.Client.logs(tail=0) returns all logs instead of no logs
This is a confusing behavior.  It also makes it harder to follow a container's logs (with `stream=True`) without getting at least some old log data.  A work-around is required, specifying `tail=1` and then discarding the first line of output.

shin-:
Add test confirming the new behavior.

Fixes #845

### DIFF
--- a/tests/integration/container_test.py
+++ b/tests/integration/container_test.py
@@ -613,8 +613,8 @@ class LogsTest(api_test.BaseTestCase):
             BUSYBOX, 'echo {0}'.format(snippet)
         )
         id = container['Id']
-        self.client.start(id)
         self.tmp_containers.append(id)
+        self.client.start(id)
         exitcode = self.client.wait(id)
         self.assertEqual(exitcode, 0)
         logs = self.client.logs(id)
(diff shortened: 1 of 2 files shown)


tests/integration/container_test.py
```diff
@@ -632,24 +632,24 @@ def test_logs_tail_option(self):
...
-#         for chunk in self.client.logs(id, stream=True):
-#             logs += chunk
+    def test_logs_streaming(self):
+        snippet = 'Flowering Nights (Sakuya Iyazoi)'
+        container = self.client.create_container(
+            BUSYBOX, 'echo {0}'.format(snippet)
+        )
+        id = container['Id']
+        self.client.start(id)
+        self.tmp_containers.append(id)
```
dnephin: minor: I think adding it to tmp_containers should happen before self.client.start(), otherwise start failures would leave a created container lying around

tests/integration/container_test.py
```diff
@@ -632,24 +632,24 @@ def test_logs_tail_option(self):
...
+        logs = bytes() if six.PY3 else str()
```
dnephin: very minor: this is equivalent to `logs = six.binary_type()`

tests/integration/container_test.py
```diff
@@ -664,6 +664,19 @@ def test_logs_with_dict_instead_of_id(self):
...
         self.assertEqual(logs, (snippet + '\n').encode(encoding='ascii'))
 
+    def test_logs_with_tail_0(self):
+        snippet = 'Flowering Nights (Sakuya Iyazoi)'
+        container = self.client.create_container(
+            BUSYBOX, 'echo "{0}"'.format(snippet)
+        )
+        id = container['Id']
+        self.client.start(id)
+        self.tmp_containers.append(id)
```
dnephin: same minor ordering issue here

dnephin:
LGTM, some minor suggestions, but no blockers

PR #851, opened 2015-11-12
Remove unused files

shin-:


tests/integration/__init__.py
```diff
@@ -1,13 +0,0 @@
...
-# Remove imports when merged in master and Jenkins is updated to find the
-# tests in the new location.
-from .api_test import *
-from .build_test import *
-from .container_test import *
-from .exec_test import *
-from .image_test import *
-from .network_test import *
-from .regression_test import *
-from .volume_test import *
```
dnephin: Don't you still need an empty __init__.py here for the tests to be considered python modules?

shin-: Yep. It's still there :)

shin-: https://github.com/docker/docker-py/blob/remove-ci-transition-files/tests/integration/__init__.py

dnephin: Ah, I should have looked at the expanded changed files. cool

dnephin:
LGTM

PR #853, opened 2015-11-14
Implement API for interactive exec

TomasTomecek:
Needed for: https://github.com/d11wtq/dockerpty/issues/30
Transitively for: https://github.com/docker/compose/pull/2023

### DIFF
--- a/tests/integration/exec_test.py
+++ b/tests/integration/exec_test.py
@@ -1,13 +1,13 @@
 import pytest
 
-from . import api_test
+from .. import helpers
 
-BUSYBOX = api_test.BUSYBOX
+BUSYBOX = helpers.BUSYBOX
 
 
-class ExecTest(api_test.BaseTestCase):
+class ExecTest(helpers.BaseTestCase):
     def test_execute_command(self):
-        if not api_test.exec_driver_is_native():
+        if not helpers.exec_driver_is_native():
             pytest.skip('Exec driver not native')
 
         container = self.client.create_container(BUSYBOX, 'cat',
@@ -23,7 +23,7 @@ class ExecTest(api_test.BaseTestCase):
         self.assertEqual(exec_log, b'hello\n')
 
     def test_exec_command_string(self):
-        if not api_test.exec_driver_is_native():
+        if not helpers.exec_driver_is_native():
             pytest.skip('Exec driver not native')
 
         container = self.client.create_container(BUSYBOX, 'cat',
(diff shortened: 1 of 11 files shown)


docker/api/exec_api.py
```diff
@@ -65,4 +64,15 @@ def exec_start(self, exec_id, detach=False, tty=False, stream=False):
...
+    @utils.minimum_version('1.15')
+    def exec_start(self, exec_id, detach=False, tty=False, stream=False):
+        res = self._exec_start(exec_id, detach=detach, tty=tty, stream=stream)
         return self._get_result_tty(stream, res, tty)
+
+    @utils.minimum_version('1.15')
+    def exec_start_socket(self, exec_id, detach=False, tty=False,
+                          stream=False):
+        res = self._exec_start(exec_id, detach=detach, tty=tty, stream=stream)
+        return self._get_raw_response_socket(res)
```
dnephin: This does follow the pattern established by `attach()`/`attach_socket()` but I wonder if it could be just another kwarg like stream?

What do you think @shin-  ?

```python
def exec_start(self, exec_id, detach=False, tty=False, stream=False, socket=False):
    ...
    func = get_response_handler(stream, socket)
    return func(res)

def get_response_handler(stream, socket):
    if stream and socket:
        raise ValueError(...)

    if stream:
        return functools.partial(self._get_result_tty, True)
    if socket:
        return  self._get_raw_response_socket
    return  functools.partial(self._get_result_tty, False)
```

tests/integration/exec_test.py
```diff
@@ -88,6 +93,66 @@ def test_exec_command_streaming(self):
...
 
+    def test_exec_start_socket(self):
+        if not api_test.exec_driver_is_native():
+            pytest.skip('Exec driver not native')
+
+        container = self.client.create_container(BUSYBOX, 'cat',
+                                                 detach=True, stdin_open=True)
+        id = container['Id']
+        self.client.start(id)
+        self.tmp_containers.append(id)
```
dnephin: this `self.tmp_containers.append(id)` line should come before `self.client.start(id)` so that the container gets cleaned up if start fails.

docker/api/exec_api.py
TomasTomecek: Makes sense to me. To be honest, I actually vote for `socket=False` since there is already an option for asking for stream. On top of it, you'll get this feature for free in future (for new similar API calls).

tests/integration/exec_test.py
TomasTomecek: I [copy-pasted](https://github.com/docker/docker-py/blob/4e441b2a6bbac7b9b43adca05800c8f46e18ec16/tests/integration/container_test.py#L955) from `attach container` test. Will fix.

docker/api/exec_api.py
TomasTomecek: @shin- waiting for your reply before I start rewriting.

shin-: Yes, `socket` as a param makes a lot of sense. :+1:

PR #858, opened 2015-11-23
Improve exec api

TomasTomecek:
I'm sorry for creating new PR but git/GitHub went absolutely bananas: gh closed old PR itself (new UI glitch?), messed up commits and refused to do valid rebase.

Old PR: #853 

Needed for: https://github.com/d11wtq/dockerpty/issues/30
Transitively for: https://github.com/docker/compose/pull/2023

I have addressed all comments:
 * correctly configured garbage collection during integration test
 * created new interface for stream/socket response (chose decorator instead of `Client` method) -- can rewrite if you don't like this (the reason I did it this way is that I didn't like the solution with `partial` b/c you need to supply more arguments to `_get_result_tty` than to `_get_raw_response_socket`)

### DIFF
--- a/docker/api/exec_api.py
+++ b/docker/api/exec_api.py
@@ -7,8 +7,8 @@ from .. import utils
 class ExecApiMixin(object):
     @utils.minimum_version('1.15')
     @utils.check_resource
-    def exec_create(self, container, cmd, stdout=True, stderr=True, tty=False,
-                    privileged=False, user=''):
+    def exec_create(self, container, cmd, stdout=True, stderr=True,
+                    stdin=False, tty=False, privileged=False, user=''):
         if privileged and utils.compare_version('1.19', self._version) < 0:
             raise errors.InvalidVersion(
                 'Privileged exec is not supported in API < 1.19'
@@ -25,7 +25,7 @@ class ExecApiMixin(object):
             'User': user,
             'Privileged': privileged,
             'Tty': tty,
-            'AttachStdin': False,
+            'AttachStdin': stdin,
             'AttachStdout': stdout,
             'AttachStderr': stderr,
             'Cmd': cmd
@@ -53,7 +53,11 @@ class ExecApiMixin(object):
         self._raise_for_status(res)
 
     @utils.minimum_version('1.15')
-    def exec_start(self, exec_id, detach=False, tty=False, stream=False):
+    def exec_start(self, exec_id, detach=False, tty=False, stream=False,
+                   socket=False):
+        # we want opened socket if socket == True
+        if socket:
+            stream = True
         if isinstance(exec_id, dict):
             exec_id = exec_id.get('Id')
 
@@ -65,4 +69,7 @@ class ExecApiMixin(object):
         res = self._post_json(
             self._url('/exec/{0}/start', exec_id), data=data, stream=stream
         )
+
+        if socket:
+            return self._get_raw_response_socket(res)
         return self._get_result_tty(stream, res, tty)
--- a/tests/integration/exec_test.py
+++ b/tests/integration/exec_test.py
@@ -77,8 +77,8 @@ class ExecTest(helpers.BaseTestCase):
         container = self.client.create_container(BUSYBOX, 'cat',
                                                  detach=True, stdin_open=True)
         id = container['Id']
-        self.client.start(id)
         self.tmp_containers.append(id)
+        self.client.start(id)
 
         exec_id = self.client.exec_create(id, ['echo', 'hello\nworld'])
         self.assertIn('Id', exec_id)
@@ -87,6 +87,30 @@ class ExecTest(helpers.BaseTestCase):
         for chunk in self.client.exec_start(exec_id, stream=True):
             res += chunk
         self.assertEqual(res, b'hello\nworld\n')
+
+    def test_exec_start_socket(self):
+        if not helpers.exec_driver_is_native():
+            pytest.skip('Exec driver not native')
+
+        container = self.client.create_container(BUSYBOX, 'cat',
+                                                 detach=True, stdin_open=True)
+        container_id = container['Id']
+        self.client.start(container_id)
+        self.tmp_containers.append(container_id)
+
+        line = 'yay, interactive exec!'
+        # `echo` appends CRLF, `printf` doesn't
+        exec_id = self.client.exec_create(
+            container_id, ['printf', line], tty=True)
+        self.assertIn('Id', exec_id)
+
+        socket = self.client.exec_start(exec_id, socket=True)
+        self.addCleanup(socket.close)
+
+        next_size = helpers.next_packet_size(socket)
+        self.assertEqual(next_size, len(line))
+        data = helpers.read_data(socket, next_size)
+        self.assertEqual(data.decode('utf-8'), line)
 
     def test_exec_inspect(self):
         if not helpers.exec_driver_is_native():
--- a/Makefile
+++ b/Makefile
@@ -31,6 +31,7 @@ integration-test-py3: build-py3
 	docker run -v /var/run/docker.sock:/var/run/docker.sock docker-py3 py.test tests/integration
 
 integration-dind: build build-py3
+	docker rm -vf dpy-dind || :
 	docker run -d --name dpy-dind --env="DOCKER_HOST=tcp://localhost:2375" --privileged dockerswarm/dind:1.9.0 docker -d -H tcp://0.0.0.0:2375
 	docker run --env="DOCKER_HOST=tcp://docker:2375" --link=dpy-dind:docker docker-py py.test tests/integration
 	docker run --env="DOCKER_HOST=tcp://docker:2375" --link=dpy-dind:docker docker-py3 py.test tests/integration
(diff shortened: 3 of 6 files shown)


docker/api/exec_api.py
```diff
@@ -53,7 +53,8 @@ def exec_resize(self, exec_id, height=None, width=None):
         self._raise_for_status(res)
 
     @utils.minimum_version('1.15')
-    def exec_start(self, exec_id, detach=False, tty=False, stream=False):
+    @utils.handle_interactive_response
+    def exec_start(self, exec_id, detach=False, tty=False, stream=False, socket=False):
```
shin-: This line is longer than 80 characters :(

docker/utils/decorators.py
```diff
@@ -35,3 +35,18 @@ def wrapper(self, *args, **kwargs):
             return f(self, *args, **kwargs)
         return wrapper
     return decorator
+
+
+def handle_interactive_response(f):
```
shin-: Is this intended to be used anywhere else?

TomasTomecek: Hm, since `_get_raw_response_socket` is used only in `attach_socket`, it doesn't make sense to use it anywhere else now. Not sure if it could be useable in future. Let me know if I should merge the functionality into `exec_start` directly or leave it as it is now.

TomasTomecek:
Regarding the integration test, it works fine for me locally:
```
platform linux2 -- Python 2.7.10, pytest-2.8.2, py-1.4.30, pluggy-0.3.1
plugins: capturelog-0.7
collected 8 items 

tests/integration/exec_test.py .

```
```
docker 1.9.0
```
Is this a regression between 1.8 and 1.9?

shin-:
The integration tests on Jenkins use Docker 1.9.0 as well (see the `integration-dind` task in `Makefile`). I also get successes testing outside of `dind`. I'll take a closer look at the test.

TomasTomecek:
I think that the issue could be with terminal allocation, since test runner probably doesn't allocate tty in kernel.

TomasTomecek:
After doing some brief testing the issue with integration test seems like race:

First run:

```
tests/integration/exec_test.py:141: in test_exec_start_socket
    self.assertEqual(next_size, len(line))
E   AssertionError: 0 != 22
```
```
connectionpool.py          383 DEBUG    "POST /v1.20/containers/create HTTP/1.1" 201 90
connectionpool.py          383 DEBUG    "POST /v1.20/containers/0d8a59766345c5724aabd31a4aacb259b85cb2be2f9d21bf88a8277ff7369471/start HTTP/1.1" 204 0
connectionpool.py          383 DEBUG    "POST /v1.20/containers/0d8a59766345c5724aabd31a4aacb259b85cb2be2f9d21bf88a8277ff7369471/exec HTTP/1.1" 201 74
connectionpool.py          383 DEBUG    "POST /v1.20/exec/63e7bcc6574a53c01e493079c51103ca6a8697eea14b821a098384b956224c8d/start HTTP/1.1" 200 None
connectionpool.py          383 DEBUG    "POST /v1.20/containers/0d8a59766345c5724aabd31a4aacb259b85cb2be2f9d21bf88a8277ff7369471/stop?t=1 HTTP/1.1" 204 0
connectionpool.py          383 DEBUG    "DELETE /v1.20/containers/0d8a59766345c5724aabd31a4aacb259b85cb2be2f9d21bf88a8277ff7369471?link=False&force=False&v=False HTTP/1.1" 204 0
```

Then I reran the `exec` suite commenting out `self.assertEqual(next_size, len(line))` just to see where it gets:
```
platform linux2 -- Python 2.7.10 -- py-1.4.30 -- pytest-2.6.4
plugins: capturelog
collected 7 items 

tests/integration/exec_test.py .....F.


tests/integration/exec_test.py:149: in test_exec_start_socket
    self.assertEqual(data.decode('utf-8'), "{0}".format(line))
E   AssertionError: u'yay, interactive exec!\r\n' != 'yay, interactive exec!'
```

And again just a single failing test:

```
$ py.test tests/integration/exec_test.py::ExecTest::test_exec_start_socket

platform linux2 -- Python 2.7.10 -- py-1.4.30 -- pytest-2.6.4
plugins: capturelog
collected 8 items 

tests/integration/exec_test.py .

```

And now it works for me locally.

TomasTomecek:
Okay, so this is a python 3 only issue.

TomasTomecek:
Done!

It was a race. Command in container didn't finish before tests tried to fetch results. Simple `select([socket], [], [])` solves it.

TomasTomecek:
* got rid of the decorator
* moved `import select` to top

docker/api/exec_api.py
```diff
@@ -62,7 +63,13 @@ def exec_start(self, exec_id, detach=False, tty=False, stream=False):
             'Detach': detach
         }
 
+        # socket = True implies stream = True otherwise you'll get:
+        #   AttributeError: 'NoneType' object has no attribute '_sock'
+        stream = True if socket else stream
```
aanand: This is a bit harder to read than it needs to be. How about just:

```python
if socket:
    stream = True
```

Also, this is just a personal preference, but I prefer to modify arguments at the very top of a function, so it's all in one place, so ideally this would go above the bit where we initialise `data`.

tests/integration/exec_test.py
```diff
@@ -88,6 +94,67 @@ def test_exec_command_streaming(self):
             res += chunk
         self.assertEqual(res, b'hello\nworld\n')
 
+    def test_exec_start_socket(self):
+        if not helpers.exec_driver_is_native():
+            pytest.skip('Exec driver not native')
+
+        container = self.client.create_container(BUSYBOX, 'cat',
+                                                 detach=True, stdin_open=True)
+        id = container['Id']
```
aanand: Small thing: although we don't use Python's built-in `id` function anywhere in this block, it'd still be nicer not to shadow it for the sake of clarity (for one thing, it gets picked up by syntax highlighters, which results in misleading colouring). Maybe `container_id`?

tests/integration/exec_test.py
```diff
@@ -88,6 +94,67 @@ def test_exec_command_streaming(self):
...
+                if not next_data:
+                    return 0
+                data = data + next_data
+
+            if data is None:
+                return 0
+
+            if len(data) == 8:
+                _, actual = struct.unpack('>BxxxL', data)
+                return actual
```
aanand: Could these functions be moved to the top level? Even if they're only used by this test, they make it hard to read.

tests/integration/exec_test.py
```diff
@@ -88,6 +94,67 @@ def test_exec_command_streaming(self):
...
+            next_size = next_packet_size()
+            self.assertEqual(next_size, len(line))
+
+            data = six.binary_type()
+            while len(data) < next_size:
+                next_data = read(next_size - len(data))
+                if not next_data:
+                    assert False, "Failed trying to read in the dataz"
+                data += next_data
+            self.assertEqual(data.decode('utf-8'), "{0}".format(line))
```
aanand: What's the purpose of `"{0}".format(line)` here?

tests/integration/exec_test.py
```diff
@@ -88,6 +94,67 @@ def test_exec_command_streaming(self):
...
+        try:
+            next_size = next_packet_size()
+            self.assertEqual(next_size, len(line))
+
+            data = six.binary_type()
+            while len(data) < next_size:
+                next_data = read(next_size - len(data))
+                if not next_data:
+                    assert False, "Failed trying to read in the dataz"
+                data += next_data
```
aanand: Could this loop be extracted into a `read_data(next_size)` helper function?

tests/integration/exec_test.py
```diff
@@ -88,6 +94,67 @@ def test_exec_command_streaming(self):
...
+            self.assertEqual(data.decode('utf-8'), "{0}".format(line))
+        finally:
+            socket.close()
```
aanand: This `try/finally` could alternatively be done with a `self.addCleanup(socket.close)` line right after calling `exec_start`, which might be a bit more readable, as well as catching any errors raised by `select` (assuming it could raise any).

TomasTomecek:
@aanand thanks for the review. As I said previously, [this code](https://github.com/docker/docker-py/blob/d51d06d20971a04d7fad1a823808cec90a4d122b/tests/integration/container_test.py#L972) is copy&pasted from attach container integration test. So I find it pretty funny that you are commenting on a code which is already in.

I'm assuming that I should make one helper function and use it in both codebases, right?

TomasTomecek:
@aanand very good suggestions! code for tests looks a lot cleaner now, please take a look.

TomasTomecek:
fixed the flake; rebased

aanand:
Apologies, I hadn't looked at the codebase in a while so I didn't realise it was copied. Your suggestion was a good one and the tests indeed look much better now, so thanks!

I'd like to make one more suggestion, assuming it's possible - can the `select` call be moved to the top of the `read_socket` function?

TomasTomecek:
> Apologies, I hadn't looked at the codebase in a while so I didn't realise it was copied.

No worries.

> I'd like to make one more suggestion, assuming it's possible - can the select call be moved to the top of the read_socket function?

Done.

aanand:
LGTM

dnephin:
LGTM

TomasTomecek:
\o/
dockerpty now

PR #861, opened 2015-11-24
Update repo name resolution method

### LINKED ISSUE #860: Can't pull images with . In name.
Docker images that have a `.` in their name cannot be pulled with docker-py. This is a result of:

https://github.com/docker/docker-py/blob/master/docker/auth/auth.py#L46

shin-:
More relaxed version that matches the constraints imposed by the current version of the docker daemon.
Few unit tests to verify the new cases.
Client.pull was trying to set the tag value when it wasn't supposed to, fixed now.
`utils.parse_repository_tag` is simpler and supports `@sha...` notation.

Fixes #860

### DIFF
--- a/docker/utils/utils.py
+++ b/docker/utils/utils.py
@@ -283,16 +283,14 @@ def convert_volume_binds(binds):
     return result
 
 
-def parse_repository_tag(repo):
-    column_index = repo.rfind(':')
-    if column_index < 0:
-        return repo, None
-    tag = repo[column_index + 1:]
-    slash_index = tag.find('/')
-    if slash_index < 0:
-        return repo[:column_index], tag
-
-    return repo, None
+def parse_repository_tag(repo_name):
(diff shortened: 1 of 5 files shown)


docker/utils/utils.py
```diff
@@ -283,16 +283,14 @@ def convert_volume_binds(binds):
...
-
-    return repo, None
+def parse_repository_tag(repo_name):
+    parts = repo_name.rsplit('@', 1)
+    if len(parts) == 2:
+        return tuple(parts)
+    parts = repo_name.rsplit(':', 1)
+    if len(parts) == 2 and '/' not in parts[1]:
+        return tuple(parts)
+    return repo_name, None
```
dnephin: A test with an `@` digest would be good I think

shin-:
Updated with a couple tests for the `@digest` notation.

dnephin:
LGTM

PR #863, opened 2015-11-24
Don't descend into ignored directories when building context

### LINKED ISSUE #859: dockerignore implementation is relatively slow compared to Docker's implementation
I ran into an issue in a project where my builds - run through `docker-compose` - seemed to be taking an awfully long time (around ~60 seconds) during the context build/upload stage. `strace` showed a ton of time was being spent `stat()`ing files that were included in my `.dockerignore` rules, which I found curious.

Oddly, when I simply used `docker build` to build the container, I didn't have this issue, and context build/upload took about ~3-5 seconds. I couldn't figure out what was going wrong, so I investigated `docker-py`, and found that almost all of my execution time was spent in [this `get_paths` call](https://github.com/docker/docker-py/blob/master/docker/utils/utils.py#L110).

It appears that the difference in execution time is because docker-py's implementation of dockerignore/tar exclusion is far slower than Docker's:

Docker's implementation of the dockerignore exclusion algorithm, (seen [here](https://github.com/docker/docker/blob/master/pkg/archive/archive.go#L518)), walks through each folder, but [does not descend into a directory if it matched an exclusion pattern](https://github.com/docker/docker/blob/master/pkg/archive/archive.go#L556). Meanwhile, docker-py first gets an array of *every single file in the context folder*, and then applies a filter to the array. This seems to be what is causing the massive difference in execution time when I build my project - docker-py is iterating over thousands of files that Docker correctly ignores.

I started on a fix, using what I believe are the same rules as Docker's algorithm: https://github.com/thomasboyt/docker-py/commit/9f302f6721bb8492140cf5b218a80d62a2b62e19

This runs just as fast as Docker's implementation, but doesn't fully implement exception rules (e.g. `!foo`), leading it to fail a few tests. Before I go through and add this feature, I wanted to confirm that I'm on the right path (and that no one else has a better solution/algorithm to apply).

thomasboyt:
Closes #859.

The major limitation this fix has is that, if the `dockerignore` has any exception rules, the walk algorithm will descend into ignored directories, since of course if you have a `dockerignore` like:

```
foo/bar/baz/*
!*.txt
```

it needs to descend into `foo/bar/baz` to see if there are any files that match `*.txt` to include. This is the same behavior that Docker has, so I think it's acceptable - long-term, it may be possible to further optimize this by only descending if there is an exception rule that could actually match inside the directory.

### DIFF
--- a/docker/utils/utils.py
+++ b/docker/utils/utils.py
@@ -107,38 +107,68 @@ def exclude_paths(root, patterns, dockerfile=None):
 
     exclude_patterns = list(set(patterns) - set(exceptions))
 
-    all_paths = get_paths(root)
-
-    # Remove all paths that are matched by any exclusion pattern
-    paths = [
-        p for p in all_paths
-        if not any(match_path(p, pattern) for pattern in exclude_patterns)
-    ]
-
-    # Add back the set of paths that are matched by any inclusion pattern.
-    # Include parent dirs - if we add back 'foo/bar', add 'foo' as well
-    for p in all_paths:
-        if any(match_path(p, pattern) for pattern in include_patterns):
-            components = p.split('/')
-            paths += [
-                '/'.join(components[:end])
-                for end in range(1, len(components) + 1)
-            ]
+    paths = get_paths(root, exclude_patterns, include_patterns,
+                      has_exceptions=len(exceptions) > 0)
 
     return set(paths)
 
 
-def get_paths(root):
+def should_include(path, exclude_patterns, include_patterns):
+    """
+    Given a path, a list of exclude patterns, and a list of inclusion patterns:
+
+    1. Returns True if the path doesn't match any exclusion pattern
+    2. Returns False if the path matches an exclusion pattern and doesn't match
+       an inclusion pattern
+    3. Returns true if the path matches an exclusion pattern and matches an
+       inclusion pattern
+    """
+    for pattern in exclude_patterns:
+        if match_path(path, pattern):
+            for pattern in include_patterns:
+                if match_path(path, pattern):
+                    return True
+            return False
+    return True
+
+
+def get_paths(root, exclude_patterns, include_patterns, has_exceptions=False):
     paths = []
 
-    for parent, dirs, files in os.walk(root, followlinks=False):
+    for parent, dirs, files in os.walk(root, topdown=True, followlinks=False):
         parent = os.path.relpath(parent, root)
         if parent == '.':
             parent = ''
+
+        # If exception rules exist, we can't skip recursing into ignored
+        # directories, as we need to look for exceptions in them.
+        #
+        # It may be possible to optimize this further for exception patterns
+        # that *couldn't* match within ignored directores.
+        #
+        # This matches the current docker logic (as of 2015-11-24):
+        # https://github.com/docker/docker/blob/37ba67bf636b34dc5c0c0265d62a089d0492088f/pkg/archive/archive.go#L555-L557
+
+        if not has_exceptions:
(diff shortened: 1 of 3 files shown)


docker/utils/utils.py
```diff
@@ -107,38 +107,55 @@ def exclude_paths(root, patterns, dockerfile=None):
...
+
+        # Remove excluded patterns from the list of directories to traverse.
+        #
+        # If exception rules exist, we can't skip recursing into ignored
+        # directories, as we need to look for exceptions in them.
+        #
+        # It may be possible to optimize this further for exception patterns
+        # that *couldn't* match within ignored directores.
+        #
+        # This matches the current docker logic:
```
kanzure: probably better to use a timestamp here instead of "current docker logic"

docker/utils/utils.py
```diff
@@ -107,38 +107,55 @@ def exclude_paths(root, patterns, dockerfile=None):
...
+        # https://github.com/docker/docker/blob/37ba67bf636b34dc5c0c0265d62a089d0492088f/pkg/archive/archive.go#L555-L557
+
+        if not has_exceptions:
+            dirs[:] = [d for d in dirs if
```
kanzure: live mutation of the `os.walk` output variable `dirs` is probably non-obvious and deserves a comment

docker/utils/utils.py
```diff
@@ -107,38 +107,55 @@ def exclude_paths(root, patterns, dockerfile=None):
...
 
-def get_paths(root):
+def should_include(path, exclude_patterns, include_patterns):
+    for pattern in exclude_patterns:
+        if match_path(path, pattern):
+            for pattern in include_patterns:
+                if match_path(path, pattern):
+                    return True
+            return False
+    return True
```
aanand: I feel like this could use a docstring.

aanand:
Nice. I agree with @kanzure's comments, plus it'd be good to document the logic of `should_include`.

thomasboyt:
@aanand cool, added some docs/comments!

aanand:
Thanks! I think this can be squashed to one commit.

thomasboyt:
@aanand rebased :)

aanand:
Thanks. I just noticed that we're now excluding parent directories of files mentioned in !-style exceptions. Is that in line with Docker client behaviour?

thomasboyt:
So, while before it was adding the parent *directories* of exceptions to the set of included paths, it wasn't adding back the *contents* of those directories. 

Let's say you have the following structure:

```
foo/
  a.py
  b.py
  bar/
    a.py
```

and the following rules:

```
foo
!foo/bar/a.py
```

Before, the final set of included paths would have looked like:

```
foo
foo/bar
foo/bar/a.py
```

Because the tar created in the build process [adds each path non-recursively](https://github.com/thomasboyt/docker-py/blob/fast-exclude-paths/docker/utils/utils.py#L85), this didn't actually include `foo/a.py` or `foo/b.py`, it just made empty directories for `foo/` and `foo/bar/`.

Now, the final set of included paths just looks like:

```
foo/bar/a.py
```

Even though `foo` and `foo/bar` are no longer included in the set of paths, they'll still be created when `foo/bar/a.py` is added to the build context (which the updated integration test confirms, since I wanted to make sure it was creating the directories).

aanand:
OK, so even if we're not passing the directories explicitly to `tar`, they'll still be created as needed? Sounds good. Glad the integration test now explicitly checks the "exception rule for file in ignored subdirectory" case.

LGTM

dnephin:
LGTM

PR #865, opened 2015-11-30
Allow utils.tar to take an existing file object.

dlorenc:
Signed-off-by: dlorenc <lorenc.d@gmail.com>

### DIFF
--- a/docker/utils/utils.py
+++ b/docker/utils/utils.py
@@ -74,9 +74,10 @@ def decode_json_header(header):
     return json.loads(data)
 
 
-def tar(path, exclude=None, dockerfile=None):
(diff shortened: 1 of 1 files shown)


dlorenc:
Hey,

We're using something like this to package up a docker context and stream it to another machine ourselves before running a build, and to do a few in-memory operations on the tar instead of writing it to disk. Getting this upstreamed will let us drop our "fork" of the tar function.

It's not a huge amount of code, but it would be nice to make sure we're using the exact same tar logic as docker-py over time.

aanand:
LGTM

shin-:
Seems fine. LGTM!

PR #866, opened 2015-12-01
fix typo filter param for `volumes`

ticosax:
https://docs.docker.com/engine/reference/api/docker_remote_api_v1.21/#list-volumes

Signed-off-by: Nicolas Delaby <nicolas.delaby@lock8.me>


dnephin:
LGTM

PR #867, opened 2015-12-03
Allow providing options when creating networks

sumitsahrawat:
Following the spec:

    http://docs.docker.com/engine/reference/api/docker_remote_api_v1.21/#create-a-network

I have added an options argument to create_network.

This opens up the possibility of creating isolated containers with no internet access programmatically. We require such facilities in https://github.com/jupyter/tmpnb/issues/187.

### DIFF
--- a/docker/api/network.py
+++ b/docker/api/network.py
@@ -19,10 +19,14 @@ class NetworkApiMixin(object):
         return self._result(res, json=True)
 
     @minimum_version('1.21')
-    def create_network(self, name, driver=None):
+    def create_network(self, name, driver=None, options=None):
+        if options is not None and not isinstance(options, dict):
+            raise TypeError('options must be a dictionary')
+
         data = {
             'name': name,
             'driver': driver,
+            'options': options
         }
         url = self._url("/networks/create")
         res = self._post_json(url, data=data)
--- a/docker/api/volume.py
+++ b/docker/api/volume.py
@@ -33,4 +33,3 @@ class VolumeApiMixin(object):
         url = self._url('/volumes/{0}', name)
         resp = self._delete(url)
         self._raise_for_status(resp)
-        return True
--- a/docs/api.md
+++ b/docs/api.md
@@ -780,8 +780,7 @@ Remove a volume. Similar to the `docker volume rm` command.
 
 * name (str): The volume's name
 
-**Returns** (bool): True on successful removal. Failure will raise a
-`docker.errors.APIError` exception.
+Failure to remove will raise a `docker.errors.APIError` exception.
 
 ## rename
 
--- a/docs/api.md
+++ b/docs/api.md
@@ -163,6 +163,15 @@ non-running ones
   'Status': 'Up 1 seconds'}]
 ```
 
+## connect_container_to_network
+
+Connect a container to a network.
+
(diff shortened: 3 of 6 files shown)


dnephin:
I think we should add the `options` field. This PR requires changes to the docs and tests.

There is a much easier way to disable the network for a container by create it with `net="none"`

sumitsahrawat:
> This PR requires changes to the docs and tests.

Yup, I thought so too. I'll take a look.

> There is a much easier way to disable the network for a container by create it with net="none"

We need the container to have networking, but prevent them from having generic internet access or inter-container access for tmpnb.

Using docker from the shell, we want to isolate containers as shown in https://github.com/docker/libnetwork/issues/735#issuecomment-157141350

sumitsahrawat:
Looking at https://docker-py.readthedocs.org, I found out that `Client.networks`, `Client.create_network` and other similar commands are present only in the changelog, not in the documentation.

dnephin:
LGTM

sumitsahrawat:
I'm not that experienced with Python and testing in Python. Do the tests
need any changes?

docker/api/network.py
```diff
@@ -33,6 +37,7 @@ def remove_network(self, net_id):
         url = self._url("/networks/{0}", net_id)
         res = self._delete(url)
         self._raise_for_status(res)
+        return True
```
aanand: What's the purpose of this line?

aanand:
Thanks! I think `test_create_network` should be updated, specifically [this line](https://github.com/docker/docker-py/blob/93a296f/tests/unit/network_test.py#L62) (to pass options) and [this line](https://github.com/docker/docker-py/blob/93a296f/tests/unit/network_test.py#L77) (to check that they made it through to the HTTP request).

docker/api/network.py
sumitsahrawat: I wanted to make the `remove_network` function similar to the `remove_volume` function. Now, both return `True` if no exception is caused.

https://github.com/docker/docker-py/blob/master/docker/api/volume.py#L32-L36

sumitsahrawat:
I'll take a look. Thanks for the guidance @aanand :+1:

docker/api/network.py
aanand: OK, I don't think either method should return anything, since the return value is always the same and therefore meaningless. @shin- - what was the reason for `remove_volume` returning `True`?

rgbkrk:
Thank you all! I look forward to using this.

docker/api/network.py
shin-: No real reason that I can remember. I may have been on the fence at the time about returning `False` if the operation doesn't complete in non-exceptional contexts, but I don't think we have any of those, so it's a moot point.

sumitsahrawat:
All done!

aanand:
Great. I think this can be squashed to two commits: one to get rid of `remove_volume`'s return value, and one for the new argument to `create_network`.

sumitsahrawat:
@aanand Squashed to three. I reordered the commits so that the `remove_volume` fix appears before docs are added, but GitHub seems to think otherwise.

aanand:
Thanks!

LGTM

PR #874, opened 2015-12-10
Treat 'index.docker.io' repo names as 'docker.io'

aanand:
Fixes the problem described in https://github.com/docker/compose/issues/1622#issuecomment-162988389. Auth lookups were failing for private images specified as `docker.io/username/image`.

I'm still a bit unsure as to exactly how we're supposed to treat each type of key we encounter in an auth file (e.g. whether we're supposed to strip all URLs of scheme and path, or just special-case the Docker index). I had a go at reading the Docker client code but it's hard to follow. So I'm not 100% sure we're doing the right thing here.

### DIFF
--- a/docker/auth/auth.py
+++ b/docker/auth/auth.py
@@ -21,7 +21,7 @@ import six
 
 from .. import errors
 
-INDEX_NAME = 'index.docker.io'
+INDEX_NAME = 'docker.io'
 INDEX_URL = 'https://{0}/v1/'.format(INDEX_NAME)
 DOCKER_CONFIG_FILENAME = os.path.join('.docker', 'config.json')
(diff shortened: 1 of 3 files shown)


dnephin:
Seems reasonable to me. 

Previously we weren't converting `index.docker.io` to `docker.io` which caused the "find auth config for hostname" to fail for a private repo on the hub.  After this change we normalize to `docker.io`.  Is that right?

shin-:
I'll try to check against the engine code base, I was in that part not too long ago for #861

aanand:
@dnephin correct.

shin-:
Ok, I checked against the engine code (which I agree is a mess) and this looks to imitate the behavior properly.

LGTM.

PR #886, opened 2015-12-27
HttpHeaders support (similar to docker client)

kpavel:
Signed-off-by: Pavel Kravchenko <kpavel@il.ibm.com>

Docker client already supports HttpHeaders from config file.
https://docs.docker.com/v1.7/docker/reference/commandline/cli/
Leveraging existing configuration file support in python docker client added support for generic HttpHeaders.

### DIFF
--- a/docker/auth/auth.py
+++ b/docker/auth/auth.py
@@ -158,7 +158,6 @@ def load_config(config_path=None):
         explicit config_path parameter > DOCKER_CONFIG environment variable >
         ~/.docker/config.json > ~/.dockercfg
     """
-
     config_file = find_config_file(config_path)
 
     if not config_file:
@@ -168,11 +167,17 @@ def load_config(config_path=None):
     try:
         with open(config_file) as f:
             data = json.load(f)
+            res = {}
             if data.get('auths'):
                 log.debug("Found 'auths' section")
-                return parse_auth(data['auths'])
+                res.update(parse_auth(data['auths']))
+            if data.get('HttpHeaders'):
+                log.debug("Found 'HttpHeaders' section")
+                res.update({'HttpHeaders': data['HttpHeaders']})
+            if res:
+                return res
             else:
-                log.debug("Couldn't find 'auths' section")
+                log.debug("Couldn't find 'auths' or 'HttpHeaders' sections")
                 f.seek(0)
                 return parse_auth(json.load(f))
     except (IOError, KeyError, ValueError) as e:
--- a/docker/client.py
+++ b/docker/client.py
@@ -28,7 +28,7 @@ from . import errors
 from .auth import auth
 from .unixconn import unixconn
 from .ssladapter import ssladapter
-from .utils import utils, check_resource
+from .utils import utils, check_resource, update_headers
 from .tls import TLSConfig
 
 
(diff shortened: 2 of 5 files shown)


docker/auth/auth.py
```diff
@@ -149,12 +149,19 @@ def load_config(config_path=None):
...
-                    return parse_auth(data)
-            log.debug("Couldn't find 'auths' section")
+                    if section == 'auths':
+                        log.debug("Found 'auths' section")
+                        res.update(parse_auth(data))
+                    elif section == 'HttpHeaders':
+                        log.debug("Found 'HttpHeaders' section!: " + str(section) + ':' + str(data))
+                        res.update({section: data})
+                if res:
+                    log.debug('Returning res: ' + str(res))
```
shin-: This line seems unnecessary.

docker/auth/auth.py
```diff
@@ -149,12 +149,19 @@ def load_config(config_path=None):
...
-                    log.debug("Found 'auths' section")
-                    return parse_auth(data)
-            log.debug("Couldn't find 'auths' section")
+                    if section == 'auths':
+                        log.debug("Found 'auths' section")
+                        res.update(parse_auth(data))
+                    elif section == 'HttpHeaders':
+                        log.debug("Found 'HttpHeaders' section!: " + str(section) + ':' + str(data))
+                        res.update({section: data})
+                if res:
```
shin-: `return res` should happen regardless of its value.

docker/client.py
```diff
@@ -84,6 +83,19 @@ def __init__(self, base_url=None, version=None,
                 )
             )
 
+    def update_headers(func):
```
shin-: this should be moved to `decorators.py`

docker/client.py
```diff
@@ -84,6 +83,19 @@ def __init__(self, base_url=None, version=None,
...
+        def inner(self, *args, **kwargs):
+            if 'HttpHeaders' in self._auth_configs:
+                if 'headers' not in kwargs:
+                    headers = self._auth_configs['HttpHeaders']
```
shin-: Doesn't look like `headers` is being used anywhere.

shin-:
Thanks for your contribution! 

I made a few comments, and the branch also needs to be rebased.

kpavel:
Thanks for your review!
Fixed and rebased.

shin-:
LGTM.

docker/auth/auth.py
```diff
@@ -168,11 +167,17 @@ def load_config(config_path=None):
...
+            res = {}
             if data.get('auths'):
                 log.debug("Found 'auths' section")
-                return parse_auth(data['auths'])
+                res.update(parse_auth(data['auths']))
+            if data.get('HttpHeaders'):
+                log.debug("Found 'HttpHeaders' section")
+                res.update({'HttpHeaders': data['HttpHeaders']})
+            if res:
+                return res
```
dnephin: I think this could use a new test case. I don't see anything asserting these are read properly.

tombee:
Is this PR just missing a test to check HttpHeaders are read from the `~/.docker/config.json` file?

@kpavel Are you able to add this test?  Would be great to get this PR merged.

tombee:
Awesome work @kpavel! :+1:

shin-:
Closed via #899 - Thanks!

Issue #873, opened 2015-12-07
docker-py 1.5.0 fails to ping docker 1.8.3


migueleliasweb:
Right after I update the docker-py to 1.5.0, all the ping calls started to fail =(

I don't know if this is the desired behaviour but if it is, please disconsider this issue.

The result of the .ping is:
```
APIError: 400 Client Error: Bad Request ("client is newer than server (client API version: 1.21, server API version: 1.20)")
```

```
$pip freeze | grep docker-py && python --version && docker version
docker-py==1.5.0
Python 2.7.6
Client:
 Version:      1.8.3
 API version:  1.20
 Go version:   go1.4.2
 Git commit:   f4bf5c7
 Built:        Mon Oct 12 05:37:18 UTC 2015
 OS/Arch:      linux/amd64

Server:
 Version:      1.8.3
 API version:  1.20
 Go version:   go1.4.2
 Git commit:   f4bf5c7
 Built:        Mon Oct 12 05:37:18 UTC 2015
 OS/Arch:      linux/amd64
```

migueleliasweb:
Don't know if this helps but the ping method was moved from client.py:602 to daemon.py:73 and the internal method _url changed. But I couldn't find out why this fails...

dnephin:
You'll have to set the api version if you aren't using the latest server. You can use either `docker.Client(version='1.20')` or `docker.Client(version='auto')`

migueleliasweb:
Sir, you saved my day. Thank you a lot.

I didnt knew about the version parameter when using the Client... sry

TomasTomecek:
I feel like that `auto` should be the default since this comes up so often.

aanand:
@TomasTomecek That would involve making an extra call to the Docker API on every `docker-compose` command, which is probably unacceptable.

shin-:
docker-compose would just have to not use the default, so I don't think it's too much of an issue in that respect (given we handle the version bump properly etc).

Fixed default is just much easier to debug than `'auto'` in my opinion though.

TomasTomecek:
Fair points.

How about improving the exception message then? Something like:
```
You have requested daemon API version 1.21 but it supports only 1.20.
Either supply exact version to Client `Client(version="1.20")` or let
client pick it automatically: `Client(version="auto")`
```

migueleliasweb:
@TomasTomecek That's a great idea. This exception message would help many.

+1

prabhakar-pal:
[...]

PR #887, opened 2015-12-27
provide detailed information on recovering from version mismatch

TomasTomecek:
Related: #873

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -18,6 +18,34 @@ is hosted.
 * tls (bool or [TLSConfig](tls.md#TLSConfig)): Equivalent CLI options: `docker --tls ...`
 * user_agent (str): Set a custom user agent for requests to the server.
 
+
+## Version mismatch
+
+You may encounter an error like this:
+
+```text
+client is newer than server (client API version: 1.24, server API version: 1.23)
+```
+
+To fix this, you have to either supply exact version to `Client` which you know that server supports:
+
+```python
+client = docker.Client(version="1.23")
+```
+
+or let client pick the newest version server supports:
+
+```python
+client = docker.Client(version="auto")
+```
+
+or even
+
+```python
+client = docker.AutoVersionClient()
+```
+
+
 ****
 
 ## attach
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -7,6 +7,7 @@ theme: readthedocs
 pages:
 - Home: index.md
 - Client API: api.md
+- FAQ: faq.md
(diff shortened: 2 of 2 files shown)


TomasTomecek:
rebased

TomasTomecek:
rebased; ping @dnephin @shin-

docker/client.py
```diff
@@ -151,6 +151,18 @@ def _raise_for_status(self, response, explanation=None):
         except requests.exceptions.HTTPError as e:
             if e.response.status_code == 404:
                 raise errors.NotFound(e, response, explanation=explanation)
+            if e.response.status_code == 400:
+                if response.content.startswith("client is newer than server"):
+                    explanation = (
+                        'docker-py is using newer API version than docker'
```
dnephin: It seems a bit strange to me that a library would report itself in the error message.

Usually a library is consumed by some other application, which probably doesn't want to expose these details in error messages.

I don't really know how I feel about this change. I think a better solution might be to just include the error string in the docs, that way if someone hits the issue and they search for it, they'll find the docs about how to set the version.

dnephin:
I think something like this is probably a good idea. I'm not sure if "Pitfalls" is the right term. Maybe "Gotchas" or just FAQ?

TomasTomecek:
since I'm not native speaker, I usually have hard times picking the right term; I really like FAQ, will change it to that one, thanks for suggestion

TomasTomecek:
@dnephin updated

dnephin:
LGTM

TomasTomecek:
CI failure seems unrelated:

```
docker: Error response from daemon: Cannot link to a non running container: /dpy-dind-597 AS /dpy-dind2-597/docker.
docker: Error response from daemon: Cannot link to a non running container: /dpy-dind-597 AS /dpy-dind3-597/docker.
```

TomasTomecek:
rebased

TomasTomecek:
rebased

TomasTomecek:
rebased

@shin- can we merge this, it already has one lgtm from @dnephin

bfirsh:
Shouldn't this be in the part of the documentation about instantiating clients? Creating an FAQ is always a warning sign for me that the main documentation isn't good enough.

TomasTomecek:
updated, @bfirsh PTAL

TomasTomecek:
@shin- @bfirsh rebased

TomasTomecek:
rebased

TomasTomecek:
\o/ thank you!

PR #888, opened 2015-12-29
Update the test_inspect_network

wenchma:
This  is depended by https://github.com/docker/docker/pull/18928

Related to docker/docker#18626

Signed-off-by: Wen Cheng Ma <wenchma@cn.ibm.com>

### DIFF
--- a/tests/integration/network_test.py
+++ b/tests/integration/network_test.py
@@ -35,15 +35,11 @@ class TestNetworks(helpers.BaseTestCase):
         net_name, net_id = self.create_network()
 
         net = self.client.inspect_network(net_id)
-        self.assertEqual(net, {
-            u'Name': net_name,
-            u'Id': net_id,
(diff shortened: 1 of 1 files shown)


wenchma:
This PR and docker/docker#18928 depend on each other, so it will block each other. I workaround it that firstly we remove `test_inspect_network` temporarily, once docker/docker#18928 is merged, I can revert the remove and update the test.  wdyt ? @shin-

shin-:
Maybe change the test instead so that it tests a few specific values in the result - I would check for the presence (and validity) of `Name`, `Id` and `Driver` in particular. That would prevent the current deadlock as well.

wenchma:
@shin- good idea, :+1:   already updated the PR to cover only a few specific values.

shin-:
Thanks - LGTM!

PR #889, opened 2016-01-05
Improve host devices support

### LINKED ISSUE #725: ER: Support "device" for mapping host devices to containers
We have some special host device files we need to access from inside the containers, it doesn't look as if the docker-py library permits passing --device?

shin-:
- Add unit tests for `utils.parse_devices`
- Rewrite documentation
- Support dict and string format for device declaration

Fixes #725

### DIFF
--- a/docs/host-devices.md
+++ b/docs/host-devices.md
@@ -12,7 +12,18 @@ cli.create_container(
 )
 ```
 
-Each string is a single mapping using the colon (':') as the separator. So the
-above example essentially allow the container to have read write permissions to
-access the host's /dev/sda via a node named /dev/xvda in the container. The
-devices parameter is a list to allow multiple devices to be mapped.
+Each string is a single mapping using the following format:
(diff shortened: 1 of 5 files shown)


docs/host-devices.md
```diff
@@ -12,7 +12,7 @@ cli.create_container(
...
 ```
 
-Each string is a single mapping using the colon (':') as the separator. So the
-above example essentially allow the container to have read write permissions to
-access the host's /dev/sda via a node named /dev/xvda in the container. The
-devices parameter is a list to allow multiple devices to be mapped.
+Each string is a single mapping using the following format:
+`<path_on_host>:<path_in_container>:<cgroup_permissions>`
+The above example allows the container to have read-write access to
+the host's `/dev/sda` via a node named `/dev/xvda` inside the container.
```
aanand: Should we also document the `dict` form of elements in the `devices` list?

shin-: Added :+1:

aanand:
LGTM

PR #892, opened 2016-01-09
Don't break if base_url == None and TLS is enabled

shin-:


toli:
LGTM
thanks for fixing it

dnephin:
LGTM

PR #895, opened 2016-01-14
Implement support for network-scoped aliases

aanand:
Missing docs, but network docs are already lacking. Will fill them out in a follow-up PR.

### DIFF
--- a/docker/api/container.py
+++ b/docker/api/container.py
@@ -4,6 +4,7 @@ from datetime import datetime
 
 from .. import errors
 from .. import utils
+from ..utils.utils import create_networking_config, create_endpoint_config
 
 
 class ContainerApiMixin(object):
@@ -98,7 +99,7 @@ class ContainerApiMixin(object):
                          cpu_shares=None, working_dir=None, domainname=None,
                          memswap_limit=None, cpuset=None, host_config=None,
                          mac_address=None, labels=None, volume_driver=None,
-                         stop_signal=None):
+                         stop_signal=None, networking_config=None):
 
         if isinstance(volumes, six.string_types):
             volumes = [volumes, ]
@@ -113,7 +114,7 @@ class ContainerApiMixin(object):
             tty, mem_limit, ports, environment, dns, volumes, volumes_from,
             network_disabled, entrypoint, cpu_shares, working_dir, domainname,
             memswap_limit, cpuset, host_config, mac_address, labels,
-            volume_driver, stop_signal
+            volume_driver, stop_signal, networking_config,
         )
         return self.create_container_from_config(config, name)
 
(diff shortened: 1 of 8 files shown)


docker/utils/utils.py
```diff
@@ -715,6 +715,27 @@ def create_host_config(binds=None, port_bindings=None, lxc_conf=None,
...
+    networking_config = {}
+
+    if endpoints_config:
+        networking_config["EndpointsConfig"] = endpoints_config
+
+    return networking_config
+
+
+def create_endpoint_config(version,
+                           aliases=None):
```
dnephin: could be a single line

docker/api/container.py
```diff
@@ -139,6 +139,12 @@ def create_host_config(self, *args, **kwargs):
         kwargs['version'] = self._version
         return utils.create_host_config(*args, **kwargs)
 
+    def create_networking_config(self, *args, **kwargs):
+        return utils.create_networking_config(*args, **kwargs)
+
+    def create_endpoint_config(self, *args, **kwargs):
+        return utils.create_endpoint_config(self._version, *args, **kwargs)
```
dnephin: These methods don't seem to be very much, do we need them?

docker/utils/utils.py
```diff
@@ -715,6 +715,27 @@ def create_host_config(binds=None, port_bindings=None, lxc_conf=None,
...
+    endpoint_config = {}
+
+    if aliases:
+        if version_lt(version, '1.22'):
+            raise host_config_version_error('aliases', '1.22')
```
dnephin: Should this be `'networking_config.aliases'` since the error is about a key relative to the host_config?

dnephin:
Couple small nits, otherwise LGTM

docker/utils/utils.py
aanand: Yep, just anticipating future additions (these things have a tendency to grow), but it's probably unnecessary

docker/api/container.py
aanand: `create_endpoint_config` needs to pass the version through, same as `create_host_config`. `create_networking_config` doesn't need to *right now*, but it might in the future, and I'd like to avoid having to do another deprecation like we did with `create_host_config`.

docker/utils/utils.py
aanand: Yeah. Perhaps just `endpoint_config.aliases`, since the user will have explicitly called `create_endpoint_config`. I'll update it.

docker/api/container.py
dnephin: Ok

tests/integration/network_test.py
```diff
@@ -7,7 +7,6 @@
 from ..base import requires_api_version
 
 
-@requires_api_version('1.21')
```
shin-: How about leaving this here, and only adding `requires('1.22')` to the few methods where it's necessary?

aanand: I wasn't sure if the two would interact weirdly, so I elected to be more explicit, but I don't have strong feelings about it

shin-:
LGTM

PR #900, opened 2016-01-19
Propagate headers info back from streamed responses

frol:
By streamed responses I mean `stream=True` in `.attach()`, `.attach_socket()`, and `.exec_start()`.

This PR is useful for those who need to distinguish stdout from stderr.


frol:
@shin- Please, review the PR.

frol:
@shin- @aanand Do you think it is worth to rebase this PR? I have been using it over a year now in production with no issues at all!

frol:
Given `demux` option is now implemented, there is no need in this PR anymore.

PR #902, opened 2016-01-20
Drop support for python 3.2

shin-:
Following in pip's tracks, dropping support for python 3.2: https://pip.pypa.io/en/stable/news/#release-notes


shin-:
It's likely that docker-py will still work with python 3.2 for the foreseeable future, but we won't enforce it or test for it anymore.

TomasTomecek:
Would be nice to add 3.5 support :)

aanand:
LGTM

PR #905, opened 2016-01-25
remove tempfile.mkdtemp in VolumeBindTest for non-auto-creating non-e…

keloyang:
remove tempfile.mkdtemp(change to '/tmp') in VolumeBindTest,if docker don't create the host path automatically,we must keep the host path exist.
https://github.com/docker/docker/pull/19537 for detail

Signed-off-by: yangshukui <yangshukui@huawei.com>


shin-:
Hard-coding the `/tmp` value is 100% the wrong approach.

keloyang:
why can't reopen?

PR #911, opened 2016-01-29
Support unspecified protocol in base_url when using TLS

### LINKED ISSUE #894: Problem when using the DOCKER_HOST variable in combination with docker-compose and https://
Hi,

I'm trying to use docker & docker-compose with the DOCKER_HOST env-variable to control a remote docker-host.

at first I configured the variables on the docker client machine as follows:

    export DOCKER_CERT_PATH=/vagrant/docker
    export DOCKER_TLS_VERIFY=1
    export DOCKER_HOST=my-docker-vm.cloudapp.azure.com:2376

This lead to an error in docker-compose as discussed in this issue docker/compose#2634

Once I added the `https://` protocol prefix to the hostname the connection problem with docker-compose went away...

    export DOCKER_HOST=https://my-docker-vm.cloudapp.azure.com:2376

But with this format of the DOCKER_HOST variable now the docker CLI is complaining about the format...

    docker ps
    Invalid bind address format: https://my-docker-vm.cloudapp.azure.com:2376

I think this error is triggered here: https://github.com/docker/docker-py/blob/62d9964cc1881f6f3cd021594cd40fd80a8fc855/docker/utils/utils.py#L388

The code is not expecting to find two occurrences of double-dots `:` in the host variable, but for some reason it does.

PS: also the ambiguity of the two error messages in `utils.py` could be improved, since at L368 & at L388 both errors emit the same message, but for different reasons ;)

Thanks & Regards,
Wolfgang

shin-:
Fixes #894

### DIFF
--- a/docker/client.py
+++ b/docker/client.py
@@ -45,17 +45,17 @@ class Client(
                  timeout=constants.DEFAULT_TIMEOUT_SECONDS, tls=False):
         super(Client, self).__init__()
 
-        if tls and (not base_url or not base_url.startswith('https://')):
+        if tls and not base_url:
             raise errors.TLSParameterError(
-                'If using TLS, the base_url argument must begin with '
-                '"https://".')
+                'If using TLS, the base_url argument must be provided.'
+            )
 
         self.base_url = base_url
         self.timeout = timeout
 
         self._auth_configs = auth.load_config()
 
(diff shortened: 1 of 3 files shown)


dnephin:
LGTM

PR #912, opened 2016-01-29
Advanced documentation for port bindings

### LINKED ISSUE #495: Not able to bind multiple IPs to the same port?
I am able to bind several IPs to the same port from the docker run command line, but it doesn't seem possible to do with the docker-py API?

For example this is possible from the command line:

```
docker run -p 50.23.229.126:25:25 -p 192.168.10.20:25:25 mail
```

Reading https://docker-py.readthedocs.org/en/latest/port-bindings/ and trying a few things it does not seem possible with docker-py?

shin-:
Fixes #495

### DIFF
--- a/docs/port-bindings.md
+++ b/docs/port-bindings.md
@@ -37,3 +37,20 @@ container_id = cli.create_container(
     })
 )
 ```
+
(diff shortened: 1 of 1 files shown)


dnephin:
LGTM

PR #915, opened 2016-01-30
Update outdated error message

### LINKED ISSUE #914: Probably invalid version checking in container creation
Hi.
I've stumbled upon something i think is a not up-to-date code.

utils.py at line 794 in my distribution (1.6.0) states following:

    if compare_version('1.10', version) >= 0:
        message = ('{0!r} parameter has no effect on create_container().'
                   ' It has been moved to start()')
        if dns is not None:
            raise errors.InvalidVersion(message.format('dns'))
        if volumes_from is not None:
            raise errors.InvalidVersion(message.format('volumes_from'))

As far as i can read from docs, this is not true - both Dns and VolumesFrom are seen in HostConfig for API 1.21. Even if everything is okay with this code, there is still a problem i'm struggling to solve - some of the HostConfig options can be specified in `.container_create()`, but not on `.start()`, e.g. logging driver and options may be set only during creation, but if i need to use custom DNS, i have to specify it on `.start()` call, and this will override whole HostConfig, erasing my logging configuration. Since this code directly raise exceptions, i have no option but to create whole dict directly and pass it to `container_from_configuration()`.

I'm sorry i'm not doing direct pull request instead of issue, but i don't have enough time to figure out correct version constraints and get into projects conventions - sadly, i'm no pythonist and can't do it.

shin-:
Fixes #914

### DIFF
--- a/docker/utils/utils.py
+++ b/docker/utils/utils.py
@@ -889,7 +889,7 @@ def create_container_config(
 
     if compare_version('1.10', version) >= 0:
         message = ('{0!r} parameter has no effect on create_container().'
-                   ' It has been moved to start()')
+                   ' It has been moved to host_config')
         if dns is not None:
             raise errors.InvalidVersion(message.format('dns'))
         if volumes_from is not None:


dnephin:
LGTM

Issue #816, opened 2015-10-21
use correct ssl version with urllib3 and pyopenssl


stefanfoulis:
There are situations where ``get_max_tls_protocol()`` will get a max openssl version from python ``ssl`` that is higher than what ``urllib3`` supports.

The error will be:

```
KeyError: 5
```
in ``requests/packages/urllib3/contrib/pyopenssl.py:271`` on the line with ``ctx = OpenSSL.SSL.Context(_openssl_versions[ssl_version])``

Using the following replacement for ``docker.ssladapter.ssladapter.get_max_tls_protocol()`` worked for me, but there might be other cases I am missing:

```python
from requests.packages.urllib3.contrib import pyopenssl
def get_max_tls_protocol():
    protocols = ('PROTOCOL_TLSv1_2',
                 'PROTOCOL_TLSv1_1',
                 'PROTOCOL_TLSv1',
    for proto in protocols:
        if hasattr(ssl, proto):
            proto_id = getattr(ssl, proto)
            if proto_id in pyopenssl._openssl_versions:
                return proto_id
```

posita:
FYI, this may result in `docker-py` raising a very confusing error:

```py
...
  File "/.../lib/python2.7/site-packages/docker/client.py", line 325, in __init__
    super(AutoVersionClient, self).__init__(*args, **kwargs)
  File "/.../lib/python2.7/site-packages/docker/client.py", line 77, in __init__
    self._version = self._retrieve_server_version()
  File "/.../lib/python2.7/site-packages/docker/client.py", line 92, in _retrieve_server_version
    'Invalid response from docker daemon: key "ApiVersion"'
docker.errors.DockerException: Invalid response from docker daemon: key "ApiVersion" is missing.
```

This is because [`_retrieve_server_version`](https://github.com/docker/docker-py/blob/1.5.0/docker/client.py#L87) is confusing the source of the `KeyError`:

```py
    ...
    def _retrieve_server_version(self):
        try:
            # self.version raises a KeyError from pyopenssl here ...
            return self.version(api_version=False)["ApiVersion"]
        except KeyError:
            # ... which is then misinterpreted as having come from
            # the attempt to access the "ApiVersion" key here
            raise errors.DockerException(
                'Invalid response from docker daemon: key "ApiVersion"'
                ' is missing.'
            )
        ...
```

It should probably be rewritten. Here's one suggestion:

```py
    def _retrieve_server_version(self):
        try:
            res = self.version(api_version=False)
[...]

PR #920, opened 2016-02-04
Remove obsolete SSL version computation

shin-:
Recent versions of urllib3 (including the one packaged by requests) will automatically reject SSLv2/3, and the default constant `ssl.PROTOCOL_SSLv23` will select the [highest available protocol](https://docs.python.org/2/library/ssl.html#ssl.PROTOCOL_SSLv23), including TLS.

### DIFF
--- a/docker/ssladapter/ssladapter.py
+++ b/docker/ssladapter/ssladapter.py
@@ -4,7 +4,6 @@
 """
 from distutils.version import StrictVersion
 from requests.adapters import HTTPAdapter
-import ssl
 
 try:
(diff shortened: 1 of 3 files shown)


shin-:
Should fix #816

zbyte64:
Can confirm this fixes #816 :)

`pip install git+https://github.com/docker/docker-py.git@ssl_version_simpler#docker-py`

shin-:
Cool, thank you for testing! We'll try to have it in a 1.7.1 soonish I think.

dnephin:
LGTM

PR #921, opened 2016-02-04
create tempfile in VolumeBindTest for non-auto-creating of docker

keloyang:
create tempfile in VolumeBindTestin ,if docker don't create the host path automatically,we must keep the host path exist.
https://github.com/docker/docker/pull/19953 for detail.

Signed-off-by: yangshukui <yangshukui@huawei.com>


shin-:
Thanks!

PR #924, opened 2016-02-04
Re-write kwargs_from_env to handle TLS options better

mikedougherty:
This more closely matches the way the docker client interprets the
relevant environment variables. Among other things, it's now possible to
set the equivalent of `--tlsverify=false` with `DOCKER_TLS_VERIFY=`. 

It appears this was previously attempted with the "assert_hostname" argument (see 4bf438d7b8d47c04aec5c1e33fa358417b37df55) but that doesn't capture all cases of TLS verification nor is it specifiable from a user's environment.

This allows docker-compose to communicate with daemons that have unverifiable TLS configurations.

Previously:
```
$ export DOCKER_HOST=tcp://10.7.29.178:2376
$ export DOCKER_TLS_VERIFY=
$ python
>>> import docker
>>> try: len(docker.Client(**docker.utils.kwargs_from_env()).containers())
... except Exception as e: print e
('Connection aborted.', BadStatusLine('\x15\x03\x01\x00\x02\x02\n',))

$ ~/src/compose/bin/docker-compose ps
ERROR: Couldn't connect to Docker daemon - you might need to run `docker-machine start default`.
```

With these changes:
```
$ export DOCKER_HOST=tcp://10.7.29.178:2376
$ export DOCKER_TLS_VERIFY=
$ export PYTHONPATH="$HOME/src/docker-py:$PYTHONPATH"
$ python
>>> import docker
>>> len(docker.Client(**docker.utils.kwargs_from_env()).containers())
2

$ ~/src/compose/bin/docker-compose ps
Name   Command   State   Ports
------------------------------
```

### DIFF
--- a/docker/utils/utils.py
+++ b/docker/utils/utils.py
@@ -445,26 +445,45 @@ def parse_devices(devices):
 
 def kwargs_from_env(ssl_version=None, assert_hostname=None):
     host = os.environ.get('DOCKER_HOST')
-    cert_path = os.environ.get('DOCKER_CERT_PATH')
+
+    # empty string for cert path is the same as unset.
+    cert_path = os.environ.get('DOCKER_CERT_PATH') or None
+
+    # empty string for tls verify counts as "false".
+    # Any value or 'unset' counts as true.
     tls_verify = os.environ.get('DOCKER_TLS_VERIFY')
+    if tls_verify == '':
+        tls_verify = False
+        enable_tls = True
+    else:
+        tls_verify = tls_verify is not None
+        enable_tls = cert_path or tls_verify
 
     params = {}
 
     if host:
         params['base_url'] = (host.replace('tcp://', 'https://')
-                              if tls_verify else host)
+                              if enable_tls else host)
 
-    if tls_verify and not cert_path:
+    if not enable_tls:
+        return params
+
+    if not cert_path:
         cert_path = os.path.join(os.path.expanduser('~'), '.docker')
 
-    if tls_verify and cert_path:
-        params['tls'] = tls.TLSConfig(
-            client_cert=(os.path.join(cert_path, 'cert.pem'),
-                         os.path.join(cert_path, 'key.pem')),
-            ca_cert=os.path.join(cert_path, 'ca.pem'),
-            verify=True,
-            ssl_version=ssl_version,
-            assert_hostname=assert_hostname)
+    if not tls_verify and assert_hostname is None:
+        # assert_hostname is a subset of TLS verification,
+        # so if it's not set already then set it to false.
+        assert_hostname = False
+
+    params['tls'] = tls.TLSConfig(
+        client_cert=(os.path.join(cert_path, 'cert.pem'),
+                     os.path.join(cert_path, 'key.pem')),
+        ca_cert=os.path.join(cert_path, 'ca.pem'),
+        verify=tls_verify,
+        ssl_version=ssl_version,
+        assert_hostname=assert_hostname,
+        assert_fingerprint=tls_verify)
 
     return params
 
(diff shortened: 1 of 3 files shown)


docker/utils/utils.py
```diff
@@ -445,26 +445,45 @@ def parse_devices(devices):
 
 def kwargs_from_env(ssl_version=None, assert_hostname=None):
     host = os.environ.get('DOCKER_HOST')
-    cert_path = os.environ.get('DOCKER_CERT_PATH')
+
+    # empty string for cert path is the same as unset.
+    cert_path = os.environ.get('DOCKER_CERT_PATH') or None
+
+    # empty string for tls verify counts as "false".
+    # Any value counts as true.
```
dnephin: Minor: I think could be clarified as "Any value or 'unset' counts as true"

I had to look at this a couple times before I got it.

dnephin:
I haven't looked over the docker cli logic, but I think this makes sense. I know we've had multiple requests for this in Compose.

LGTM

mikedougherty:
The main difference between this and the docker CLI is that this infers TLS is enabled if either value is set. With all of env variables set as shown, I still have to specify `docker --tls ...` (same if I also provide a `DOCKER_CERT_PATH`).

docker/utils/utils.py
mikedougherty: Will update - that was the best I could do at the time, I like yours :)

shin-:
LGTM, thanks! :)

mikedougherty:
rebased on master :+1:

shin-:
Looks like `KwargsFromEnvTest.test_kwargs_from_env_tls_verify_false_no_cert` is failing on Travis!

mikedougherty:
Ah, looks like I didn't follow the pattern of overriding `HOME` for a test that uses the default cert path. working on a fix now.

shin-:
Cool, thanks!

mountkin:
This change is breaking the docker integration tests.
According https://github.com/docker/docker/blob/master/docker/common.go#L28 , if the environment variable `DOCKER_TLS_VERIFY` is an empty string, `tls_verify` is treated as `False` and TLS is not enabled.

This PR treated an empty `DOCKER_TLS_VERIFY` environment variable as if the user want to disable tls verify but enable tls.
Meanwhile, docker integration test is passing an empty `DOCKER_TLS_VERIFY` environment variable  to the docker-py test https://github.com/docker/docker/blob/master/hack/make.sh#L242

docker/utils/utils.py
```diff
@@ -445,26 +445,45 @@ def parse_devices(devices):
...
+
+    # empty string for tls verify counts as "false".
+    # Any value or 'unset' counts as true.
     tls_verify = os.environ.get('DOCKER_TLS_VERIFY')
+    if tls_verify == '':
+        tls_verify = False
+        enable_tls = True
+    else:
+        tls_verify = tls_verify is not None
+        enable_tls = cert_path or tls_verify
```
mountkin: I think the block above should be changed to 
```
    tls_verify = os.environ.get('DOCKER_TLS_VERIFY') or False
    enable_tls = tls_verify or cert_path or False
```
to keep align with the Docker CLI flags.

shin-:
Yeah, this is addressed by #988

docker/utils/utils.py
mikedougherty: If I understand correctly that would enable my use case, so I'd be fine with that. However it seems like it would be better to enable verification by default and require an explicit setting to turn it off (specifically, setting DOCKER_TLS_VERIFY to a falsy value).

Also you don't need the trailing `or False`, it would be more pythonic to write your snippet as:

```python
tls_verify = os.environ.get('DOCKER_TLS_VERIFY')
enable_tls = tls_verify or cert_path
```

or perhaps wrap the values in `bool` such as `bool(os.environ.get('DOCKER_TLS_VERIFY'))`

PR #928, opened 2016-02-09
Never exclude Dockerfile from the build context.

### LINKED ISSUE #926: [1.7] regression in .dockerignore handling
If the `Dockerfile` is being ignored by a path in the `.dockerignore` file, it is incorrectly being removed from the context. There is a special case handling when the file is being excluded directly, but it should also apply when there is a path which includes the `Dockerfile`.

Possibly caused by #863

```
docker-py version: 1.7.0

$ tree -a
.
├── Docker
│   ├── dc.yml
│   └── Dockerfile.debug
└── .dockerignore

1 directory, 3 files

$ cat Docker/Dockerfile.debug

FROM alpine:3.3
RUN  echo building
CMD  echo run

$ cat Docker/dc.yml

app:
  build: ..
  dockerfile: Docker/Dockerfile.debug

$ cat .dockerignore
Docker/

docker-compose -f Docker/dc.yml build
ERROR: Cannot locate specified Dockerfile: Docker/Dockerfile.debug
```

shin-:
Fixes #926

### DIFF
--- a/docker/utils/utils.py
+++ b/docker/utils/utils.py
@@ -128,7 +128,13 @@ def exclude_paths(root, patterns, dockerfile=None):
     paths = get_paths(root, exclude_patterns, include_patterns,
                       has_exceptions=len(exceptions) > 0)
 
-    return set(paths)
+    return set(paths).union(
+        # If the Dockerfile is in a subdirectory that is excluded, get_paths
+        # will not descend into it and the file will be skipped. This ensures
+        # it doesn't happen.
+        set([dockerfile])
+        if os.path.exists(os.path.join(root, dockerfile)) else set()
+    )
 
 
 def should_include(path, exclude_patterns, include_patterns):
--- a/tests/unit/utils_test.py
+++ b/tests/unit/utils_test.py
@@ -736,6 +736,7 @@ class ExcludePathsTest(base.BaseTestCase):
         'foo/b.py',
         'foo/bar/a.py',
         'bar/a.py',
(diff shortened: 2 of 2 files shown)


docker/utils/utils.py
```diff
@@ -128,7 +128,13 @@ def exclude_paths(root, patterns, dockerfile=None):
     paths = get_paths(root, exclude_patterns, include_patterns,
                       has_exceptions=len(exceptions) > 0)
 
-    return set(paths)
+    return set(paths).union(
+        # If the Dockerfile is in a subdirectory that is excluded, get_paths
+        # will not descend into it and the file will be skipped. This ensures
+        # it doesn't happen.
+        set([dockerfile])
+        if os.path.exists(os.path.join(root, dockerfile)) else {}
```
dnephin: I doesn't seem like we do this `os.path.exists()` check in other places, is it necessary?

dnephin: and I believe the default should be `set()` not `{}`, right?

shin-: If utils.tar() is called on a folder that doesn't contain a Dockerfile, it would break without this change. It doesn't affect docker-py directly I think (we have some tar unit tests that break), but it's part of our public API and some people seem to use it: https://github.com/docker/docker-py/pull/865

shin-: Yeah, `set()`. Implicit type cast coming to the rescue obviously :)

dnephin:
LGTM

PR #930, opened 2016-02-09
Use gzip to compress tar balls before sending.

speedplane:
Seems like a straightforward improvement to gzip our tarballs before sending them. For me, it greatly improved performance (reduced my own docker image from 120MB to 40MB).

Signed-off-by: Michael Sander <michael.sander@docketalarm.com

### DIFF
--- a/docker/api/build.py
+++ b/docker/api/build.py
@@ -46,7 +46,8 @@ class BuildApiMixin(object):
             if os.path.exists(dockerignore):
                 with open(dockerignore, 'r') as f:
                     exclude = list(filter(bool, f.read().splitlines()))
-            context = utils.tar(path, exclude=exclude, dockerfile=dockerfile, gzip=True)
+            context = utils.tar(path, exclude=exclude, dockerfile=dockerfile,
+                gzip=True)
             encoding = 'gizp'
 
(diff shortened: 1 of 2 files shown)


speedplane:
Also, FYI, I caught this while experiencing pain with Docker on Google's managed VM.
https://code.google.com/p/google-cloud-sdk/issues/detail?id=456

docker/api/build.py
```diff
@@ -46,7 +46,9 @@ def build(self, path=None, tag=None, quiet=False, fileobj=None,
             if os.path.exists(dockerignore):
                 with open(dockerignore, 'r') as f:
                     exclude = list(filter(bool, f.read().splitlines()))
-            context = utils.tar(path, exclude=exclude, dockerfile=dockerfile)
+            context = utils.tar(path, exclude=exclude, dockerfile=dockerfile,
+                gzip=True)
+            encoding = 'gizp'
```
TomasTomecek: typo

dnephin: I think there should be a kwarg `gzip_context` (default to `False`) to enable it. There may be cases where it's undesirables, and defaulting it to false keeps it backwards compatible.

The new kwarg will need a small docs addition.

shin-:
Closing via #1008

PR #932, opened 2016-02-09
Add a Progress Printer for Uploading

speedplane:
Uploading a docker file takes time. Adding a progress dialog reduces stress for everyone. Also, this commit adds a small 32k buffer for the upload. Previously, it was not really buffered, and we performing many thousands of unnecessary file reads, adding a significant overhead.

Signed-off-by: Michael Sander michael.sander@docketalarm.com

### DIFF
--- a/docker/api/build.py
+++ b/docker/api/build.py
@@ -89,9 +89,7 @@ class BuildApiMixin(object):
             if os.path.exists(dockerignore):
                 with open(dockerignore, 'r') as f:
                     exclude = list(filter(bool, f.read().splitlines()))
-            context = utils.tar(path, exclude=exclude, dockerfile=dockerfile, 
-                gzip=True)
-            encoding = 'gzip'
+            context = utils.tar(path, exclude=exclude, dockerfile=dockerfile)
 
         if utils.compare_version('1.8', self._version) >= 0:
             stream = True
--- a/docker/api/build.py
+++ b/docker/api/build.py
@@ -46,7 +46,11 @@ class FileUploadBuffer(object):
         if not len(out):
             raise StopIteration
         return out
-        
+    
+    def __next__(self):
+        # For python 3
+        return self.next()
+    
     def read(self, *args, **kwargs):
         return self.chk_print(fn = lambda: self.file.read(*args, **kwargs))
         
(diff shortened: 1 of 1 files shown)


docker/api/build.py
```diff
@@ -42,6 +83,8 @@ def build(self, path=None, tag=None, quiet=False, fileobj=None,
             raise TypeError("You must specify a directory to build in path")
         else:
             dockerignore = os.path.join(path, '.dockerignore')
+            # If there are a lot of files, this takes time. Tell the user.
+            log.debug('taring and gzipping files: %s'%path)
```
dnephin: I don't think this is being gzipped.

The formatting is off a bit, there should be spaces before and after `%`.

I believe it's also common practice to use double quotes for strings that are messages to users (and single quotes for strings that are internal constants).

docker/api/build.py
```diff
@@ -11,7 +12,47 @@
...
+        return out
+        
+    def next(self):
+        out = self.read(self.BUFFER_SIZE)
+        if not len(out):
+            raise StopIteration
+        return out
+        
+    def read(self, *args, **kwargs):
+        return self.chk_print(fn = lambda: self.file.read(*args, **kwargs))
```
dnephin: This function seems unnecessary. Why not just move the implementation of `chk_print` here?

docker/api/build.py
```diff
@@ -11,7 +12,47 @@
...
+class FileUploadBuffer(object):
+    '''
+    Wrap a file object, buffer, and print progress.
+    '''
+    BUFFER_SIZE = 1024*32
+    def __init__(self, file):
+        self.read_so_far = 0
+        self.num_reads = 0
+        self.starttime = datetime.datetime.utcnow()
+        self.file = file
```
dnephin: I like the idea of this abstraction, however I think reporting total reads is a bit weird.

I think it would be good to first get the full size of `file`, and report "send / total" like the docker-cli does.

This class implements both the iterator interface and part of the file interface. Are both necessary?

speedplane: - I can remove total reads, I initially just used it to properly size the buffer to something reasonable, but it's probably not necessary anymore.
- As it stands, this function does not rely on fseek or ftell, which would be required to implement the file size. Not sure we want to add that dependency.

dnephin: I think `seek()`, and `tell()` should be available on the `file` object already

ulyssessouza:
Stale

Issue #935, opened 2016-02-10
Insert IPv4 and IPv6 address for container on /networks/<id>/connect call


mdaue:
Docker API 1.22 allows to specify static IPv4 and IPv6 addresses. This is an insert into /networks/<id>/connect call by the network id.

shin-:
Closing via #936

PR #936, opened 2016-02-10
Implements #935 - Insert IPv4 and IPv6 call in /networks/id/connect

mdaue:
Requires API 1.22

Unit tests passed, had some trouble running integration tests:
E   APIError: 400 Client Error: Bad Request ("client is newer than server (client API version: 1.22, server API version: 1.21)")

Docker version:
Client:
 Version:      1.11.0-dev
 API version:  1.23
 Go version:   go1.5.3
 Git commit:   415dd86
 Built:        Sun Feb  7 21:47:30 2016
 OS/Arch:      linux/amd64

Server:
 Version:      1.11.0-dev
 API version:  1.23
 Go version:   go1.5.3
 Git commit:   415dd86
 Built:        Sun Feb  7 21:47:30 2016
 OS/Arch:      linux/amd64

### DIFF
--- a/docker/api/network.py
+++ b/docker/api/network.py
@@ -1,6 +1,8 @@
 import json
 
+from ..errors import InvalidVersion
 from ..utils import check_resource, minimum_version, normalize_links
+from ..utils import version_lt
 
 
 class NetworkApiMixin(object):
@@ -48,6 +50,7 @@ class NetworkApiMixin(object):
     @check_resource
     @minimum_version('1.21')
     def connect_container_to_network(self, container, net_id,
+                                     ipv4_address=None, ipv6_address=None,
                                      aliases=None, links=None):
         data = {
             "Container": container,
@@ -56,6 +59,21 @@ class NetworkApiMixin(object):
                 "Links": normalize_links(links) if links else None,
             },
         }
+
+        # IPv4 or IPv6 or neither:
+        if ipv4_address or ipv6_address:
+            if version_lt(self._version, '1.22'):
+                raise InvalidVersion('IP address assignment is not '
+                                     'supported in API version < 1.22')
+
+            data['EndpointConfig']['IPAMConfig'] = dict()
+            if ipv4_address:
+                data['EndpointConfig']['IPAMConfig']['IPv4Address'] = \
+                    ipv4_address
+            if ipv6_address:
+                data['EndpointConfig']['IPAMConfig']['IPv6Address'] = \
+                    ipv6_address
+
         url = self._url("/networks/{0}/connect", net_id)
         res = self._post_json(url, data=data)
         self._raise_for_status(res)
--- a/tests/integration/network_test.py
+++ b/tests/integration/network_test.py
@@ -235,3 +235,51 @@ class TestNetworks(helpers.BaseTestCase):
         )
 
         self.execute(container, ['nslookup', 'bar'])
+
+    @requires_api_version('1.22')
+    def test_connect_with_ipv4_address(self):
+        net_name, net_id = self.create_network()
+
+        container = self.create_and_start(
+            host_config=self.client.create_host_config(network_mode=net_name))
+
+        self.client.disconnect_container_from_network(container, net_name)
+        self.client.connect_container_to_network(
+            container, net_name,
+            ipv4_address='192.168.0.1')
+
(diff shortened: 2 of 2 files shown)


docker/api/network.py
```diff
@@ -46,9 +46,10 @@ def inspect_network(self, net_id):
         return self._result(res, json=True)
 
     @check_resource
-    @minimum_version('1.21')
+    @minimum_version('1.22')
```
dnephin: This isn't correct, the function needs to stay at `1.21`. You can add an `if` branch within the function to require a higher version for specific params (there are examples of this already).

tests/integration/network_test.py
```diff
@@ -100,7 +100,7 @@ def test_remove_network(self):
         self.client.remove_network(net_id)
         self.assertEqual(len(self.client.networks()), initial_size)
 
-    @requires_api_version('1.21')
+    @requires_api_version('1.22')
```
dnephin: same here

tests/integration/network_test.py
```diff
@@ -100,7 +100,7 @@ def test_remove_network(self):
...
     def test_connect_and_disconnect_container(self):
         net_name, net_id = self.create_network()
```
dnephin: same here

dnephin:
Thanks for the PR. I left aa couple comments.

Once it's ready to go, we'll also ask you to squash your commits down to one or two.

mdaue:
Great! I squashed to one commit, leaving interesting comments in the notes. I set the minimum required API back to 1.21 where identified and compared with master before committing. I removed my addition to the unit test as it failed, possibly because it wasn't run with API 1.22. Alternatively, I can add that test again and perform the same version check.

Thanks!

docker/api/network.py
```diff
@@ -56,6 +58,16 @@ def connect_container_to_network(self, container, net_id,
                 "Links": normalize_links(links) if links else None,
             },
         }
+
+        # IPv4 or IPv6 or neither:
+        if version_gte(self._version, '1.22'):
```
dnephin: I think this needs to be the other way around. 

```python
if version_lt(...):
    raise errors.InvalidVersion(...)

```

mdaue:
Just in case it was missed, I did update the code base on the suggestions and squashed the commit.

docker/api/network.py
```diff
@@ -56,6 +59,19 @@ def connect_container_to_network(self, container, net_id,
                 "Links": normalize_links(links) if links else None,
             },
         }
+
+        # IPv4 or IPv6 or neither:
+        if ip4 or ip6:
+            if version_lt(self._version, '1.22'):
+                raise InvalidVersion('IP address assignment is not in '
+                                     'supported in API ver < 1.2')
```
shin-: Let's spell out "version" entirely 
+ typo `1.22`

docker/api/network.py
```diff
@@ -48,7 +50,8 @@ def inspect_network(self, net_id):
     @check_resource
     @minimum_version('1.21')
     def connect_container_to_network(self, container, net_id,
-                                     aliases=None, links=None):
+                                     ip4=None, ip6=None, aliases=None,
```
shin-: Let's call those `ipv4_address` an `ipv6_address`. The additional characters are worth the clarity they provide, I think.

shin-:
A few small comments, should be good after that.

mdaue:
Thanks - made changes.

shin-:
LGTM :+1:

tests/integration/network_test.py
```diff
@@ -235,3 +235,33 @@ def test_connect_with_links(self):
...
+        container_data = self.client.inspect_container(container)
+        self.assertEqual(
+            container_data['NetworkSettings']['Networks'][net_name]
+                          ['IPv4Address'],
+            '192.168.0.1')
+
+        self.assertEqual(
+            container_data['NetworkSettings']['Networks'][net_name]
+                          ['IPv6Address'],
+            '2001:389::1')
```
dnephin: Minor: This could probably be two separate tests

dnephin:
LGTM

mdaue:
I split the integration tests as per @dnephin suggestion. Also found that I had forgotten to change the arg to the function calls in the test (ip4 vs. ipv4_address) after making the change in network.py (fixed).

PR #938, opened 2016-02-10
Synthesize executable bit on Windows

### LINKED ISSUE #937: Executable permissions not synthesized when building from Windows
When building a Docker image the build context is tarred and sent to the Docker daemon. However when you build an image on a Windows client the build context's permissions are slightly messed up due to Unix and Windows differences. This is problematic because of missing executable permissions, which might render an image built from a Windows client unusable.

`docker build` just synthesizes the executable bit when you're building from Windows, however docker-py does not. This leads to very surprising situations where images built with `docker build` are perfectly fine, but images build with `docker-compose build` are broken because the entrypoint is not executable.

docker-py should follow `docker build`'s decision and just set the executable bit on every file in the tarball in order to prevent such breakage and to be consistent with the Docker CLI.

See the following for more details:

- "Unix file permissions and permission bits" in @ahmetalpbalkan's blog post about [Porting Docker Client to Windows](https://ahmetalpbalkan.com/blog/porting-docker-client-to-windows/)
- docker/docker#11047
- docker/docker@1a22418f9f1573ca8521831d52c8e0562cb3ef8f

seschwar:
Be consistent with `docker build`, which marks the files in the context tarball as executable. See also https://github.com/docker/docker/issues/11047.

Fixes #937.

This should work, but probably needs some massaging and tests before it can be merged.

### DIFF
--- a/docker/utils/utils.py
+++ b/docker/utils/utils.py
@@ -18,6 +18,7 @@ import os
 import os.path
 import json
 import shlex
+import sys
 import tarfile
 import tempfile
 import warnings
@@ -100,7 +101,10 @@ def tar(path, exclude=None, dockerfile=None, fileobj=None):
     exclude = exclude or []
 
     for path in sorted(exclude_paths(root, exclude, dockerfile=dockerfile)):
-        t.add(os.path.join(root, path), arcname=path, recursive=False)
+        i = t.gettarinfo(os.path.join(root, path), arcname=path)
+        if sys.platform == 'win32':
+            i.mode = i.mode & 0o755 | 0o111
+        t.addfile(i)
 
     t.close()
     fileobj.seek(0)


docker/utils/utils.py
```diff
@@ -100,7 +101,14 @@ def tar(path, exclude=None, dockerfile=None, fileobj=None):
     exclude = exclude or []
 
     for path in sorted(exclude_paths(root, exclude, dockerfile=dockerfile)):
-        t.add(os.path.join(root, path), arcname=path, recursive=False)
+        i = t.gettarinfo(os.path.join(root, path), arcname=path)
+        if sys.platform == 'win32':
+            i.mode = i.mode & 0o755 | 0o111
+        if i.isreg():
+            with open(os.path.join(root, path), 'rb') as f:
+                t.addfile(i, f)
```
dnephin: I don't think this `if i.isref()` branch is necessary. I think the warning in the docs is only for the case where you're passing a `fileobj`. I don't think it means that it is required to pass it on windows.

Did you run into problems without it?

docker/utils/utils.py
```diff
@@ -100,7 +101,14 @@ def tar(path, exclude=None, dockerfile=None, fileobj=None):
     exclude = exclude or []
 
     for path in sorted(exclude_paths(root, exclude, dockerfile=dockerfile)):
-        t.add(os.path.join(root, path), arcname=path, recursive=False)
+        i = t.gettarinfo(os.path.join(root, path), arcname=path)
+        if sys.platform == 'win32':
+            i.mode = i.mode & 0o755 | 0o111
```
dnephin: This looked weird, but it does match `pkg/archive/archive_windows.go` in docker/docker, so I guess we should keep it this way. Not sure why write is stripped from group/other.

docker/utils/utils.py
seschwar: I did play around with the `tarfile` module in the Python interpreter on Linux and was just getting zeroed files in my tarball without it. However `docker.Client.build()` seems to work fine without it. I guess we can drop it then.

docker/utils/utils.py
seschwar: Yes, I am just following `docker build`'s behaviour.

seschwar:
Updated PR to remove seemingly superfluous branch.

shin-:
Closing via #1228

PR #939, opened 2016-02-11
Add Tmpfs support to the python API

janLo:
This adds support for the recently (1.10) added "--tmpfs" option in docker.

It is used to made private temporary filesystem mounts inside the container,
even if the container itself is not privileged to mount anything. Its a bit modeled
after the "binds" feature for volumes. The first commit provide the necessary
code changes. The second provides some unittests and the third adds
documentation.

You can see details about the tmpfs feature in the related docker PR:
https://github.com/docker/docker/pull/13587

### DIFF
--- a/docker/utils/utils.py
+++ b/docker/utils/utils.py
@@ -337,6 +337,33 @@ def convert_volume_binds(binds):
     return result
 
 
+def convert_tmpfs_mounts(tmpfs):
+    if isinstance(tmpfs, dict):
+        return tmpfs
+
+    if not isinstance(tmpfs, list):
+        raise ValueError(
+            'Tmpfs spec must be a list'
+        )
+
+    result = {}
+    for mount in tmpfs:
+        if isinstance(mount, six.string_types):
+            if ":" in mount:
+                name, options = mount.split(":", 1)
+            else:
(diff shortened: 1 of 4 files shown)


janLo:
Can someone explain why the unittests are failling?

dnephin:
The expected output doesn't match the actual output. Look at `tmpfs /mnt`

janLo:
Ah. A typo. Should be fixed now.

janLo:
Wtf, They ARE signed ...

docker/utils/utils.py
```diff
@@ -337,6 +337,33 @@ def convert_volume_binds(binds):
...
 
 
+def convert_tmpfs_mounts(tmpfs):
+    if isinstance(tmpfs, dict):
+        return tmpfs
+
+    if not isinstance(tmpfs, list):
+        raise ValueError(
+            'Tmpfs spec must be a list'
+        )
```
dnephin: There seems to be some inconsistencies in the error messages here. This one says must be a list, but the previous branch accepted dicts.  The next one says they must be something else.

dnephin:
Merged as part of #994

PR #942, opened 2016-02-15
Separate params stream and follow for logs.

### LINKED ISSUE #934: logs() separate param for stream and follow
From: https://github.com/docker/compose/pull/2720/files#r52222296

Current the `follow` param is set based on `stream`

I think `follow=True` does imply `stream=True`, but `stream=True` doesn't imply `follow=True`, you may still want to stream without following.

seguins:
This closes #934

### DIFF
--- a/docker/api/container.py
+++ b/docker/api/container.py
@@ -193,12 +193,14 @@ class ContainerApiMixin(object):
 
     @utils.check_resource
     def logs(self, container, stdout=True, stderr=True, stream=False,
-             timestamps=False, tail='all', since=None):
+             timestamps=False, tail='all', since=None, follow=None):
         if utils.compare_version('1.11', self._version) >= 0:
+            if follow is None:
+                follow = stream
             params = {'stderr': stderr and 1 or 0,
                       'stdout': stdout and 1 or 0,
                       'timestamps': timestamps and 1 or 0,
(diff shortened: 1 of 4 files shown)


docker/api/container.py
```diff
@@ -193,12 +193,12 @@ def kill(self, container, signal=None):
...
     @utils.check_resource
     def logs(self, container, stdout=True, stderr=True, stream=False,
-             timestamps=False, tail='all', since=None):
+             timestamps=False, tail='all', since=None, follow=False):
         if utils.compare_version('1.11', self._version) >= 0:
             params = {'stderr': stderr and 1 or 0,
                       'stdout': stdout and 1 or 0,
                       'timestamps': timestamps and 1 or 0,
-                      'follow': stream and 1 or 0,
+                      'follow': follow and 1 or 0,
```
dnephin: I wonder if we should set `follow=None` and do:

```python

'follow': follow if follow is not None else int(stream)
```

to keep it backwards compatible

dnephin:
Looks good, just one consideration for backwards compatibility.

seguins:
I have added the backward compatibility and I have created a new unit test for it.

dnephin:
LGTM

shin-:
LGTM, thanks!

PR #239, opened 2014-06-10
Passing stdin to request attach.

Piwosz:
Simple implementation of passing body content to post request.

PR #946, opened 2016-02-22
Pass stdin data on attach

intrepidlemon:
Currently, there is no simple way of passing in `stdin` content into an attached container. The implementation provided by #239 did not work for me. This PR adds an additional parameter where one can pass in string or byte data to be written into the `stdin` of an attached container. 

In addition, this PR also contains: 
- Simple documentation for `stdin`
- Integration tests for `stdin`
- Unit tests for `attach`

### DIFF
--- a/docker/api/container.py
+++ b/docker/api/container.py
@@ -10,15 +10,19 @@ from ..utils.utils import create_networking_config, create_endpoint_config
 class ContainerApiMixin(object):
     @utils.check_resource
     def attach(self, container, stdout=True, stderr=True,
-               stream=False, logs=False):
+               stdin=None, stream=False, logs=False):
         params = {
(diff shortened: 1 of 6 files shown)


SakuraSound:
I commented in #239 about a workaround for Python 3 that was brought up by @Slix. While I would like an easier way to pass something into stdin upon creation of a container, i think the workaround is actually the right way to go about it if you want to keep your container running and accept more stdin input.

bfirsh:
This would be great, but unfortunately we've had a pretty major API change in the development version. Would you mind rebasing, @intrepidlemon?

intrepidlemon:
cleanup

paddie:
@bfirsh so it seems all progress on getting an interactive container has stopped?

PR #947, opened 2016-02-22
Rewrite README in RST, which PyPI will render

hobarrera:
[PyPI](https://pypi.python.org/pypi/docker-py) render RST, but not markdown (as you can see, it looks quite ugly right now).
This PR merely converts the README into RST, and nothing else. The GitHub page remains identical.


hobarrera:
Looks like the bot is bugged and thinks the commits are still unsigned. :(

shin-:

However, we're not looking to switch to RST at the moment. The resulting lack of format on PyPi is unfortunate, but ultimately trivial.

hobarrera:
PyPi is *the* official source for python packages; is there some issue for explicitly wanting to *not* render the README there?

At a first glance, with a broken readme, the project looked like an unpolished library, which is definitely is not.

dnephin:
This is what we do for Compose to convert the markdown to rst before publishing to pypi:

https://github.com/docker/compose/blob/master/script/release/push-release#L57-L66

PR #950, opened 2016-02-23
Fix TLS regressions

aanand:
- Set `verify` to the CA cert path if one has been specified, rather than `True`
- Don't set `assert_fingerprint`


dnephin:
LGTM, based on my understanding of the problem

An integration test against a mock tls server would be great, but not something we have time for atm.

PR #952, opened 2016-02-23
Format environment variables to match docker-cli.

### LINKED ISSUE #948: Support environment variables without values
See https://github.com/docker/compose/pull/2997 for some context.

docker-py assumes that all variables should have an `=` but the docker-cli passes environment variables without values as just the key, which is super weird, but apparently swarm is relying on that currently.

dnephin:
Fixes #948

### DIFF
--- a/docker/utils/utils.py
+++ b/docker/utils/utils.py
@@ -821,6 +821,14 @@ def split_command(command):
     return shlex.split(command)
 
 
+def format_environment(environment):
+    def format_env(key, value):
+        if not value:
+            return key
+        return '{key}={value}'.format(key=key, value=value)
+    return [format_env(*var) for var in six.iteritems(environment)]
+
+
 def create_container_config(
     version, image, command, hostname=None, user=None, detach=False,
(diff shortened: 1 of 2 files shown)


tests/integration/container_test.py
```diff
@@ -368,7 +368,17 @@ def test_create_host_config_exception_raising(self):
                           self.client.create_host_config, mem_swappiness='40')
 
         self.assertRaises(ValueError,
-                          self.client.create_host_config, pid_mode='40')
+                           self.client.create_host_config, pid_mode='40')
```
shin-: You've got an extra space here making flake8 break.

shin-:
LGTM!

docker/utils/utils.py
```diff
@@ -821,6 +821,14 @@ def split_command(command):
     return shlex.split(command)
 
 
+def format_environment(environment):
+    def format_env(key, value):
+        if not value:
```
dnephin: hmm, I just looked at this again, and maybe this needs to be `if value is None`

shin-: ah, yeah, good point.

PR #954, opened 2016-02-24
Readme improvements

bfirsh:
Improved description and updated link text.

The link text is a little thing, but this made me hesitate before clicking the link because I thought I would be going to the Read the Docs home page, not the docker-py documentation.


shin-:
thanks!

PR #956, opened 2016-02-24
Add docker.from_env() shortcut

bfirsh:
A much neater way of getting started with docker-py. Also updated the boot2docker docs to reflect the current Toolbox/Machine world.

A question: are the security concerns of doing `assert_hostname=False` true? The client is validating the specific cert that the server has, I think, so there is no need to also check the hostname. I think we would still have to specifically say "if you are using Machine, this is safe", because in the context of using a normal CA cert it would not be safe.


docker/client.py
```diff
@@ -28,10 +28,14 @@
 from .auth import auth
 from .unixconn import unixconn
 from .ssladapter import ssladapter
-from .utils import utils, check_resource, update_headers
+from .utils import utils, check_resource, update_headers, kwargs_from_env
 from .tls import TLSConfig
 
 
+def from_env(**kwargs):
```
shin-: I'm torn on the necessity of having this function in the module namespace when people could simply call `docker.Client.from_env()`

docs/machine.md
```diff
@@ -0,0 +1,30 @@
+# Using with Docker Toolbox and Machine
+
+In development, we recommend using [Docker Toolbox](https://www.docker.com/products/docker-toolbox) to set up Docker. It includes a tool called Machine which will create a VM running Docker Engine and point your shell at it using environment variables.
```
shin-: I dislike the phrasing "we recommend using Docker Toolbox", because while I understand it's the product we at Docker want people to use, I feel it's irrelevant to docker-py in particular.

Can we keep a more neutral approach here?

tests/unit/client_test.py
```diff
@@ -0,0 +1,26 @@
...
+
+
+class ClientTest(base.BaseTestCase):
+    def setUp(self):
+        self.os_environ = os.environ.copy()
+
+    def tearDown(self):
+        os.environ = self.os_environ
+
+    def test_from_env(self):
```
shin-: This can probably be a method of `utils_test.KwargsFromEnvTest` instead.

shin-:
> A question: are the security concerns of doing assert_hostname=False true? [...]
> I think we would still have to specifically say "if you are using Machine, this is safe", because in the context of using a normal CA cert it would not be safe.

That's fair, it should probably be fine to do so for a Machine environment. I believe the warning was written in a broader context, which is also why the recommendation was kept fairly vague.

docs/machine.md
bfirsh: It _is_ the easiest way to get started, though! It's what 95% of users will want to do.

tests/unit/client_test.py
bfirsh: I felt it was more suitable to associate it with `Client` because it was testing a code path in there.

docker/client.py
bfirsh: I really like the neatness of it. :D 

@aanand WDYT?

docs/machine.md
TomasTomecek: That's a pretty bold statement. Have you asked those 95% of users?

bfirsh: To be clear: the vast majority of **first-time users in development** are going to have the best experience if we recommend using Toolbox, so we put that front and centre. For users doing something more complex, they are probably knowledgeable enough that they don't need this handholding.

bfirsh: (Also – not all of them, but I talk to a lot of users so I have a pretty good sample size. ;)

docker/client.py
aanand: I like the neatness too. It's always a tiny mental speed-bump when I have to learn a new class when starting out with a library, so having that removed is :+1:

bfirsh:
@shin- Updated the docs to say that `assert_hostname=False` is safe if you're using Machine.

aanand:
LGTM - needs rebase

bfirsh:
Rebased - thanks.

PR #967, opened 2016-02-29
Support cidfile option in create command

JesusTinoco:


dnephin:
Are you sure this is a client-side feature? Since it's configured by `HostConfig` I would expect it to be handled by the daemon.

JesusTinoco:
@dnephin I think so, it seems that the API Rest provided by docker only set the ContainerIDFile and doesn't create the file. Checking the  [api client](https://github.com/docker/docker/blob/df2b74188ec51422e84ec1dbdc58abf08c215019/api/client/create.go) used by docker seems that it creates that file as well. Am I wrong?

JesusTinoco:
@dnephin any new update?

dnephin:
you're right it is client side.

JesusTinoco:
@dnephin thanks for the reply! Should I open and issue? or just wait for your review?

JesusTinoco:
Do you think that this pull request could be interesting for docker-py? It would help to implement a solution for this issue: https://github.com/docker/compose/issues/940

MateusAmin:
@JesusTinoco Thanks for the contribution! I was looking into doing this myself. If @dnephin thinks this needs anymore work I would be happy to help out if wanted.

I took a brief look at the commit and based on my meager understanding of the code base everything looks good. I think the branch name does not follow the Docker contribution convention: XXXX-something where XXXX is the number of the issue.

jamiehannaford:
@dnephin Apart from a rebase, does this PR need anything else? I would also like this feature.

AndreLouisCaron:
I started working on a fix for docker/compose#940 and I hit a wall when I saw that the `docker-py` doesn't have an option to create the CID file.

I was going to implement the change when I stumbled upon this existing PR.  Any chance we can have this fix included?

Thanks!

shin-:
I don't think it should be a docker-py feature, to be honest. If you want to write a container ID into a file, it's currently very simple to do so from Python.

AndreLouisCaron:
@shin- I think the basic idea was feature parity of `docker-compose` with `docker-run`.  Currently, you can `docker run -cidfile ...`, but you can't do the same from a compose file.

I personally don't mind whether it's implemented inside `docker-py` or internally in `docker-compose`, but why not solve the problem once for everyone in `docker-py`?

shin-:
@AndreLouisCaron That's not what ContainerIDFile does - the file is created on the host, not inside the container.

AndreLouisCaron:
@shin- OK maybe I need to take a step back.  I got here thinking I could use this to let the container know what its own ID is (see suggestions in this StackOverflow question: https://stackoverflow.com/q/20995351/313063).  The other suggestion seems to be to use `$HOSTNAME`, but that only gives a short ID.

shin-:
@AndreLouisCaron Yeah - that issue is outside the scope of the Python SDK.

mschuckmann:
> @AndreLouisCaron That's not what ContainerIDFile does - the file is created on the host, not inside the container.

From reading the Docker documentation and the examples I'm pretty sure the cidfile is created on the container side and it is very useful if you need to know the container ID from within the container, you can't count on $HOSTNAME as it can default to the host's HOSTNAME when using host network mode.

PR #968, opened 2016-03-01
Use decode on push and pull operations.

atopuzov:
Stream helper has the ability to decode the responses from strings to
JSON. This commit enables this functionality on push and pull operations.


dnephin:
LGTM

shin-:
LGTM, thanks!

PR #971, opened 2016-03-02
Use protocol TLS v1.0 by default when none is set.

### LINKED ISSUE #963: tlsv1 alert protocol version on 1.7.1 and 1.7.2 but not on 1.7.0
Similar to #949 I'm discovering issues with latest versions of ``docker-py`` running against docker 1.10.2 instance. I'm using ``docker.utils.kwargs_from_env(assert_hostname=False)``. Things work fine with version 1.7.0.

Docker client is initialized via

```
client = docker.Client(
    version='auto',
    **docker.utils.kwargs_from_env(assert_hostname=False))
```

with docker environment variables being set to the following (via docker-machine)

```
DOCKER_HOST=tcp://192.168.156.137:2376
DOCKER_MACHINE_NAME=dev2
DOCKER_TLS_VERIFY=1
DOCKER_CERT_PATH=/Users/benjixx/.docker/machine/machines/dev2
```

docker-py 1.7.1 and 1.7.2 now raise the following exception:

```
DockerException: Error while fetching server API version: [Errno 1] _ssl.c:507: error:1407742E:SSL routines:SSL23_GET_SERVER_HELLO:tlsv1 alert protocol version
```

Any idea what's happening here?

shin-:
Should fix #963 

People who care about using 1.1 or 1.2 can specify it as an argument to `kwargs_from_env` or when building `TLSConfig`

### DIFF
--- a/docker/tls.py
+++ b/docker/tls.py
@@ -1,4 +1,5 @@
 import os
+import ssl
 
 from . import errors
 from .ssladapter import ssladapter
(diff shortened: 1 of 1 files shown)


TomasTomecek:
so there won't be a 1.7.3 release for this issue, right?

shin-:
@TomasTomecek I don't plan to - see rationale [here](https://github.com/docker/docker-py/issues/963#issuecomment-191909814)

dnephin:
LGTM

This seems to be the most reliable option for now, and users can always override the default if they really care, and know that a newer version is available to them.

Issue #986, opened 2016-03-11
default NetworkMode should be "bridge", not "default"


tbsf:
https://github.com/docker/docker-py/blob/master/docker/utils/utils.py#L646
 
```python
 if network_mode:
        host_config['NetworkMode'] = network_mode
    elif network_mode is None and compare_version('1.19', version) > 0:
        host_config['NetworkMode'] = 'default'
```
  
should be:

```python
  if network_mode:
        host_config['NetworkMode'] = network_mode
    elif network_mode is None and compare_version('1.19', version) > 0:
        host_config['NetworkMode'] = 'bridge'
```
 
The value of "default" gives an error like the following:

```
FAILED! => {"changed": true, "failed": true, "msg": "Docker API Error: Cannot start container 400145f2e47c2f686abf1513d9d0c49b1fce1e6c8c9f2303c24c61085d5b094c: invalid network mode: default"}
```

Running on Mac OS 10.11.3

```
ansible --version
ansible 2.0.1.0
  config file = /etc/ansible/ansible.cfg
  configured module search path = Default w/o overrides
```
Host is 
Ubuntu 15.10

```docker-py==1.7.2
Python 2.7.10
Client version: 1.6.2
Client API version: 1.18
Go version (client): go1.5.1
Git commit (client): 7c8fca2
OS/Arch (client): linux/amd64
Server version: 1.6.2
Server API version: 1.18
Go version (server): go1.5.1
Git commit (server): 7c8fca2
OS/Arch (server): linux/amd64
```

tbsf:
Created PR with this fix:

https://github.com/docker/docker-py/pull/987

shin-:
It seems to be working fine for me.
```python
In [1]: import docker

In [2]: c = docker.Client()

In [3]: c.create_host_config()
Out[3]: {'NetworkMode': 'default'}

In [4]: c.create_container('busybox', 'top', host_config=_3)
Out[4]: 
{u'Id': u'd8a885ec91c58e36c441ed4825528f7e047dfcde5b6678b867afe798a3b32c7d',
 u'Warnings': None}

In [5]: c.start(_4)

In [7]: c.containers()
Out[7]: 
[{u'Command': u'top',
  u'Created': 1457728245,
  u'HostConfig': {u'NetworkMode': u'default'},
  u'Id': u'd8a885ec91c58e36c441ed4825528f7e047dfcde5b6678b867afe798a3b32c7d',
  u'Image': u'busybox',
[...]

PR #987, opened 2016-03-11
Change default value of NetworkMode to 'bridge'

tbsf:
This is a Fix for Docker API >= 1.19.

The default network is 'bridge' according to :

https://docs.docker.com/engine/userguide/networking/default_network/

and 'default' not listed as a valid option for the Remote API:

https://docs.docker.com/engine/reference/api/docker_remote_api_v1.19/

Signed-off-by: Daniel Rice <drice304@gmail.com>


bfirsh:
IIRC, we should just pass no value by default, because Swarm often uses the overlay driver by default, and this would override that default.

shin-:
@bfirsh Yes, that's exactly what the `default` value means "bridge for engine, overlay for swarm". See discussion in #986 .

I'll close this as a result.

PR #988, opened 2016-03-11
Fix enable_tls issue

### LINKED ISSUE #984: Certificate error in docker ci for test-docker-py
https://jenkins.dockerproject.org/job/Docker-PRs/24848/console for detail.
in docker-py, when checkout to the commit of 387db11009f4b4f64a4f2c6fd64d3eeb01828585,the error appears,if I remove the commit ,we will not have the error.
```

==================================== ERRORS ====================================
_________________ ERROR at setup of InformationTest.test_info __________________
/docker-py/tests/integration/conftest.py:17: in setup_test_session
    c = docker_client()
/docker-py/tests/helpers.py:61: in docker_client
    return docker.Client(**docker_client_kwargs(**kwargs))
/docker-py/tests/helpers.py:65: in docker_client_kwargs
    client_kwargs = docker.utils.kwargs_from_env(assert_hostname=False)
/docker-py/docker/utils/utils.py:486: in kwargs_from_env
    assert_fingerprint=tls_verify)
/docker-py/docker/tls.py:47: in __init__
    'Path to a certificate and key files must be provided'
E   TLSParameterError: Path to a certificate and key files must be provided through the client_config 
param. TLS configurations should map the Docker CLI client configurations. 
See https://docs.docker.com/engine/articles/https/ for API details.
________________ ERROR at setup of InformationTest.test_search _________________
...
```

shin-:
Fixes #984 (?)

### DIFF
--- a/docker/utils/utils.py
+++ b/docker/utils/utils.py
@@ -460,16 +460,16 @@ def kwargs_from_env(ssl_version=None, assert_hostname=None):
     tls_verify = os.environ.get('DOCKER_TLS_VERIFY')
     if tls_verify == '':
         tls_verify = False
-        enable_tls = True
     else:
         tls_verify = tls_verify is not None
-        enable_tls = cert_path or tls_verify
+    enable_tls = cert_path or tls_verify
 
     params = {}
 
     if host:
-        params['base_url'] = (host.replace('tcp://', 'https://')
-                              if enable_tls else host)
+        params['base_url'] = (
+            host.replace('tcp://', 'https://') if enable_tls else host
+        )
 
     if not enable_tls:
         return params
--- a/tests/unit/utils_test.py
+++ b/tests/unit/utils_test.py
@@ -228,19 +228,7 @@ class KwargsFromEnvTest(base.BaseTestCase):
                           DOCKER_TLS_VERIFY='')
         os.environ.pop('DOCKER_CERT_PATH', None)
         kwargs = kwargs_from_env(assert_hostname=True)
-        self.assertEqual('https://192.168.59.103:2376', kwargs['base_url'])
-        self.assertTrue('ca.pem' in kwargs['tls'].ca_cert)
-        self.assertTrue('cert.pem' in kwargs['tls'].cert[0])
-        self.assertTrue('key.pem' in kwargs['tls'].cert[1])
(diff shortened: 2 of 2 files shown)


keloyang:
I have a try and https://github.com/docker/docker-py/issues/984 is fixed.thank you.

tests/unit/utils_test.py
```diff
@@ -228,19 +228,7 @@ def test_kwargs_from_env_tls_verify_false_no_cert(self):
...
-        self.assertEqual(True, kwargs['tls'].assert_hostname)
-        self.assertEqual(False, kwargs['tls'].verify)
-        try:
-            client = Client(**kwargs)
-            self.assertEqual(kwargs['base_url'], client.base_url)
-            self.assertEqual(kwargs['tls'].cert, client.cert)
-            self.assertFalse(kwargs['tls'].verify)
-        except TypeError as e:
-            self.fail(e)
+        self.assertEqual('tcp://192.168.59.103:2376', kwargs['base_url'])
```
dnephin: From this test change it's not clear to me what the expected behaviour is supposed to be.

It seems like we're reverting part of the recent change, is that right?

docker/utils/utils.py
```diff
@@ -460,16 +460,16 @@ def kwargs_from_env(ssl_version=None, assert_hostname=None):
     tls_verify = os.environ.get('DOCKER_TLS_VERIFY')
     if tls_verify == '':
         tls_verify = False
-        enable_tls = True
     else:
         tls_verify = tls_verify is not None
-        enable_tls = cert_path or tls_verify
+    enable_tls = cert_path or tls_verify
```
dnephin: So in the case where `DOCKER_TLS_VERIFY=` we're changing from tls being enabled by default, to tls being disabled by default?

tests/unit/utils_test.py
shin-: Yes. I tested that case locally with the docker CLI, and setting DOCKER_TLS_VERIFY to empty string and DOCKER_HOST to a TCP address let me connect to an unsecured daemon, so I believe this change is correct.

docker/utils/utils.py
shin-: Yes - this in accordance with what people reported in #984.

dnephin:
LGTM

Issue #978, opened 2016-03-06
Cryptic error message using not allowed tag


uggla:
Hello,

Below is the error message received trying to build a Dockerfile and setting a tag with uppercase (which is not allowed)

I think it could be clearer.

Error msg:
```
[uggla@ugglalaptop docker-py_debug]$ python dockerbuild.py 
Traceback (most recent call last):
  File "dockerbuild.py", line 10, in <module>
    rm=True)]
  File "/usr/lib/python2.7/site-packages/docker/client.py", line 380, in build
    timeout=timeout,
  File "/usr/lib/python2.7/site-packages/docker/client.py", line 100, in _post
    return self.post(url, **self._set_request_timeout(kwargs))
  File "/usr/lib/python2.7/site-packages/requests/sessions.py", line 511, in post
    return self.request('POST', url, data=data, json=json, **kwargs)
  File "/usr/lib/python2.7/site-packages/requests/sessions.py", line 468, in request
    resp = self.send(prep, **send_kwargs)
  File "/usr/lib/python2.7/site-packages/requests/sessions.py", line 576, in send
    r = adapter.send(request, **kwargs)
  File "/usr/lib/python2.7/site-packages/requests/adapters.py", line 426, in send
    raise ConnectionError(err, request=request)
requests.exceptions.ConnectionError: ('Connection aborted.', error(104, 'Connection reset by peer'))
```

Code :
```python
# coding=utf-8
from docker import Client

cli = Client(base_url='unix://var/run/docker.sock')


response = [line for line in cli.build(
    path='.',
    tag='ESSAI',
    rm=True)]

print(response)
```

aldenpeterson-wf:
It would also be helpful if you could simply print the full cli.build() results.

Instead of your above, it would be useful if it could work like:

print cli.build(
    path='.',
    tag='ESSAI',
    rm=True)

It is... not clear at all you need to read _each_ line of the build output (your example is the only example I saw of this) in order to see your full `docker build` output.

When I try just a print on the base cli.build() command the following obscure output is printed, since it does not return the build text:

    <generator object _stream_helper at 0x10fa295a0>

shin-:
Reported issue in https://github.com/docker/docker/issues/21069

We could do client-side validation to mitigate the issue, but the server closing the connection like that is definitely a bug.

[...]

PR #989, opened 2016-03-14
Client side validation of build tag

uggla:
Improve reporting of invalid tag error reported in #978.
Of course that will be better to have the server reporting
the appropriate error itself.
As mentioned by @shin-, this is a temporary mitigation.


shin-:
Thanks! This won't be needed anymore with docker engine 1.11, so I'm going to close this.

PR #995, opened 2016-03-17
Update default API version to 1.22

shin-:
Update engine version used for integration tests
IP address configuration tests needed tweaking


dnephin:
LGTM

PR #999, opened 2016-03-17
If TCP host is provided while TLS is enabled, convert to https

shin-:
Fixes the `tcp://` protocol issue mentioned in docker/compose#3139


aanand:
LGTM

PR #1004, opened 2016-03-21
makes it possible to have '=' in the env file

klj613:
note: the docker command like --env-file also allows '=' in the env file


dnephin:
Thanks, I think this is a necessary fix. The commit just needs to be signed.

shin-:
Thank you!

PR #1008, opened 2016-03-23
Use gzip to compress tar balls before sending

shin-:
Carries #930


dnephin:
LGTM, test failure seems unrelated?

PR #1011, opened 2016-03-23
Specify default value of stop's timeout

tomleb:
I was using docker-py for a project and I needed to know the default value of the timeout option for the stop function of the docker api (in docker-py/api/container.py). It wasn't specified in the documentation so I decided to add it myself because other like me might need it.

Signed-off-by: Tom Lebreux <tomlebreux@hotmail.com>


shin-:
LGTM, thank you!

dnephin:
LGTM

PR #1015, opened 2016-03-30
Improve port binding docs

### LINKED ISSUE #1014: Mistake in Port_bindings documentation?
In your port_bindings documentation it says: 

```
Similarly for several container ports bound to a single host port:

cli.create_host_config(port_bindings={
    1111: [1234, 4567]
})
```

Isn't it the other way around? Several host ports bound to a single container port?

aanand:
- Reword a sentence that was wrong ("container" and "host" were the wrong way round)
- Put the multiple-ports example before the multiple-IPs example

Closes #1014.

### DIFF
--- a/docs/port-bindings.md
+++ b/docs/port-bindings.md
@@ -38,19 +38,21 @@ container_id = cli.create_container(
 )
 ```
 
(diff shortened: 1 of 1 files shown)


shin-:
LGTM!

Issue #1016, opened 2016-03-30
Add other filters to ClientAPI.containers documentation


SakuraSound:
I am happy to know through my own testing that the filter tag "ancestor" works for the containers function.  If the other filters work as well, it may be good to add them to the "Available Filters" list in the documentation, so that others know that it works. If this is something that I can do, please point me in the right direction.

Example use:  `c = cli.containers(filters={"ancestor": "custom/redis"})` properly returns those containers that spawned from the custom/redis image.

SakuraSound:
I issued a pr for updating the documentation. Am I missing something. I want to help at least clean up the docs on this project

bfirsh:
Thanks for the contribution! This should be live now.

PR #1017, opened 2016-03-30
Added extra filtering options to containers command in ClientAPI

SakuraSound:
From #1016, I added some extra documentation notes for the containers command in ClientAPI.
Signed-off-by: Joir-dan Gumbs <jdagumbs@gmail.com>

### DIFF
--- a/docker/version.py
+++ b/docker/version.py
@@ -1,2 +1,2 @@
-version = "1.8.0-rc3"
+version = "1.8.0-rc4"
 version_info = tuple([int(d) for d in version.split("-")[0].split(".")])
--- a/docs/api.md
+++ b/docs/api.md
@@ -148,6 +148,11 @@ non-running ones
     - `exited` (int): Only containers with specified exit code
     - `status` (str): One of `restarting`, `running`, `paused`, `exited`
     - `label` (str): format either `"key"` or `"key=value"`
+    - `id` (str): The id of the container.
+    - `name` (str): The name of the container.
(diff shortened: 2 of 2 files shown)


shin-:
Thanks! Do you mind rebasing this against the current `master`? Changes LGTM.

SakuraSound:
Done

MHBauer:
Looks more comprehensive. I see a filter for `volume` and `isolation` in the docker docs. Given windows isn't prime time yet, I think you could leave `isolation` off, but I'd like to see `volume` included.

shin-:
Thanks!
If we expect this list will continue to grow, maybe a better solution would be to link to the source instead of trying to play catch up. Or at least add a disclaimer like "`A comprehensive list is available [here](link-docker-docs)`"

SakuraSound:
I could probably add that at the bottom of the updated docs, and it should work as long as the link carries with each new version.

shin-:
Ah, yeah, fair point. I don't think the docker docs have a dynamic link to the latest API docs page. Maybe we can just link to https://docs.docker.com/engine/reference/api/docker_remote_api/ for now.

SakuraSound:
Looking at the docs link, it seems we may be able to get away with using links:

https://docs.docker.com/engine/reference/commandline/ps/ seems like it carries across versions.

shin-:
Agreed. Do you mind making the change?

SakuraSound:
Can do

shin-:
Awesome, thank you!

PR #1020, opened 2016-04-05
Remove not needed dot in README.md

mschnitzer:
I fixed just a little mistake in the README.md. Nothing special here.


shin-:
Thanks!

PR #1022, opened 2016-04-05
Don't raise InvalidConfigError when auth dict doesn't have an 'auth' key

### LINKED ISSUE #1021: Empty auth dictionary should be valid
docker/compose#3265

shin-:
Fixes #1021


dnephin:
LGTM when build is green. Not sure why they all failed, maybe just a jenkins issue?

PR #1028, opened 2016-04-07
Note explicit support for Python 3.5

felixonmars:


TomasTomecek:
dependency updates look like a pretty big change; some of them even jump several minor releases -- it could be very beneficial to run compose's test suite with the newer deps

felixonmars:
I can fix the flake8 errors here by adding some noqa, and will run compose's test suite soon.

Do you have any idea for the `E   docker.errors.APIError: 500 Server Error: Internal Server Error ("b'Invalid signal: Signals.SIGKILL'")` failure in janky? I tried a bit but it's still strange :(

graingert:
fixed in https://github.com/docker/docker-py/pull/1071

graingert:
@felixonmars with a better approach to the flake8 errors

felixonmars:
Thank you :)

shin-:
Thank you for your help! This has been fixed by #1071, and I will close this issue as a result.

PR #1029, opened 2016-04-07
Use backports.ssl_match_hostname

felixonmars:
The py2-ipaddress module unfortunately conflicts with the pypi:ipaddress module, which is in the dependency tree of widely used pyOpenSSL. I think it would be a good idea to use a well maintained backport of the Python 3.5 implementation of match_hostname() instead of duplicating the effort and maintain another.

All tests are passing here.

Signed-off-by: Felix Yan <felixonmars@archlinux.org>


felixonmars:
Hmm, this is really weird. The IP Address test passes here with Python 2.7.11, and also passes on the Jenkins build, but failed on Travis.

felixonmars:
Oops, nevermind. Turns out the `ipaddress` module is also needed. Updated and all should work fine now!

felixonmars:
Looks like the Jenkins failure is not related to this PR.

setup.py
```diff
@@ -13,7 +13,8 @@
 ]
 
 extras_require = {
-    ':python_version < "3"': 'py2-ipaddress >= 3.4.1',
+    ':python_version < "3.5"': 'backports.ssl_match_hostname >= 3.5',
+    ':python_version < "3.5"': 'ipaddress',
```
shin-: I think this is only needed for `python < 3.3` ( https://docs.python.org/3/library/ipaddress.html )

shin-:
Thanks! I didn't know this had been backported already. Obviously it is much better than having to maintain it ourselves.

Just one small issue with the requirements, otherwise this is good.

setup.py
felixonmars: According to https://github.com/felixonmars/docker-py/blob/backports/docker/ssladapter/ssladapter.py#L18 and https://bugs.python.org/issue23239, the IP address support wasn't backported to Python 3.3 and 3.4 (yet?), so when I use `< 3.3` the tests failed.

TomasTomecek:
This is awesome! I'm currently struggling with docker-py 1.8 in Fedora world since we don't have `py2-ipaddress` packaged.

setup.py
shin-: `ipaddress` is available from python 3.3 onwards. Change in #1039 is what I meant and seems to be working fine.

felixonmars: Ah, I see. Thanks :)

PR #1033, opened 2016-04-11
Make ipaddress fully optional

dims:
* Fail only when ipaddress is actually used.
* Move the py2-ipaddress reference to test-requirements since we
  are already marking it as extras in setup.py

NOTE: there's at least a couple of implementations
- https://bitbucket.org/kwi/py2-ipaddress
- https://github.com/phihag/ipaddress
So, we should be okay if either one is present. Of course, we
would test and use py2-ipaddress as our preferred impl. So
we specify py2-ipaddress in our test-requirements.txt.

### DIFF
--- a/docker/ssladapter/ssl_match_hostname.py
+++ b/docker/ssladapter/ssl_match_hostname.py
@@ -3,8 +3,12 @@
 # Changed to make code python 2.x compatible (unicode strings for ip_address
 # and 3.5-specific var assignment syntax)
 
-import ipaddress
 import re
+
+try:
+    import ipaddress as _ipaddress
+except ImportError:
+    _ipaddress = None
 
 try:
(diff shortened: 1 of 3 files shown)


dims:
@GordonTheTurtle thanks! done.

Also, fyi, here is some context about the problem we encountered in OpenStack:
http://markmail.org/message/sbjqgzcj7tjcfulg

shin-:
Does #1039 solve your problem? It swaps the dependency to `ipaddress`, which seems to be the one you are using.

I absolutely want to avoid requiring the user to perform manual steps before using the library if at all possible, so I don't think this PR is the right approach.

dims:
@shin- yep, switching to ipaddress works for us. thanks!

PR #1053, opened 2016-05-03
Add support for DNS Search to create container

romaimperator:
Currently, only the client.start api accepts the dns_search parameter. This commit makes the client.create_container api method also accept dns_search. The Ansible core module for docker uses client.create_container so this is necessary to support the dns_search argument over there.


romaimperator:
After reviewing my test with the Ansible module, it appears that passing the dns search param works with the existing version 1.8.1 of docker-py already so this pull request isn't required for the ansible module. I'm not sure if this would still be useful anyways though.

shin-:
`dns_search` is part of [HostConfig](http://docker-py.readthedocs.io/en/stable/hostconfig/)

PR #1058, opened 2016-05-06
Globally set default API to auto

ubante:
In Ansible modules, we're setting docker_api_version to auto.  Instead of doing that to all the modules, we can make the change here.

Understandably, this may not be appropriate for the Universe.


shin-:
Thanks! See my comment here: https://github.com/docker/docker-py/pull/505#issuecomment-77285993

The best experience using docker-py for users will always be to use the version it's tested and verified against. `auto` is provided as an option, but should not be the default.

ubante:
Thanks for the quick review.

PR #1065, opened 2016-05-16
Resolve path traversal in .dockerignore patterns

aanand:
We were failing to match `.dockerignore` patterns such as `./a.py`, as reported in https://github.com/docker/compose/issues/1607#issuecomment-218765317. I've checked, and Engine does indeed support both `.` and `..` traversal in patterns.


dnephin:
LGTM

shin-:
Thanks!

PR #1071, opened 2016-05-19
support Python 3.5

graingert:


docker/utils/utils.py
```diff
@@ -546,12 +546,6 @@ def datetime_to_timestamp(dt):
     return delta.seconds + delta.days * 24 * 3600
 
 
-def longint(n):
```
graingert: all this does is allow users to enter their storage size in pounds (300lb) on python 2:

tests/unit/utils_test.py
```diff
@@ -299,56 +299,30 @@ def test_convert_volume_binds_no_mode(self):
         self.assertEqual(convert_volume_binds(data), ['/mnt/vol1:/data:rw'])
 
     def test_convert_volume_binds_unicode_bytes_input(self):
-        if six.PY2:
```
dnephin: I don't think we want to remove the tests for PY2. Having them in separate test cases with `pytest.mark.skip` would be fine.

graingert: They're still there. I just combined the 3K and 2 tests

dnephin: Oh, I missed that, sounds good.

graingert:
@dnephin any update on a merge?

shin-:
I thought I had commented on there, sorry. Changes LGTM.

dnephin:
LGTM

PR #1079, opened 2016-06-03
npipe support

### LINKED ISSUE #1024: Support windows named pipes for docker api
https://github.com/docker/compose/issues/3170

Apparently the pywin32 API provides some tools for working with this (http://docs.activestate.com/activepython/3.4/pywin32/win32pipe.html). Also available as `pypiwin32` on pypi. Examples: http://codereview.stackexchange.com/questions/88672/python-wrapper-for-windows-pipes

shin-:
Fixes #1024 

Still very much experimental, hasn't yet been formally tested. Feedback is welcome.


shin-:
cc @aanand @dnephin 

https://github.com/docker/compose/issues/3170

docker/transport/npipesocket.py
```diff
@@ -0,0 +1,194 @@
...
+            None,
+            win32file.OPEN_EXISTING,
+            cSECURITY_ANONYMOUS | cSECURITY_SQOS_PRESENT,
+            0
+        )
+        self.flags = win32pipe.GetNamedPipeInfo(handle)[0]
+        # self.state = win32pipe.GetNamedPipeHandleState(handle)[0]
+
+        # if self.state & cPIPE_READMODE_MESSAGE != 0:
+        #     raise RuntimeError("message readmode pipes not supported")
```
dnephin: some commented out code here

docker/transport/npipesocket.py
```diff
@@ -0,0 +1,194 @@
...
+            raise NotImplementedError()
+        rawio = NpipeFileIOBase(self)
+        if bufsize is None:
+            bufsize = io.DEFAULT_BUFFER_SIZE
+        return io.BufferedReader(rawio, buffer_size=bufsize)
+
+    @check_closed
+    def recv(self, bufsize, flags=0):
+        err, data = win32file.ReadFile(self._handle, bufsize)
+        return data
```
dnephin: err is not used here, should it raise an exception if there is an error?

docker/transport/npipesocket.py
```diff
@@ -0,0 +1,194 @@
...
+        err, data = win32file.ReadFile(
+            self._handle,
+            readbuf[:nbytes] if nbytes else readbuf
+        )
+        return len(data)
+
+    @check_closed
+    def send(self, string, flags=0):
+        err, nbytes = win32file.WriteFile(self._handle, string)
+        return nbytes
```
dnephin: same with this err

dnephin:
This looks great. Just a couple minor comments.

docker/transport/npipesocket.py
shin-: According to the documentation, 

> The result is a tuple of (`hr`, `string`/`PyOVERLAPPEDReadBuffer`), where `hr` may be 0, `ERROR_MORE_DATA` or `ERROR_IO_PENDING`. 

None of those are actually "errors". `MORE_DATA` is irrelevant in the context of a socket emulation, and `IO_PENDING` only occurs when we use an overlapped buffer, which we don't in our case.

Other (actual) errors will be raised as standard Python exceptions.

docker/transport/npipesocket.py
shin-: > The result is a tuple of (errCode, nBytesWritten). If errCode is not zero, it will be ERROR_IO_PENDING (ie, it is an overlapped request). 

Since we're not messing with overlapped structures, we don't need to handle this.

docker/transport/npipesocket.py
dnephin: cool sgtm

dnephin:
LGTM

PR #1082, opened 2016-06-07
Invoke self._result passing `json=True` when `decode=True`

masci:
The generator should yield a decoded object if the the `_stream_helper` was invoked with the `decode=True` option in the first place.

Added test case to address the specific issue.

### DIFF
--- a/docker/client.py
+++ b/docker/client.py
@@ -235,7 +235,7 @@ class Client(
         else:
             # Response isn't chunked, meaning we probably
             # encountered an error immediately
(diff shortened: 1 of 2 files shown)


shin-:
Thanks!

Hm, that's interesting. I don't know if we can or should count on the response being a JSON object in that situation. On the other hand, if it's truly an error case like we assume, it doesn't matter whether we pass the json parameter or not.

All in all, seems fairly safe. LGTM!

PR #1083, opened 2016-06-08
Support ipv4_address and ipv6_address in create_endpoint_config

### LINKED ISSUE #982: Support for net aliases
The API change to give network name aliases to containers was implemented here:
https://github.com/docker/docker/pull/19242

I cannot find this option exposed in create_container, should I look elsewhere?

### LINKED ISSUE #1042: --ip and --ip6 options
I can specify specific IPv4 or IPv6 addresses when running `docker run --ip 1.2.3.4` manually, but it's not possible to do the same via API (thus, with automation tools like ansible).

`docker.Client.create_container()` should accept optional `ip` and `ip6` parameters.

### LINKED ISSUE #1088: create_ipam_config() got an unexpected keyword argument 'subnet'
https://docker-py.readthedocs.io/en/latest/networks/ this tells me I can use subnet argument, but when I launch my program it throws `TypeError: create_ipam_config() got an unexpected keyword argument 'subnet'`

docker-py version is 1.8.1

shin-:
Also fleshing out networking docs

Fixes #1042 
Fixes #982 
Fixes #1088

### DIFF
--- a/docs/networks.md
+++ b/docs/networks.md
@@ -40,9 +40,12 @@ docker_client.create_network("network1", driver="bridge", internal=True)
 
 ## Container network configuration
 
-In order to specify which network(s) a container will be connected to and
+In order to specify which network a container will be connected to, and
 additional configuration, use the `networking_config` parameter in
-`Client.create_container`
+`Client.create_container`. Note that at the time of creation, you can
+only connect a container to a single network. Later on, you may create more
+connections using `Client.connect_container_to_network`.
+
 
 ```python
 networking_config = docker_client.create_networking_config({
--- a/docs/networks.md
+++ b/docs/networks.md
@@ -1,27 +1,171 @@
 # Using Networks
 
+## Network creation
+
 With the release of Docker 1.9 you can now manage custom networks.
 
 
-Here you can see how to create a network named ```network1``` using the ```bridge``` driver
+Here you can see how to create a network named `network1` using
+the `bridge` driver
 
 ```python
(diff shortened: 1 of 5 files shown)


shin-:
cc @aanand because networking stuff :cat:

docs/networks.md
```diff
@@ -1,27 +1,173 @@
...
 ```python
 
 docker_client.create_network("network1", driver="bridge", internal=True)
 ```
+
+## Container network configuration
+
+In order to specify which network(s) a container will be connected to and
+additional configuration, use the `networking_config` parameter in
+`Client.create_container`
```
aanand: This is a little misleading. When creating a container, you can only connect it to one network, using the `network_mode` argument to `create_host_config`. You can only specify that one network's config in the `networking_config` dict (so the example below is fine).

To connect a container to multiple networks, you have to first create it with one of the networks specified, then call `connect_container_to_network` once for each remaining network, with the relevant endpoint config. We should probably be explicit about this.

aanand:
Code changes look good. One comment about the docs.

We could also use docs for the other methods related to networks - `create_network`, `connect_container_to_network`, etc, but that can be done in a separate PR.

shin-:
Updated the docs to mention the 1 network limit.
I agree re: other methods, although those are already documented in `api.md`, so they're less of a priority. 
Moving forward, maybe separating the API docs by theme makes more sense given how the API has grown. Food for thought.

aanand:
Ah, I missed that. LGTM, just needs rebase.

Issue #1076, opened 2016-05-26
docker-py tries to resolves localunixsocket.local


dnephin:
From: https://github.com/docker/compose/issues/3419#issuecomment-221793401

On OSX (at least) docker-py seems to try to resolve the fake hostname that is used by the unix socket adapter. This is causing some latency for docker4mac users.

shin-:
Fixed by #1928

PR #1087, opened 2016-06-09
Remove default adapters when connecting through a unix socket.

shin-:
Related: #1076 
Kind of a shot in the dark. Can someone test this patch and report if they see any difference? ( @aanand @FrenchBen )


FrenchBen:
I believe the dependency was in compose. Probably better use-case to have @aanand test it.

PR #1104, opened 2016-06-24
Add missing create_network params

ColinHuang:
There some parameters in the API but not in docs.

```
 def create_network(self, name, driver=None, options=None, ipam=None, check_duplicate=None, internal=False):
```

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -284,9 +284,12 @@ Create a network, similar to the `docker network create` command.
 **Params**:
 
 * name (str): Name of the network
-* driver (str): Name of the driver used to create the network
-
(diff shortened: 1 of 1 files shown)


docs/api.md
```diff
@@ -284,9 +284,12 @@ Create a network, similar to the `docker network create` command.
 **Params**:
 
 * name (str): Name of the network
-* driver (str): Name of the driver used to create the network
-
+* driver (str): Name of the network driver plugin to use. Defaults to bridge driver
```
shin-: This is incorrect - it defaults to `bridge` for regular Docker Engine, but the default for Swarm is `overlay`. Please remove the second part.

PR #1109, opened 2016-06-28
signal in Client.kill can be a string containing the signal's name

shin-:
Fixes issue introduced in #1071


dnephin:
Looks like the integration test is unhappy.

shin-:
@dnephin Should look good now.

aanand:
LGTM

PR #1113, opened 2016-06-30
Added support for user namespace.

srikalyan:
Please let me know if you need any changes.

### DIFF
--- a/docker/utils/utils.py
+++ b/docker/utils/utils.py
@@ -620,7 +620,7 @@ def create_host_config(binds=None, port_bindings=None, lxc_conf=None,
                        device_write_bps=None, device_read_iops=None,
                        device_write_iops=None, oom_kill_disable=False,
                        shm_size=None, version=None, tmpfs=None,
-                       oom_score_adj=None):
+                       oom_score_adj=None, userns_mode=None):
 
     host_config = {}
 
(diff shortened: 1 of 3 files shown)


srikalyan:
@GordonTheTurtle Done.

docker/api/container.py
```diff
@@ -305,7 +305,7 @@ def start(self, container, binds=None, port_bindings=None, lxc_conf=None,
               dns=None, dns_search=None, volumes_from=None, network_mode=None,
               restart_policy=None, cap_add=None, cap_drop=None, devices=None,
               extra_hosts=None, read_only=None, pid_mode=None, ipc_mode=None,
-              security_opt=None, ulimits=None):
+              security_opt=None, ulimits=None, user_namespace=None):
```
shin-: We are not adding new parameters to `start()` - anything but the container ID is deprecated anyways.

docker/utils/utils.py
```diff
@@ -620,7 +620,7 @@ def create_host_config(binds=None, port_bindings=None, lxc_conf=None,
                        device_write_bps=None, device_read_iops=None,
                        device_write_iops=None, oom_kill_disable=False,
                        shm_size=None, version=None, tmpfs=None,
-                       oom_score_adj=None):
+                       oom_score_adj=None, user_namespace=None):
```
shin-: Let's stick close to the engine's names and use `userns_mode` instead.

shin-:

I have left a couple comments on the code, please take the time to address those when you can.

Also, @GordonTheTurtle is a bot, so he won't be able to reply to you :)

docker/utils/utils.py
srikalyan: Done.

srikalyan:
@shin- ,Thank you for the quick review. I made the changes you have suggested. @GordonTheTurtle seems not be an intelligent bot to reply to my questions :p.

shin-:
Great, thanks! One last thing before going forward is to add the parameter to the docs, specifically in `docs/hostconfig.md` - and also while you are at it, please squash your commits! :)

srikalyan:
@shin-, added doc and squashed commit. Please let me know if you need anything else.

shin-:
Thanks, LGTM!

PR #1119, opened 2016-07-06
Pass X-Registry-Auth when building an image

jmichalicek:
This fixes a bug where X-Registry-Auth was not set because the `headers` variable in BuildApiMixin.build() did not get set properly by _set_auth_headers() if `headers` was `None` when `set_auth_headers()` got called.  This resulted in Dockerfiles with `FROM <private_registry>` to fail to authenticate.

* Initialize headers variable in BuildApiMixin.build() as a dict rather than as None.  This way the correct object gets passed to _set_auth_headers() even if no headers were set in build()

* Changing object from None to {} in BuildApiMixin._set_auth_headers() removed because it changes the object reference, so has no effect on calling code.

Signed-off-by: Justin Michalicek <jmichalicek@gmail.com>

### DIFF
--- a/tests/unit/build_test.py
+++ b/tests/unit/build_test.py
@@ -2,8 +2,9 @@ import gzip
 import io
 
 import docker
+from docker import auth
 
-from .api_test import DockerClientTest
+from .api_test import DockerClientTest, fake_request, url_prefix
 
 
 class BuildTest(DockerClientTest):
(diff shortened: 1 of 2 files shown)


shin-:
Thanks!

Changes make sense - would it be possible to add a test to validate it and make sure we don't break it in the future?

jmichalicek:
I'd love to.  I looked into it a bit when writing this, but didn't immediately see a clean way to do it with where the current mocks occur.  Let me take another look at it today with my head clear and brain fresh, though.  I probably just missed something obvious.

jmichalicek:
There we go, pretty obvious coming at it fresh.  Test cases are in place.

shin-:
Thanks, tests look good!

LGTM - cc @aanand @dnephin

tests/unit/build_test.py
```diff
@@ -103,3 +121,44 @@ def test_build_container_invalid_container_limits(self):
                 'foo': 'bar'
             })
         )
+
+    def test__set_auth_headers_with_empty_dict_and_auth_configs(self):
```
aanand: Small thing, but there's a double underscore after `test`, which we don't do anywhere else. Same thing with the other new tests.

aanand:
Thanks! Minor style nit and needs a rebase, otherwise fine.

tests/unit/build_test.py
jmichalicek: Sorry about that, that was just keeping the name exact with the method's leading underscore.  I'll change it first thing in the morning, rebase, and push it up.

aanand:
LGTM

PR #1120, opened 2016-07-07
Update api.md

Faylixe:
Just added an example for the tag method as it took me some time to understand the parameter usage :)


bfirsh:
replaced by #1316

PR #1124, opened 2016-07-08
Send connection upgrade headers

### LINKED ISSUE #1123: The api client should send connection upgrade headers

>

```
Upgrade: tcp
Connection: Upgrade
```

dgageot:
Fixes #1123

I'm all new to this codebase, so I'm not sure I did the things right.

I basically looked at [docker cli code](https://github.com/docker/docker/blob/07f580489908bf6a3373daac1473045406e1130d/vendor/src/github.com/docker/engine-api/client/hijack.go#L34) and found out where the `postHijacked` function is called:
 + `/attach` api calls https://github.com/docker/docker/blob/7fd53f7c711474791ce4292326e0b1dc7d4d6b0f/vendor/src/github.com/docker/engine-api/client/container_attach.go#L33
 + attached `/exec/ calls https://github.com/docker/docker/blob/7fd53f7c711474791ce4292326e0b1dc7d4d6b0f/vendor/src/github.com/docker/engine-api/client/container_exec.go#L33 and https://github.com/docker/docker/blob/6b4a46f28266031ce1a1315f17fb69113a06efe1/api/client/exec.go#L90

I tried to add a unit test for `attach` but couldn't find any to get inspiration from.

### DIFF
--- a/docker/api/container.py
+++ b/docker/api/container.py
@@ -15,10 +15,16 @@ class ContainerApiMixin(object):
             'logs': logs and 1 or 0,
             'stdout': stdout and 1 or 0,
             'stderr': stderr and 1 or 0,
-            'stream': stream and 1 or 0,
+            'stream': stream and 1 or 0
         }
+
+        headers = {
+            'Connection': 'Upgrade',
+            'Upgrade': 'tcp'
+        }
+
         u = self._url("/containers/{0}/attach", container)
-        response = self._post(u, params=params, stream=stream)
+        response = self._post(u, headers=headers, params=params, stream=stream)
 
         return self._get_result(container, stream, response)
 
--- a/docker/api/exec_api.py
+++ b/docker/api/exec_api.py
@@ -66,8 +66,16 @@ class ExecApiMixin(object):
             'Detach': detach
         }
 
+        headers = {} if detach else {
+            'Connection': 'Upgrade',
+            'Upgrade': 'tcp'
+        }
+
         res = self._post_json(
-            self._url('/exec/{0}/start', exec_id), data=data, stream=stream
+            self._url('/exec/{0}/start', exec_id),
+            headers=headers,
+            data=data,
+            stream=stream
         )
 
         if socket:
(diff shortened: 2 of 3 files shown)


bfirsh:
Looks like this is sending it as part of the request payload rather than as header. You probably want to pass it as a `headers` argument to `_post_json`.

dgageot:
@bfirsh sure...

dgageot:
@bfirsh should be better now

shin-:
I suppose we need some client-side code to handle the change of protocol?

The `Connection Aborted` issue only happens when we're executing the tests from inside a container, this is what I get running locally.

```
$ py.test tests/integration/exec_test.py 
============================= test session starts ==============================
platform linux2 -- Python 2.7.9 -- py-1.4.30 -- pytest-2.7.2
rootdir: /home/joffrey/work/pydocker, inifile: pytest.ini
plugins: cov
collected 7 items 

tests/integration/exec_test.py FFFF..F

______________________ ExecTest.test_exec_command_as_root ______________________
tests/integration/exec_test.py:74: in test_exec_command_as_root
    self.assertEqual(exec_log, b'root\n')
E   AssertionError: '' != 'root\n'
______________________ ExecTest.test_exec_command_as_user ______________________
tests/integration/exec_test.py:58: in test_exec_command_as_user
    self.assertEqual(exec_log, b'default\n')
E   AssertionError: '' != 'default\n'
_____________________ ExecTest.test_exec_command_streaming _____________________
tests/integration/exec_test.py:92: in test_exec_command_streaming
    self.assertEqual(res, b'hello\nworld\n')
E   AssertionError: '' != 'hello\nworld\n'
______________________ ExecTest.test_exec_command_string _______________________
tests/integration/exec_test.py:42: in test_exec_command_string
    self.assertEqual(exec_log, b'hello world\n')
E   AssertionError: '' != 'hello world\n'
________________________ ExecTest.test_execute_command _________________________
tests/integration/exec_test.py:26: in test_execute_command
    self.assertEqual(s, b'hello\n')
E   AssertionError: '' != 'hello\n'
```

dgageot:
Closing in favor of https://github.com/docker/docker-py/pull/1126

PR #1125, opened 2016-07-08
Set custom user agent on client

bfirsh:
See https://github.com/docker/compose/issues/3550

### DIFF
--- a/docker/constants.py
+++ b/docker/constants.py
@@ -1,4 +1,5 @@
 import sys
+from .version import version
 
 DEFAULT_DOCKER_API_VERSION = '1.22'
 DEFAULT_TIMEOUT_SECONDS = 60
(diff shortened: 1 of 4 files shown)


docker/constants.py
```diff
@@ -12,3 +13,5 @@
     'is deprecated and non-functional. Please remove it.'
 
 IS_WINDOWS_PLATFORM = (sys.platform == 'win32')
+
+DEFAULT_USER_AGENT = "docker-py/%s" % version
```
shin-: Let's use `format` instead to be consistent with the rest of the codebase!

bfirsh: Fixed :)

dnephin:
LGTM

bfirsh:
Rebased. ping @shin- @aanand

aanand:
LGTM

PR #1126, opened 2016-07-08
Support for TCP upgrade 

### LINKED ISSUE #1123: The api client should send connection upgrade headers

shin-:
Carry #1124 
Fixes #1123

### DIFF
--- a/tests/helpers.py
+++ b/tests/helpers.py
@@ -54,7 +54,7 @@ def exec_driver_is_native():
         c = docker_client()
         EXEC_DRIVER = c.info()['ExecutionDriver']
         c.close()
-    return EXEC_DRIVER.startswith('native')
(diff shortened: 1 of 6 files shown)


tests/helpers.py
```diff
@@ -105,7 +105,7 @@ def read_data(socket, packet_size):
     while len(data) < packet_size:
         next_data = read_socket(socket, packet_size - len(data))
         if not next_data:
-            assert False, "Failed trying to read in the dataz"
+            assert False, "Failed trying to read in the data"
```
aanand: Rather than a conditional followed by `assert False`, can't this just read `assert next_data, "Failed trying to read in the data"`?

aanand:
Closing in favour of #1130

Issue #1085, opened 2016-06-08
Add Swarm management


bfirsh:
[In Docker 1.12, Swarm management is being added to the Remote API.](https://github.com/docker/docker/pull/23362) We should support that in docker-py too.

[I made a half-arsed start in case anyone wants to carry it and work on it.](https://github.com/bfirsh/docker-py/commit/8a0416f40808f01e05a55d7f34d195467e8e77c7)

ghostsquad:
@bfirsh - I saw your demo at DockerCon, pretty cool. I'd love to help carry the torch. I'm working on a hack for the Hackathon inspired by your work.

bfirsh:
@ghostsquad Feel free! I'm not working on it at the moment. :)

Tuxity:
Any ETA for that awesome feature ?
I saw the milestone 1.10.0 but don't know if it's A question of days, weeks or months

shin-:
Sorry, no ETA for 1.10.0 yet.

#1127 and #1147 have been merged to master, so I'm closing this.

shin-:
1.10.0 now available on pypi.

PR #1127, opened 2016-07-08
Service support implementation

shin-:
Services and tasks.

See: #1085


AkihiroSuda:
Seems need to apply some fixes:
e.g.
https://github.com/docker/engine-api/commit/14c70cc34bfee8f1335137b77cc8d496d0169bf8
https://github.com/docker/engine-api/commit/2d2ff7c7d6ee8bc7408e7c2440a4da1cf309c198

pacoxu:
Any update on this? 

we get error during update service
{"message":"Invalid service version '': strconv.ParseUint: parsing \"\": invalid syntax"}
as https://forums.docker.com/t/docker-swarm-scale-service-using-update-api/19589

atakada:
Is there anyone still working on it? I found some parameter names to be updated, but I'll be happy if there are any branches I can try.
https://github.com/atakada/docker-py/tree/service-support-fix

pacoxu:
@atakada 
Finally, we found that you have to use the update API like
POST /services/<service_id>/update?version=<current-version-index>
{
 "Name": "**"
"TaskTemplate" : {***}
}

Hope it can help you.

shin-:
Thank you for your help, folks! I've been focused on other tasks lately but I should be able to pick this back up soon.

atakada:
Thanks for comments&updates! Unfortunately I'm working for other stuff, but will come back.

shin-:
- I have rebased this on top of #1147 so I could use the swarm init methods in the integration tests.
- Updated methods and added the missing `update_service`
- Moved the `TaskTemplate`, `ContainerSpec` and similar classes to the `docker.types` package.
- Started writing docs. API docs still need to be filled out.

At that point everything should work properly, but please leave a comment if you're testing this and running into issues. :)

tests/integration/service_test.py
```diff
@@ -0,0 +1,190 @@
+import random
+
+import docker
+# import pytest
```
dnephin: commented out import

docker/types/services.py
```diff
@@ -0,0 +1,181 @@
...
+                self['VolumeOptions'] = volume_opts
+            if propagation:
+                raise errors.DockerError(
+                    'Mount type is volume but `propagation` argument has been '
+                    'provided.'
+                )
+
+    @classmethod
+    def parse_mount_string(cls, string):
+        parts = string.split(':')
```
dnephin: The CLI doesn't support this syntax for mounts, so I'm not sure it should be supported here

dnephin:
LGTM

setup.py
```diff
@@ -9,7 +9,7 @@
 SOURCE_DIR = os.path.join(ROOT_DIR)
 
 requirements = [
-    'requests >= 2.5.2',
+    'requests >= 2.5.2, < 2.11',
```
shin-: https://github.com/docker/docker-py/pull/1204

glensc: hmm, #1204 just removes the pin. so it's actually ok to use requests-2.11 without any other code changes?

shin-: You can use any version >= 2.5.2, except 2.11.0. 
2.11.1 fixed the bug so that version or above is fine as well.

glensc: thanks for clarification!

PR #1130, opened 2016-07-13
Support for TCP upgrade

### LINKED ISSUE #1123: The api client should send connection upgrade headers

aanand:
- Carried @dgageot's changes from #1124 
- Carried @shin-'s changes from #1126 
- Refactored common code between `docker/client.py` and `tests/helpers.py` into `docker/utils/socket.py`
- Cleaned up and documented socket methods

Fixes #1123 

ping @dnephin

### DIFF
--- a/docker/api/exec_api.py
+++ b/docker/api/exec_api.py
@@ -66,8 +66,16 @@ class ExecApiMixin(object):
             'Detach': detach
         }
 
+        headers = {} if detach else {
+            'Connection': 'Upgrade',
+            'Upgrade': 'tcp'
+        }
+
         res = self._post_json(
-            self._url('/exec/{0}/start', exec_id), data=data, stream=stream
+            self._url('/exec/{0}/start', exec_id),
+            headers=headers,
+            data=data,
+            stream=stream
         )
 
         if socket:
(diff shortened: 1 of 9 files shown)


docker/api/exec_api.py
```diff
@@ -66,10 +64,18 @@ def exec_start(self, exec_id, detach=False, tty=False, stream=False,
...
+            'Connection': 'Upgrade',
+            'Upgrade': 'tcp'
+        }
+
         res = self._post_json(
-            self._url('/exec/{0}/start', exec_id), data=data, stream=stream
+            self._url('/exec/{0}/start', exec_id),
+            headers=headers,
+            data=data,
+            stream=True
```
dnephin: `stream` changed here to always `True`. Is that correct?

docker/utils/socket.py
```diff
@@ -0,0 +1,68 @@
...
+    return data
+
+
+def next_frame_size(socket):
+    """
+    Returns the size of the next frame of data waiting to be read from socket,
+    according to the protocol defined here:
+
+    https://docs.docker.com/engine/reference/api/docker_remote_api_v1.24/#/attach-to-a-container
+    """
```
dnephin: Well that's fun

dnephin:
code LGTM, waiting on janky

docker/api/exec_api.py
shin-: Late answer, but yes, that's intentional. Without it, `_read_from_socket` cannot work properly. As far as the user is concerned, the output will still be presented according to their choice.

Issue #1105, opened 2016-06-24
Support create network EnableIPv6 and Labels options 


linlinlinlin:
Check the remote API:
https://docs.docker.com/engine/reference/api/docker_remote_api_v1.23/#create-a-network

There are two missing JSON parameters:
```
EnableIPv6 - Enable IPv6 on the network
Labels - Labels to set on the network, specified as a map: {"key":"value" [,"key2":"value2"]}
```

linlinlinlin:
Found another missing parameters:
In `disconnect_container_from_network` API.
```
Force - Force the container to disconnect from a network
```

PR #1132, opened 2016-07-15
Add missing network API options

t-tran:
Add support to some options in these API functions:

 * create_network: EnableIPv6, Labels
 * disconnect_container_from_network: Force

Patch for issue #1105


shin-:
We went with #1178 instead, so I'll close this now that the feature is in master.

PR #1137, opened 2016-07-25
Default to npipe address on Windows

aanand:
See https://github.com/docker/compose/issues/3701.

ping @dgageot

### DIFF
--- a/docker/utils/utils.py
+++ b/docker/utils/utils.py
@@ -36,6 +36,8 @@ from .types import Ulimit, LogConfig
 
 DEFAULT_HTTP_HOST = "127.0.0.1"
 DEFAULT_UNIX_SOCKET = "http+unix://var/run/docker.sock"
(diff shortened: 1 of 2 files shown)


shin-:
Do you think this is too early? The `npipe` code hasn't really been tested extensively yet. I'd wait for 1.10

docker/utils/utils.py
```diff
@@ -390,7 +390,7 @@ def parse_host(addr, is_win32=False, tls=False):
     path = ''
 
     if not addr and is_win32:
-        addr = '{0}:{1}'.format(DEFAULT_HTTP_HOST, 2375)
+        addr = 'npipe:////./pipe/docker_engine'
```
shin-: Maybe this should be a constant

aanand:
Makes sense to me. I'll move it to the 1.10 milestone.

shin-:
Cool, LGTM :+1:

PR #1139, opened 2016-07-25
Send LinkLocalIPs as part of IPAMConfig dictionary

shin-:
Previous PR didn't include a test. This is also being rectified here.


aanand:
LGTM

PR #1145, opened 2016-07-29
Add support for sysctl when creating container

### LINKED ISSUE #1144: Add 'sysctl' option support
PR docker/docker#19265 merged 'sysctl' support into docker Remote API as of v1.24, but there seems to be no way to use this feature with docker-py. Has anyone started work on such a feature? If not, I will create a PR for a patch that we're using internally.

The sysctl option expects a dictionary of ```{ <name>: <value> }``` (for example: ```{ "net.ipv4.ip_forward": "1" }```), where the types of both objects are of type string. My patch checks if <value> is not a string type, and if that's the case, it casts it using str(). I originally intended to do the casting in docker-compose (for which I have another patch), but now that I've looked closer at the API I believe it belongs in the API. I'm open to debate on that matter though.

jtakkala:
Implemented sysctl support for setting kernel parameters when creating a container.
Closes #1144

Signed-off-by: Jari Takkala <jtakkala@gmail.com>

### DIFF
--- a/docker/utils/utils.py
+++ b/docker/utils/utils.py
@@ -619,7 +619,7 @@ def create_host_config(binds=None, port_bindings=None, lxc_conf=None,
                        blkio_weight_device=None, device_read_bps=None,
                        device_write_bps=None, device_read_iops=None,
                        device_write_iops=None, oom_kill_disable=False,
-                       shm_size=None, version=None, tmpfs=None,
+                       shm_size=None, sysctls=None, version=None, tmpfs=None,
                        oom_score_adj=None):
 
     host_config = {}
@@ -724,6 +724,13 @@ def create_host_config(binds=None, port_bindings=None, lxc_conf=None,
             raise host_config_type_error('security_opt', security_opt, 'list')
 
         host_config['SecurityOpt'] = security_opt
+
+    if sysctls:
+        if not isinstance(sysctls, dict):
+            raise host_config_type_error('sysctls', sysctls, 'dict')
+        host_config['Sysctls'] = {}
+        for k, v in six.iteritems(sysctls):
+            host_config['Sysctls'][k] = six.text_type(v)
 
     if volumes_from is not None:
         if isinstance(volumes_from, six.string_types):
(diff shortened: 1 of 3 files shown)


docker/utils/utils.py
```diff
@@ -725,6 +725,15 @@ def create_host_config(binds=None, port_bindings=None, lxc_conf=None,
 
         host_config['SecurityOpt'] = security_opt
 
+    if sysctls:
+        if not isinstance(sysctls, dict):
+            raise host_config_type_error('sysctls', sysctls, 'dict')
+        host_config['Sysctls'] = {}
+        for k, v in six.iteritems(sysctls):
+            if not isinstance(v, six.string_types):
+                v = str(v)
```
aanand: I think we should use `six.text_type` here - `str` means different things in Python 2/3.

aanand: (we can also remove the `isinstance` conditional)

aanand:
Thanks! Just one small code thing.

jtakkala:
@aanand fixed and squashed. The Jenkins build is failing on disk space, but I ran unit-tests locally and all passed.

aanand:
retest this please

docker/api/container.py
```diff
@@ -320,7 +320,7 @@ def start(self, container, binds=None, port_bindings=None, lxc_conf=None,
               dns=None, dns_search=None, volumes_from=None, network_mode=None,
               restart_policy=None, cap_add=None, cap_drop=None, devices=None,
               extra_hosts=None, read_only=None, pid_mode=None, ipc_mode=None,
-              security_opt=None, ulimits=None):
+              security_opt=None, sysctls=None, ulimits=None):
```
shin-: We don't want to add any `start` arguments, as their use has been deprecated for over a year now. Do you mind removing those changes?

shin-:
Thanks! I have just one request regarding the addition to the `Client.start` method. Rest is LGTM.

jtakkala:
Ok, fixed. Hope it's good now!

shin-:
LGTM, thank you!

PR #1114, opened 2016-06-30
Added support for docker swarm api version 1.24.

minzhang28:
3 API are added
swarm_init()
swarm_leave()
swarm_join()

PR #1147, opened 2016-08-01
Swarm API support

shin-:
Based on #1114


shin-:
#1085

shin-:
cc @dnephin This should be ready for review. :smile:

PR #1152, opened 2016-08-04
Add support for passing healthcheck to create_container

madjam002:
I'm looking into adding healthcheck support to the Docker Compile yaml file.
This is an attempt at adding support for healthcheck into docker-py.

There are no tests yet so please don't merge - I'd just like to know whether I'm on the right lines and whether you'll consider merging this.

healthcheck should be passed into create_container like so:

```
{
  interval: "5s",
  timeout: "5s",
  command: "exit 0"
}
```

I've also made the necessary changes in my local docker compose repo to add support in the v2 yaml file and it is working fine.

### DIFF
--- a/docker/utils/utils.py
+++ b/docker/utils/utils.py
@@ -24,7 +24,6 @@ import warnings
 from distutils.version import StrictVersion
 from datetime import datetime
 from fnmatch import fnmatch
-from pytimeparse.timeparse import timeparse
 
 import requests
 import six
@@ -66,6 +65,22 @@ def create_ipam_config(driver='default', pool_configs=None):
         'Driver': driver,
         'Config': pool_configs or []
     }
+
+
+def create_healthcheck(test=None, interval=None, timeout=None, retries=None):
+    return {
+        'Test': test,
+        'Interval': interval,
+        'Timeout': timeout,
+        'Retries': retries,
+    }
+
+
+def create_healthcheck_test_from_command(command=None):
+    return [
+        'CMD-SHELL',
+        command
+    ]
 
 
 def mkbuildcontext(dockerfile):
@@ -953,8 +968,6 @@ def format_environment(environment):
         return '{key}={value}'.format(key=key, value=value)
     return [format_env(*var) for var in six.iteritems(environment)]
 
-def duration_from_string(duration_string):
-    return timeparse(duration_string) * 1000000000
 
 def create_container_config(
     version, image, command, hostname=None, user=None, detach=False,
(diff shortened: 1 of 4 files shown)


aanand:
Thanks. This is definitely something we want in docker-py. I know it's WIP but I have a few comments about the code.

docker/utils/utils.py
```diff
@@ -980,6 +984,11 @@ def create_container_config(
             'stop_signal was only introduced in API version 1.21'
         )
 
+    if healthcheck is not None and compare_version('1.24', version) < 0:
+        raise errors.InvalidVersion(
+            'Health options were only introduced in API version 1.24'
+        )
+
```
aanand: You can use the `host_config_version_error` utility function here instead of writing a custom error string.

docker/utils/utils.py
```diff
@@ -1028,6 +1037,17 @@ def create_container_config(
...
+        healthcheck = {
+            'Test': [
+                'CMD-SHELL',
+                healthcheck.get('command')
+            ],
+            'Interval': duration_from_string(healthcheck.get('interval')) if healthcheck.has_key('interval') else None,
+            'Timeout': duration_from_string(healthcheck.get('timeout')) if healthcheck.has_key('timeout') else None,
+            'Retries': healthcheck.get('retries')
+        }
+
```
aanand: We need to raise an error if `healthcheck` isn't the right type. You can use `host_config_value_error`.

docker/utils/utils.py
```diff
@@ -952,6 +953,8 @@ def format_env(key, value):
         return '{key}={value}'.format(key=key, value=value)
     return [format_env(*var) for var in six.iteritems(environment)]
 
+def duration_from_string(duration_string):
+    return timeparse(duration_string) * 1000000000
```
aanand: I don't think we should parse duration strings ourselves and introduce another dependency. Users of docker-py can instead use whatever library they prefer, whether it's `timeparse` or `arrow` or doing the arithmetic themselves.

madjam002: Should I make it so you just pass in the raw `{ Test, Interval, Timeout, Retries }` object then?
The consumer could then be responsible for converting the duration string into a duration if necessary, and docker-compose could do the conversion from the cleaner `{ command, interval, timeout, retries }` object then.

aanand: Yeah. We should also provide a utility function for building the object, following the pattern we have with the networking data structures (e.g. `create_ipam_pool`):

```python
def create_healthcheck(test=None, interval=None, timeout=None, retries=None):
    return {
        'Test': test,
        'Interval': interval,
        'Timeout': timeout,
        'Retries': retries,
    }
```

madjam002: Cool, I'll do that then, thanks 😃

docker/utils/utils.py
madjam002: `host_config_version_error` only appears to be used inside `create_host_config`, this is error is thrown in `create_container_config`?

aanand: Ah, my mistake.

madjam002: I've updated the PR with some of the changes mentioned, I'm just in the process of making the subsequent changes in docker compose and writing unit tests.

shin-:
Closing via #1294

PR #1156, opened 2016-08-09
Exclude requests 2.11 from setup.py to work around unicode bug

### LINKED ISSUE #1155: Requests v2.11.0 causes breakage
This is a known issue within requests and looks like it will be fixed in v2.11.1. Documenting for the benefit of anyone else who runs into this :).

After a recent pip upgrade I found that docker-py was passing along an error from requests:

File "/usr/local/lib/python2.7/dist-packages/docker/client.py", line 307, in _stream_raw_result
   for out in response.iter_content(chunk_size=1, decode_unicode=True):
 File "/usr/local/lib/python2.7/dist-packages/requests/utils.py", line 372, in stream_decode_response_unicode
   raise UnicodeError("Unable to decode contents with encoding %s." % encoding)
UnicodeError: Unable to decode contents with encoding None.

This has already been reported to requests (https://github.com/kennethreitz/requests/issues/3481) and fixed (https://github.com/kennethreitz/requests/commit/d7f56ba9383575a6b7d361db0123a93c70a2b42f) for the next version.

If you are running into this issue, the easiest fix for now appears to be reverting to a pre 2.11 version of requests.

shin-:
Fixes #1155

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -9,7 +9,7 @@ ROOT_DIR = os.path.dirname(__file__)
 SOURCE_DIR = os.path.join(ROOT_DIR)
 
 requirements = [
-    'requests >= 2.5.2',
+    'requests >= 2.5.2, < 2.11',
     'six >= 1.4.0',
     'websocket-client >= 0.32.0',
 ]


dnephin:
LGTM

PR #1159, opened 2016-08-15
Fix for Client.from_env method

lsbardel:
It allows key-value parameters not used by `kwargs_from_env` function to be passed to the Client initialization method

### DIFF
--- a/docker/utils/utils.py
+++ b/docker/utils/utils.py
@@ -485,7 +485,8 @@ def parse_devices(devices):
     return device_list
 
 
-def kwargs_from_env(ssl_version=None, assert_hostname=None, environment=None):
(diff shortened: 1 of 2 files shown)


dnephin:
I don't think I like this idea. You can always do this:

```python

params = kwargs_from_env(...)
params.update(custom_params)
```

or, if you want to override defaults:

```python

params = {...}
params.update(kwargs_from_env(...))
```

Passing in defaults here is unnecessary, it makes it very difficult to expand the params of this function in the future.

shin-:
Thanks!

We went with a case-by-case approach in #1167 instead, to avoid surprises and make it easier to maintain in the future.
I'm going to close this PR as a result.

Issue #1030, opened 2016-04-07
CpuShares moved to HostConfig in API version 1.18 


dnephin:
See:
- https://docs.docker.com/engine/reference/api/docker_remote_api_v1.17/#create-a-container
- https://docs.docker.com/engine/reference/api/docker_remote_api_v1.18/#create-a-container

Looks like we're still passing it in container config. I wonder if engine is supporting this still?

It's breaking with swarm.

joshpurvis:
I needed this for swarm, so I implemented the changes in PR #1160.

bfirsh:
Fixed in #1160.

PR #1160, opened 2016-08-15
Move cpu_shares and cpuset_cpu to HostConfig when API >= 1.18

joshpurvis:
This PR moves the `cpu_share` and `cpuset_cpus` options to `HostConfig`, per the Docker remote API >= 1.18.

Previously `cpu_share` and `cpuset` arguments were passed into `create_container` as kwargs. I've added an InvalidVersion exception there, when the API version is < 1.18 and these arguments are passed. I can't seem to find documentation for < 1.18 to know if this should be a warning instead?

I should also note I've renamed `cpuset` to `cpuset_cpus` to match the existing `container_update` functionality, which is already expecting these being in `HostConfig`. 

Related issues: #1030, docker/compose#3284, docker/swarm#1983

### DIFF
--- a/docker/utils/utils.py
+++ b/docker/utils/utils.py
@@ -620,7 +620,7 @@ def create_host_config(binds=None, port_bindings=None, lxc_conf=None,
                        device_write_bps=None, device_read_iops=None,
                        device_write_iops=None, oom_kill_disable=False,
                        shm_size=None, sysctls=None, version=None, tmpfs=None,
-                       oom_score_adj=None):
+                       oom_score_adj=None, cpu_shares=None, cpuset_cpus=None):
 
     host_config = {}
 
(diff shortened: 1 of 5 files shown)


joshpurvis:
I've pushed a change/rebased the branch to fix an issue with the `InvalidVersion` error on the earlier API versions. (I had the API version comparison reversed in `create_container`.)

I also switched it from an error to a deprecation warning, as this mistake made me realize the old cpushare/cpuset unit tests would be broken when testing against newer API versions. (those that still pass these as kwargs into `create_container`: https://github.com/docker/docker-py/blob/7de30c3a9b75b12f26745c662f5d5bd3e1839021/tests/unit/container_test.py#L270)

Let me know if the old unit tests should simply be removed.

shin-:
Thanks, changes LGTM!

PR #1162, opened 2016-08-16
Remove references to "ExecutionDriver"

cpuguy83:
Docker no longer has an `ExecutionDriver` as of Docker 1.11.
The field in the `docker info` API will not be present in 1.13.

Found this while working on docker/docker#25721

Signed-off-by: Brian Goff <cpuguy83@gmail.com>


dnephin:
LGTM

PR #1164, opened 2016-08-19
Clean up Makefile

yuexiao-wang:
Clean up Makefile and keep the same targets between rules and PHONY.

Signed-off-by: yuexiao-wang <wang.yuexiao@zte.com.cn>


shin-:
Thanks for bringing this to our attention!

I just merged #1191 which should have the same effect.

PR #1167, opened 2016-08-23
Support version parameter in `Client.from_env`

### LINKED ISSUE #1075: Feature Request: docker.from_env(version='auto')
Feature request to add auto api version support for ```docker.from_env()``` similar to ```docker.Client(version='auto')```?

I noticed that one of the suggestions from #402 for the ```version='auto'``` option was now available for ```docker.Client()``` but doesn't work for ```docker.from_env()```.

shin-:
Fixes #1075

### DIFF
--- a/docker/client.py
+++ b/docker/client.py
@@ -114,7 +114,8 @@ class Client(
 
     @classmethod
     def from_env(cls, **kwargs):
-        return cls(**kwargs_from_env(**kwargs))
+        version = kwargs.pop('version', None)
+        return cls(version=version, **kwargs_from_env(**kwargs))
 
     def _retrieve_server_version(self):
         try:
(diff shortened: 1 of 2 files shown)


docker/client.py
```diff
@@ -114,7 +114,11 @@ def __init__(self, base_url=None, version=None,
 
     @classmethod
     def from_env(cls, **kwargs):
-        return cls(**kwargs_from_env(**kwargs))
+        version = None
+        if 'version' in kwargs:
+            version = kwargs['version']
+            del kwargs['version']
```
dnephin: I think you can write this as :

```python
version  = kwargs.pop('version', None)
```

dnephin:
LGTM, just a minor nit

PR #1177, opened 2016-08-31
Allow custom PID mode for the container

stepanstipl:
Docker added support for sharing PID namespaces with other containers
since version 1.12 (see https://github.com/docker/docker/pull/22481).

This removes limitation to `host` only option.

Signed-off-by: Stepan Stipl stepan@stipl.net


shin-:
Closing via #1297 - Thank you!

PR #1180, opened 2016-09-02
Fix Mount bind type sanity check

kanaka:
any() expects a single collection argument, not a list of arguments.


kanaka:
Fix is trivial and was made via web UI (signing is not worth it). Please consider this a bug report with a suggested fix.

shin-:
Oops. :disappointed: 

LGTM - thanks!

Issue #1023, opened 2016-04-05
Support credentials store API


shin-:
https://github.com/docker/docker/pull/20107

komuW:
after the merging of https://github.com/docker/docker-py/pull/1181 , I believe this issue ought to be closed now.

shin-:
Quite right :+1:

PR #1181, opened 2016-09-03
Add credentials store support

shin-:


shin-:
https://github.com/shin-/dockerpy-creds

shin-:
Fixes #1023

docker/auth/auth.py
```diff
@@ -105,6 +114,34 @@ def resolve_authconfig(authconfig, registry=None):
...
+            res['IdentityToken'] = data['Secret']
+        else:
+            res.update({
+                'Username': data['Username'],
+                'Password': data['Secret'],
+            })
+        return res
+    except dockerpycreds.StoreError as e:
+        log.error('Credentials store error: {0}'.format(repr(e)))
+
```
dnephin: I wonder if this should raise an exception? The user explicitly asked to use the credential store, right ?

dnephin:
LGTM, just one comment.

I'm not all that familiar with these credential stores, but the code looks good to me.

docker/auth/auth.py
shin-: Yeah, I need to differentiate a runtime error from a missing entry in dockerpycreds, then we can raise when there's an issue, and proceed with an unauthenticated push/pull if we just don't have credentials.

PR #1184, opened 2016-09-05
Fix licenses

bfirsh:
* Complete main LICENSE file
* Remove unnecessary and incorrect licenses from individual files


shin-:
Thanks!

PR #1187, opened 2016-09-07
Fix unit test which doesn't do anything

bfirsh:
It also overrode the fake API inspect endpoint with a broken
response.


shin-:
Nice. LGTM!

PR #1191, opened 2016-09-08
Add .PHONY for each makefile instruction

bfirsh:
Makes it much easier to keep this maintained properly. See also:

http://clarkgrubb.com/makefile-style-guide#phony-targets

Replaces #1164


dnephin:
LGTM

PR #1192, opened 2016-09-08
Drop support for Python 2.6

bfirsh:
Because it's ancient. If you're still using it, you can use an old
version of docker-py.


bfirsh:
I have incorporated the require_api_version change in #1217. When that's merged, I'll rebase on top of that.

bfirsh:
Rebased on top of #1217.

bfirsh:
If we're going to wait for 2.0 anyway, we may as well roll this into #1186.

PR #1208, opened 2016-09-13
Number of pools in adapter is configurable

### LINKED ISSUE #1207: Configurable num_pools
docker/compose#3586

shin-:
Default increased from 10 to 25

Fixes #1207


dnephin:
How will this integrate with Compose? Will we set the number of concurrent operations based on this constant in docker-py?

shin-:
@dnephin We could. As I wrote, I increased the default from 10 to 25 which should be enough to cover the majority of use cases, but if we find out some people still have issues it's possible to make Compose smarter with handling those.

dnephin:
I think that would be good.

LGTM

PR #1210, opened 2016-09-14
Add support for identity tokens in config file

shin-:


docker/auth/auth.py
```diff
@@ -189,12 +189,17 @@ def parse_auth(entries, raise_on_error=False):
...
-        conf[registry] = {
-            'username': username,
-            'password': password,
-            'email': entry.get('email'),
-            'serveraddress': registry,
-        }
+        if 'identitytoken' in entry:
+            conf[registry] = {
+                'IdentityToken': entry['identitytoken']
+            }
```
dnephin: It looks like `auth` is still required otherwise this function returns early. Is that expected or could someone has an `identitytoken` without an `auth` key?

shin-: Yeah, the token holds all the information. I had Matt Bentley test it with
UCP yesterday and he confirmed it works, too.

docker/auth/auth.py
```diff
@@ -182,13 +191,14 @@ def parse_auth(entries, raise_on_error=False):
                 'Auth data for {0} is absent. Client might be using a '
                 'credentials store instead.'
             )
-            return {}
+            conf[registry] = {}
```
shin-: Technically, either the client is using a cred store and all entries are empty, or they're not and none should be, so returning an empty dict here is not a big problem - but just in case weird hybrid `config.json` files start popping up, this should allow it to continue working.

shin-:
@dnephin PR updated - PTAL!

docker/auth/auth.py
dnephin: I think there is a missing return or continue here, no?

shin-: ah crud, yeah it needs a `continue`.

dnephin:
LGTM

PR #1214, opened 2016-09-14
Ignore not in swarm error when force leaving

bfirsh:
Real errors were getting swallowed in these tests, producing
other confusing cascading errors. This makes it much easier to
make sure a node is not in a Swarm, while also handling other
errors correctly.

### DIFF
--- a/docker/api/swarm.py
+++ b/docker/api/swarm.py
@@ -1,5 +1,6 @@
-from .. import utils
 import logging
+from six.moves import http_client
+from .. import utils
 log = logging.getLogger(__name__)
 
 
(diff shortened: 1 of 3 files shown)


docker/api/swarm.py
```diff
@@ -53,6 +53,9 @@ def join_swarm(self, remote_addrs, join_token, listen_addr=None,
     def leave_swarm(self, force=False):
         url = self._url('/swarm/leave')
         response = self._post(url, params={'force': force})
+        # Ignore "this node is not part of a swarm" error
+        if force and response.status_code == 406:
```
dnephin: `s/406/httplib.NOT_ACCEPTABLE/`

bfirsh: Thanks - fixed.

dnephin:
LGTM

PR #1217, opened 2016-09-15
Move BaseTestCase to BaseIntegrationTest

bfirsh:
Because two things called `BaseTestCase` is quite confusing. I haven't bothered refactoring the other `BaseTestCase` because that disappears anyway when we drop Python 2.6 support in #1192.

### DIFF
--- a/docker/client.py
+++ b/docker/client.py
@@ -108,8 +108,10 @@ class Client(
 
     @classmethod
     def from_env(cls, **kwargs):
+        timeout = kwargs.pop('timeout', None)
         version = kwargs.pop('version', None)
-        return cls(version=version, **kwargs_from_env(**kwargs))
+        return cls(timeout=timeout, version=version,
+                   **kwargs_from_env(**kwargs))
 
     def _retrieve_server_version(self):
         try:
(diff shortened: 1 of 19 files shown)


bfirsh:
Rebased. /cc @docker/core-docker-py-maintainers

shin-:
@bfirsh Looks like you accidentally included 7f64928 ?

bfirsh:
Intentionally -- it allows `timeout` to be passed to `from_env()`.

shin-:
I think we've recently ruled against doing something similar, see https://github.com/docker/docker-py/pull/1159#issuecomment-239861637 
and the solution we went with instead: #1167 

I think we should keep with the same guidelines here.

bfirsh:
Hm, alright. `from_env()` should accept all kwargs for `Client` though, so I'll pull that out into `from_env()`.

bfirsh:
@shin- Made a simpler version which just adds `timeout` to `from_env()`. :)

aanand:
LGTM

bfirsh:
ping @shin-

shin-:
LGTM, thank you!

PR #1218, opened 2016-09-15
Simplify some unnecessarily complex file structures

bfirsh:


bfirsh:
Rebased. /cc @docker/core-docker-py-maintainers

dnephin:
This LGTM, but I'd like to make sure @shin- is ok with it.

shin-:
Not against it in principle, slightly worried about breaking stuff for people who might have a `from docker.auth.auth` import or similar. May be reasonable to put it in the 2.0 milestone?

bfirsh:
Sure! Sounds reasonable to me - it's plausible that it could break code and
it isn't urgent.

bfirsh:
Rolled into #1186

PR #1219, opened 2016-09-16
Add healthcheck to create-container

bubaflub:
Builds on work from PR #1152.  Differences include:
- Resolved the merge conflict
- pulled out the healthcheck into a separate file in docker/types
- added an integration test that exercises a healthcheck

I'm open to any suggestions to clean this up.

### DIFF
--- a/docker/api/container.py
+++ b/docker/api/container.py
@@ -115,7 +115,8 @@ class ContainerApiMixin(object):
                          cpu_shares=None, working_dir=None, domainname=None,
                          memswap_limit=None, cpuset=None, host_config=None,
                          mac_address=None, labels=None, volume_driver=None,
-                         stop_signal=None, networking_config=None):
+                         stop_signal=None, networking_config=None,
+                         healthcheck=None):
 
         if isinstance(volumes, six.string_types):
             volumes = [volumes, ]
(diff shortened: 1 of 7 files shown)


bubaflub:
Also including @ryan0x44 who helped me with this and @shin- who we've chatted with briefly about this.

docker/api/container.py
```diff
@@ -320,7 +321,7 @@ def start(self, container, binds=None, port_bindings=None, lxc_conf=None,
               dns=None, dns_search=None, volumes_from=None, network_mode=None,
               restart_policy=None, cap_add=None, cap_drop=None, devices=None,
               extra_hosts=None, read_only=None, pid_mode=None, ipc_mode=None,
-              security_opt=None, ulimits=None):
+              security_opt=None, ulimits=None, healthcheck=None):
```
shin-: This part is incorrect - we do not want to update the `start` method with new parameters. Especially in this case, healthcheck is part of the `ContainerConfig`, not the `HostConfig`, so it's not valid to have here.

docker/utils/utils.py
```diff
@@ -1025,6 +1025,11 @@ def create_container_config(
             'stop_signal was only introduced in API version 1.21'
         )
 
+    if healthcheck is not None and compare_version('1.24', version) < 0:
```
shin-: You can use `version_gte(version, '1.24')` here, which is more explicit.

docker/api/container.py
shin-: If you fix that it should solve the issue with the tests as well.

bubaflub:
Great, I'll fix things up and get the tests to pass.

bubaflub:
I've rebased off of master and fixed the merge conflict.  I've made the changes you requested and now all of `make test` passes locally.  I also appeased pep8.  Slightly related -- I used autopep8 to fix up those nits.  Highly recommended.

shin-:
LGTM now, thanks!

shin-:
Merged via #1294

PR #1158, opened 2016-08-15
Add support for updating RestartPolicy

mvdstam:
This is a really quick fix, but hoping to get some help from more experienced Python devs here to get the PR nice and in order.

Issue #1106, opened 2016-06-24
Support updating the restart policy attribute of a running container


gkze:
Hi,

It looks like updating the restart policy of a running container is not supported yet (https://github.com/docker/docker-py/blob/88811a26593e8a87e8d820d8820736fea6d8d20a/docker/api/container.py#L404).

Is there any work being done on this? Didn't want to duplicate efforts, I would be more than happy to issue a PR.

Thanks!

mferon:
Hi,

Here is a pull request about it : #1158

PR #1224, opened 2016-09-20
Add support for restart policy update

mferon:
Hi all,

Here is a patch to be able to update the restart policy of a container. (Issue #1106)

A PR already exists (#1158), but it seems abandoned...

I'm open to every suggestion ;)

### DIFF
--- a/docker/api/container.py
+++ b/docker/api/container.py
@@ -421,7 +421,8 @@ class ContainerApiMixin(object):
     def update_container(
         self, container, blkio_weight=None, cpu_period=None, cpu_quota=None,
         cpu_shares=None, cpuset_cpus=None, cpuset_mems=None, mem_limit=None,
-        mem_reservation=None, memswap_limit=None, kernel_memory=None
+        mem_reservation=None, memswap_limit=None, kernel_memory=None,
+        restart_policy=None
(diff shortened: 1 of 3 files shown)


docker/api/container.py
```diff
@@ -416,12 +416,13 @@ def unpause(self, container):
         res = self._post(url)
         self._raise_for_status(res)
 
-    @utils.minimum_version('1.22')
+    @utils.minimum_version('1.23')
```
shin-: This should not be changed - people still need to be able to update containers on 1.22

docker/api/container.py
```diff
@@ -445,6 +446,8 @@ def update_container(
             data['MemorySwap'] = utils.parse_bytes(memswap_limit)
         if kernel_memory:
             data['KernelMemory'] = utils.parse_bytes(kernel_memory)
+        if restart_policy:
```
shin-: There needs to be a `version_lt` check here for 1.23 instead

mferon:
Thank you for your review ;)

I updated the branch to fix what you said.

shin-:
Thanks - LGTM!

dnephin:
LGTM

PR #1226, opened 2016-09-26
~~remove~~ document ipaddress dependency

TomasTomecek:
EDIT: instead document it

~~as it's not being used anymore:~~

```
$ grep -R "import" . | grep "ipaddress"

$ grep -R "ipaddress" --exclude-dir .git .
./requirements.txt:ipaddress==1.0.16 ; python_version < '3.3'
./setup.py:    ':python_version < "3.3"': 'ipaddress >= 1.0.16',
```

~~it looks like it was used only in ssl_match_hostname module~~

```
git rev-list --all | xargs git grep ipaddress | grep import
d80e75f5eaca18c19257c13b91aae1a2493449d5:docker/ssladapter/ssl_match_hostname.py:import ipaddress
...
```

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -20,6 +20,9 @@ if sys.platform == 'win32':
 
 extras_require = {
     ':python_version < "3.5"': 'backports.ssl_match_hostname >= 3.5',
+    # While not imported explicitly, the ipaddress module is required for
+    # ssl_match_hostname to verify hosts match with certificates via
+    # ServerAltname: https://pypi.python.org/pypi/backports.ssl_match_hostname
     ':python_version < "3.3"': 'ipaddress >= 1.0.16',
 }
 


shin-:
From the `backports.ssl_match_hostname` [pypi page](https://pypi.python.org/pypi/backports.ssl_match_hostname):

> If you want to verify hosts match with certificates via ServerAltname 
IPAddress fields, you need to install the `ipaddress module`_. 
backports.ssl_match_hostname will continue to work without ipaddress but 
will only be able to handle ServerAltName DNSName fields, not IPAddress. 
System packagers (Linux distributions, et al) are encouraged to add 
this as a hard dependency in their packages.

TomasTomecek:
@_@

Thanks for clarification. We live in a crazy world.

So I put a comment on top of the requirement in `setup.py` to clarify instead.

setup.py
```diff
@@ -20,6 +20,9 @@
 
 extras_require = {
     ':python_version < "3.5"': 'backports.ssl_match_hostname >= 3.5',
+    # ipaddress module is actually not used, it's required for
```
shin-: I'd suggest changing the wording to

> While not imported explicitly, the ipaddress module is required for [...]

shin-:
@TomasTomecek That sounds good. I'm sure other people will be confused by it.

Just a small comment on wording, otherwise LGTM.

TomasTomecek:
I like your rewrite. Updated.

TomasTomecek:
rebased

shin-:
Thanks :+1:

TomasTomecek:
\o/

PR #1227, opened 2016-09-26
Fix documenation for parameter `domainname` in `create_container`

interesse:
Before it looked more like the description of the parameter `dns_search` which is correctly part of the host config. See https://docs.docker.com/engine/reference/api/docker_remote_api_v1.25/#/create-a-container for the original docker documentation for this parameter.


ijc:
PR grooming: Thanks for your PR, sorry nobody got to it at the time. I think the changes here were obsoleted by c7a3aa7e446536e048bc02ca881ec14016f08675 so I'm going to close. Thanks again.

PR #1228, opened 2016-09-27
Synthesize execute permission bit

### LINKED ISSUE #937: Executable permissions not synthesized when building from Windows

### LINKED ISSUE #1206: Client.build should +x all context files on Windows
docker/compose#3716

christianbundy:
This includes @seschwar's commit from #938 as well as a fix the empty file object that was causing tests to fail. Please let me know if there are any issues, thanks for maintaining this project!

Resolves #937 
Resolves #1206 

After this is merged, a version bump [here](https://github.com/docker/compose/blob/fc6791f3f0e3e8ae23d9b380398f39bea2e83101/requirements.txt#L4) and [here](https://github.com/docker/compose/blob/fc6791f3f0e3e8ae23d9b380398f39bea2e83101/setup.py#L37) should resolve docker/compose#3716.

---

It looks like this may depend on #1192, unless backward-compatibility with Python 2.6 is important. Backporting wouldn't be too difficult, but we'd have to add some complexity for wrangling the `IsADirectoryError` exception before we try to pass a file object with `TarFile.add()`. I can either rebase on top of d21feea or backport to Python 2.6, just let me know.

### DIFF
--- a/docker/utils/utils.py
+++ b/docker/utils/utils.py
@@ -94,9 +94,20 @@ def tar(path, exclude=None, dockerfile=None, fileobj=None, gzip=False):
 
     for path in sorted(exclude_paths(root, exclude, dockerfile=dockerfile)):
         i = t.gettarinfo(os.path.join(root, path), arcname=path)
+
         if sys.platform == 'win32':
+            # Windows doesn't keep track of the execute bit, so we make files
+            # and directories executable by default.
             i.mode = i.mode & 0o755 | 0o111
-        t.addfile(i)
+
+        try:
+            # We open the file object in binary mode for Windows support.
+            f = open(os.path.join(root, path), 'rb')
+        except IOError:
+            # When we encounter a directory the file object is set to None.
+            f = None
+
+        t.addfile(i, f)
 
     t.close()
     fileobj.seek(0)
--- a/docker/utils/utils.py
+++ b/docker/utils/utils.py
@@ -4,6 +4,7 @@ import os
 import os.path
 import json
 import shlex
+import sys
 import tarfile
 import tempfile
 import warnings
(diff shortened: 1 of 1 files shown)


dnephin:
It looks like #1192 is in the 2.0 milestone, so probably won't be merged right away. It would be great to get this in earlier. It's been a commonly raised issue in Compose and docker-py.

If you could find a way to backport `filter=` that would be great.

christianbundy:
Thanks for the quick review @dnephin. I amended my commit to use a 2.6-friendly `try...except` block around an `open()` call. For what it's worth, I've also used the `'rb'` mode for opening the file, since [the docs](https://docs.python.org/2.6/tutorial/inputoutput.html#reading-and-writing-files) say that's important for Windows. It looks like the test are all passing now, please let me know if you notice anything else.

docker/utils/utils.py
```diff
@@ -92,7 +93,17 @@ def tar(path, exclude=None, dockerfile=None, fileobj=None, gzip=False):
     exclude = exclude or []
 
     for path in sorted(exclude_paths(root, exclude, dockerfile=dockerfile)):
-        t.add(os.path.join(root, path), arcname=path, recursive=False)
+        i = t.gettarinfo(os.path.join(root, path), arcname=path)
+
+        if sys.platform == 'win32':
+            i.mode = i.mode & 0o755 | 0o111
+
+        try:
```
dnephin: A comment explaining why this is necessary would be great

christianbundy: On it!

christianbundy: Done.

shin-:
LGTM - thank you!

seschwar:
Thanks for having picked this up, @christianbundy. My employer only had a very short lived interest in being able to build on Windows.

christianbundy:
@seschwar Thanks for getting it started! Our team supports Windows so this is super helpful for us.

PR #1232, opened 2016-09-28
replace on_failure with on-failure

pacoxu:
According to docker docs and docker-py docs, it should be on-failure.

Signed-off-by:  徐俊杰<roollingstone@gmail.com>

### DIFF
--- a/docker/types/services.py
+++ b/docker/types/services.py
@@ -152,7 +152,7 @@ class UpdateConfig(dict):
 class RestartConditionTypesEnum(object):
     _values = (
         'none',
-        'on_failure',
+        'on-failure',
         'any',
     )
     NONE, ON_FAILURE, ANY = _values


pacoxu:
1. Remote API 1.25: https://docs.docker.com/engine/reference/api/docker_remote_api_v1.25/#/create-a-service
RestartPolicy – Specification for the restart policy which applies to containers created as part of this service.
Condition – Condition for restart (none, on-failure, or any).
2. Docker cli: https://github.com/docker/docker/blob/c2d6e76a704604a964d2a593b8634d0bcbc53cd0/docs/reference/commandline/service_create.md
      --restart-condition string       Restart when condition is met (none, on-failure, or any)

thaJeztah:
Related: https://github.com/docker/docker/pull/27014, as there's a bug in the "inspect" output of services through the API

shin-:
LGTM, thanks!

PR #1234, opened 2016-09-29
Adjust test_create_network_with_ipam_config

aboch:
This is related to the change in https://github.com/docker/docker/pull/26880
in which docker-py run as part of experimental CI cathes the API change.

IPAM now also includes `Data` besides `Config`, therefore the raw equality check can no longer be performed the way it was done before.

Signed-off-by: Alessandro Boch <aboch@docker.com>


shin-:
LGTM, Thanks!

PR #1235, opened 2016-09-29
Name is not required when creating a docker volume

rmb938:
The name of a volume isn't required when creating. See: https://docs.docker.com/engine/reference/api/docker_remote_api_v1.25/#/create-a-volume

```
JSON parameters:

    Name - The new volume’s name. If not specified, Docker generates a name.
    Driver - Name of the volume driver to use. Defaults to local for the name.
    DriverOpts - A mapping of driver options and values. These options are passed directly to the driver and are driver specific.
    Labels - Labels to set on the volume, specified as a map: {"key":"value","key2":"value2"}

```

### DIFF
--- a/docker/api/volume.py
+++ b/docker/api/volume.py
@@ -38,7 +38,8 @@ class VolumeApiMixin(object):
         return self._result(self._get(url, params=params), True)
 
     @utils.minimum_version('1.21')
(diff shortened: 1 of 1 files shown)


shin-:
Thank you :+1:

Issue #1244, opened 2016-10-04
unexpected keyword argument 'num_pools'


tardyp:
Got the following error since docker-py 1.10.3

```
        File "[]/local/lib/python2.7/site-packages/docker/client.py", line 89, in __init__
	    num_pools=num_pools
	  File "[]/local/lib/python2.7/site-packages/docker/ssladapter/ssladapter.py", line 32, in __init__
	    super(SSLAdapter, self).__init__(**kwargs)
	exceptions.TypeError: __init__() got an unexpected keyword argument 'num_pools'
```
looks like issue comes with 
https://github.com/docker/docker-py/pull/1208/files#diff-5262437d121a30fc85a81c4b1da57e4bR89

requests's HTTPAdapter API is pool_connections for number of connection of the pool

PR #1245, opened 2016-10-04
fix for got an unexpected keyword argument 'num_pools'

tardyp:
requests's HTTPAdapter API is pool_connections for number of connection of the pool

fix for #1244


shin-:
LGTM, thank you!

tardyp:
Hi @shin- Can we have a 1.10.4 for this fix? this would avoid me to fix my requirement.txt to docker-py==1.10.2

PR #1250, opened 2016-10-11
do not assume that read will consume the number of bytes requested

### LINKED ISSUE #1211: attach is causing an "Invalid Argument" exception from os.read
``` python
stream = client.attach(container, stream=True, stdout=True, stderr=True)
for chunk in stream:
    pass
```

Results in:

```
  File "/Users/michael/work/oss/marina/marina/build.py", line 695, in watcher
    for chunk in stream:
  File ".venv/lib/python3.5/site-packages/docker/utils/socket.py", line 67, in frames_iter
    yield read(socket, n)
  File ".venv/lib/python3.5/site-packages/docker/utils/socket.py", line 25, in read
    return os.read(socket.fileno(), n)
OSError: [Errno 22] Invalid argument
```

Using docker-py 1.10.2 on OS X 10.11.6 with docker for mac 1.12.0-rc3. Reverting to 1.9.0 fixes the issue.

mmerickel:
The issue is that `os.read` does not always read the expected number of
bytes, and thus we are moving to the next frame too early causing drift
in the byte stream. When the reading drifts, it starts reading garbage
as the next frame size. Some examples of frame sizes were
4032897957 bytes, etc. Values this large were causing the exceptions
from `os.read`.

fixes #1211

Signed-off-by: Michael Merickel michael@merickel.org

### DIFF
--- a/docker/utils/socket.py
+++ b/docker/utils/socket.py
@@ -69,7 +69,11 @@ def frames_iter(socket):
     """
     Returns a generator of frames read from socket
     """
-    n = next_frame_size(socket)
-    while n > 0:
-        yield read(socket, n)
+    while True:
         n = next_frame_size(socket)
+        if n == 0:
+            break
+        while n > 0:
+            result = read(socket, n)
+            n -= len(result)
+            yield result


mmerickel:
What's the status here? If I rebase it will someone promise to review it this time?

docker/utils/socket.py
```diff
@@ -64,5 +64,8 @@ def frames_iter(socket):
     """
     n = next_frame_size(socket)
     while n > 0:
```
shin-: I feel like this could be rewritten in a more intuitive way. I feel like `n` should not be reset inside the loop that checks for its value. Here's what I would suggest as an alternative:

```python
while True:
  n = next_frame_size(socket)
  if n == 0:
    break
  while n > 0:
    result = read(socket, n)
    n -= len(result)
    yield result
```

mmerickel: Updated in cd53960.

mmerickel:
@aanand Ok rebased for the second time...

shin-:
Thank you!

PR #1254, opened 2016-10-12
Several fixes to npipe support

shin-:
- Do not allow bufsize to be 0 in `NpipeSocket.makefile()`
- Fix `_get_raw_response_socket` to always return the `NpipeSocket` object
- Override `NpipeHTTPConnectionPool._get_conn` to avoid crash in urllib3
- Fix `NpipeSocket.recv_into` for Python 2
- Do not call `select()` on `NpipeSocket` objects

### DIFF
--- a/docker/client.py
+++ b/docker/client.py
@@ -220,7 +220,9 @@ class Client(
 
     def _get_raw_response_socket(self, response):
         self._raise_for_status(response)
-        if six.PY3:
+        if self.base_url == "http+docker://localnpipe":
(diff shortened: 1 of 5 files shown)


christianbundy:
Looks great, thank you! Excited to test this in Compose.

shin-:
@christianbundy If you want, we have a dev build [here](https://ci.appveyor.com/project/docker/compose/build/master-2433/artifacts) that includes those fixes. Testing and feedback will be massively appreciated :)

PR #1261, opened 2016-10-17
Requirements are aligned with compose repo

mrjj:
This was made to avoid requests versions conflict during installation docker-compose and docker-py into the same venv


TomasTomecek:
Why not let compose update to newer `requests`?

ijc:
PR grooming: Sorry noone looked at this for such a long time. I can see that right now the two are in sync at `requests==2.20.0` so I'll close. Thanks though.

PR #1264, opened 2016-10-18
create_service env can be a list or a dict

manics:
The docker-py docs say the `env` parameter of `create_service` is a dictionary, but this isn't accepted, it has to be a list:

```
>>> import docker
>>> c = docker.Client()
>>> c.version()
{u'KernelVersion': u'4.4.20-moby', u'Os': u'linux', u'BuildTime': u'2016-08-18T17:52:38.255390991+00:00', u'ApiVersion': u'1.24', u'Version': u'1.12.1', u'GitCommit': u'23cf638', u'Arch': u'amd64', u'GoVersion': u'go1.6.3'}
```

```
>>> contspec = docker.types.ContainerSpec(image='httpd', env={'A': 'B'})
>>> c.create_service(docker.types.TaskTemplate(contspec), name='test')
  File "docker/utils/decorators.py", line 35, in wrapper
    return f(self, *args, **kwargs)
  File "docker/api/service.py", line 33, in create_service
    self._post_json(url, data=data, headers=headers), True
  File "docker/client.py", line 180, in _result
  File "docker/client.py", line 176, in _raise_for_status
    raise errors.APIError(e, response, explanation=explanation)
docker.errors.APIError: 500 Server Error: Internal Server Error ("json: cannot unmarshal object into Go value of type []string")
```

```
>>> contspec = docker.types.ContainerSpec(image='httpd', env=['A=B'])
>>> c.create_service(docker.types.TaskTemplate(contspec), name='test')
{u'ID': u'5f4u3my4beeiecjzhx90q7aqc'}
```

### DIFF
--- a/docker/types/services.py
+++ b/docker/types/services.py
@@ -39,6 +39,7 @@ class ContainerSpec(dict):
     def __init__(self, image, command=None, args=None, env=None, workdir=None,
                  user=None, labels=None, mounts=None, stop_grace_period=None):
         from ..utils import split_command  # FIXME: circular import
+        from ..utils import format_environment  # FIXME: circular import
 
         self['Image'] = image
 
@@ -48,7 +49,10 @@ class ContainerSpec(dict):
         self['Args'] = args
 
         if env is not None:
-            self['Env'] = env
+            if isinstance(env, dict):
+                self['Env'] = format_environment(env)
+            else:
+                self['Env'] = env
         if workdir is not None:
             self['Dir'] = workdir
         if user is not None:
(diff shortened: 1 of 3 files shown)


shin-:
Rather than modify the documentation, I think we should modify the code in `ContainerSpec` to call `docker.utils.format_environment`.

manics:
@shin- I've tried to modify the code as suggested (separate branch): https://github.com/manics/docker-py/commit/bf1e41f70f9b6059a07218a528b52413e5fd047f
However this fails because adding `from ..utils.utils import format_environment` to `docker/types/services.py` introduces a circular dependency with `docker/utils/__init__.py`:
```
from ..types import LogConfig, Ulimit
from ..types import SwarmExternalCA, SwarmSpec
```
Do you have any advice on how best to deal with this?

shin-:
@manics Sorry, I seem to have missed your message. A temporary solution would be to do a delayed import like here: https://github.com/docker/docker-py/blob/master/docker/types/services.py#L41

manics:
@shin- Thanks, I've rebased, removed the previous commit and updated the code.

shin-:
Thank you - LGTM!

PR #1266, opened 2016-10-18
Add --rm to docker run in Makefile

bfirsh:
So tests don't leave containers all over the place. The downside
is this makes it a bit harder to debug a test's filesystem -
you'll have to remove the "--rm" and run the test again.

Signed-off-by: Ben Firshman ben@firshman.co.uk


bfirsh:
ping @shin- @dnephin

dnephin:
LGTM

PR #1274, opened 2016-10-27
Implement retry logic when the npipe open procedure fails

shin-:
Fixes the compose `parallel_remove` bug.

cc @aanand @dnephin 

Signed-off-by: Joffrey F <joffrey@docker.com>


aanand:
LGTM

PR #1277, opened 2016-10-28
Do not traverse excluded directories that are not prefixes of dockerignore exceptions.

### LINKED ISSUE #1276: Creating build context may take a long time when dockerignore contains an exception to a rule
Ran into an issue at work where our local builds using docker-compose where taking a significant amount of time more than our builds using the docker cli tool. This only occurred where we had some exception line in the dockerignore file and some other excluded folder that contained a significant amount of files. The below would be an example of the problematic dockerignore file

```
# Folder containing lots and lots of small files that should be excluded from a build
node_modules

#Other folder that should be excluded
config/env

#Exception to the above rule
!config/env/docker.js
```

Something like the above resulted in docker-compose taking nearly a minute or more longer to complete the create build context step for one of our images when compared to doing the same using docker cli. This created a lot of frustration in team members trying to do simple 'Change Code -> Rebuild -> Test' development.

I was able to isolate it down to a difference between how docker-py handles exceptions in dockerignore versus how docker handles the exceptions. While docker-py currently conforms to the older algorithm for docker that checked every file for a potential match to the exception, the newer docker versions skip excluded directories completely if their path is not a prefix to some exception rule.

I created a patch that has the algorithm conform to the newer docker algorithm. I'll submit a PR for your review, as I think using the newer algorithm will significantly speed up our build times.

```
docker-compose version 1.8.0, build f3628c7
docker-py version: 1.9.0
CPython version: 2.7.9
OpenSSL version: OpenSSL 1.0.2h  3 May 2016
```

bodnarbm:
Closes #1276.

Updates the algorithm implemented in #863 to not traverse down into excluded directories if their path is not a prefix for some .dockerignore exception.

I believe this change conforms with the [current handling](https://github.com/docker/docker/blob/master/pkg/archive/archive.go#L640-L671) for dockerignore exceptions by docker itself.

Signed-off-by: Brandon Bodnar bodnarbm@gmail.com

### DIFF
--- a/docker/utils/utils.py
+++ b/docker/utils/utils.py
@@ -214,6 +214,31 @@ def should_include(path, exclude_patterns, include_patterns):
     return True
 
 
+def should_check_directory(directory_path, exclude_patterns, include_patterns):
+    """
+    Given a directory path, a list of exclude patterns, and a list of inclusion
+    patterns:
+
+    1. Returns True if the directory path should be included according to
+       should_include.
+    2. Returns True if the directory path is the prefix for an inclusion
+       pattern
+    3. Returns False otherwise
+    """
+
+    # To account for exception rules, check directories if their path is a
+    # a prefix to an inclusion pattern. This logic conforms with the current
+    # docker logic (2016-10-27):
+    # https://github.com/docker/docker/blob/bc52939b0455116ab8e0da67869ec81c1a1c3e2c/pkg/archive/archive.go#L640-L671
+
+    path_with_slash = directory_path + os.sep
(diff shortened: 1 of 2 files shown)


dnephin:
A test would be great

bodnarbm:
I should be able to add a unit test for the should_check_directory function this evening.

bodnarbm:
I added some unit tests for the new `should_check_directory` function in `docker/utils/utils.py`

The overall behavior of processing `.dockerignore` when creating the tar should still be covered by the integration tests added in #863.

bfirsh:
LGTM

Thanks!

PR #1279, opened 2016-10-29
fix up a couple links in the docs

cglewis:


bfirsh:
We had a big docs refactor and none of this exists any longer. Thanks for the contribution, though!

PR #1281, opened 2016-11-01
ssl,test: OpenSSL may not support ssl2

### LINKED ISSUE #1265: test_only_uses_tls: failure on Fedora with 1.10.4
So I was updating docker-py in Fedora land to `1.10.4` and this happened during the RPM build process:

```
tests/unit/ssladapter_test.py:27: in test_only_uses_tls
    assert ssl_context.options & OP_NO_SSLv2
E   AssertionError: assert (2181170175L & 0)
E    +  where 2181170175L = <ssl.SSLContext object at 0x3ffa4cbdde8>.options
```

https://kojipkgs.fedoraproject.org//work/tasks/2344/16112344/build.log

Since we unbundle urllib3 from requests it could be the issue. Or some system default changed. Didn't have time to dig through.

Creating the issue just to see if any other distros bumped into this.

TomasTomecek:
Fixes #1265 

Should I also apply the same code for ssl3?

### DIFF
--- a/tests/unit/ssladapter_test.py
+++ b/tests/unit/ssladapter_test.py
@@ -24,7 +24,8 @@ class SSLAdapterTest(base.BaseTestCase):
         ssl_context = ssladapter.urllib3.util.ssl_.create_urllib3_context()
 
         assert ssl_context.options & OP_NO_SSLv3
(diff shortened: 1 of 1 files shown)


TomasTomecek:
rebased

PR #1239, opened 2016-10-01
Fix several create_service arguments

shin-:
Fixes #1212 
Fixes #1233

PR #1288, opened 2016-11-04
Use EndpointSpec for Create and Update service

### LINKED ISSUE #1233: Service Port Publishing Example
Does the library support defining ports on a service? (i.e. with swarm) It looks like the endpoint_config parameter could be used for this but I can't see an example in the docs. (The general API docs also seem a little confused about how to define ports.)

JoelJ:
Fixes #1233

See documentation for reference for what the proper fix is: https://docs.docker.com/engine/reference/api/docker_remote_api_v1.24/#/update-a-service

### DIFF
--- a/docker/api/service.py
+++ b/docker/api/service.py
@@ -27,7 +27,7 @@ class ServiceApiMixin(object):
             'Mode': mode,
             'UpdateConfig': update_config,
             'Networks': networks,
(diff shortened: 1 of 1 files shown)


JoelJ:
Closing because I just noticed this is a duplicate of PR: #1239

PR #1294, opened 2016-11-11
Add support for passing healthcheck to create_container

aanand:
Carries @madjam002's work in #1152 and @bubaflub's in #1219.

### DIFF
--- a/docker/types/healthcheck.py
+++ b/docker/types/healthcheck.py
@@ -1,0 +1,47 @@
+from .base import DictType
+
+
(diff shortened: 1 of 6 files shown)


docker/types/healthcheck.py
```diff
@@ -0,0 +1,53 @@
+from .base import DictType
+
+import six
+
+
+class Healthcheck(DictType):
```
dnephin: Is this class basically just a namedtuple?

PR #1301, opened 2016-11-22
Fix ContainerApiMixin.copy with dict container arg

biniambekele:
Signed-off-by: biniambekele <biniambekele@users.noreply.github.com>


shin-:
Thanks, LGTM.

PR #1303, opened 2016-11-23
Show a helpful warning when people try to call `client.containers()`

aanand:
People upgrading to docker-py 2.0 without being aware of the new client API will likely try to call the old `containers()` method. This adds a helpful warning telling them to use APIClient to get the old API.

### DIFF
--- a/tests/unit/client_test.py
+++ b/tests/unit/client_test.py
@@ -1,5 +1,6 @@
 import datetime
 import docker
+from docker.utils import kwargs_from_env
 import os
 import unittest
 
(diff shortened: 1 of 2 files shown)


tests/unit/client_test.py
```diff
@@ -1,5 +1,6 @@
 import datetime
 import docker
+from docker.utils import kwargs_from_env
```
aanand: I know I could have just used `docker.from_env()` in the test, but I wanted something as close as possible to what users of the old docker-py would actually have written. Let me know if you think this is silly.

bfirsh: This is silly in a good, thoughtful way.

bfirsh:
LGTM

PR #1305, opened 2016-11-23
Update setup.py and things

bfirsh:
A few things here:

- Update package name to be `docker`, rather than `docker-py`. This wasn't explicitly mentioned in #1186, but seems like like a sensible thing to me because:
  1. [We have the package on PyPi](https://pypi.python.org/pypi/docker)
  2. It means `pip install docker-py` won't break when we release 2.0.
  3. It means `pip install ...` and `import ...` are consistent.
- Tidy up some metadata in `setup.py`
- Use `find_packages` in `setup.py`
- Bump the version to `2.0.0-dev`


shin-:
LGTM!

PR #1308, opened 2016-11-24
Rename Client -> DockerClient

shin-:
cc @bfirsh @aanand @dnephin 

This is intended as a follow-up to https://github.com/docker/docker-py/pull/1186#discussion_r79425908

Reasons I think we should make this change:

- Avoids confusion with the `Client` object in 1.x. We already acknowledge that confusion exists with the overrides to `Client.__getattr__` and `Container.__call__`, so it seems sensible to prevent this scenario as much as possible.
- If one class is called `Client` and the other `<SomeQualifier>Client`, it's easy to assume the latter to be a subclass of the former. 
- The name `DockerClient` denotes it is the "Docker" experience with streamlined, user-friendly interface, contrasting with `APIClient` which is the lower-level interacting class.

I'm happy to discuss other naming options, but I believe reusing the `Client` name is a mistake that will cost us in the long term.


docs/client.rst
```diff
@@ -1,19 +1,19 @@
-Client
-======
+DockerClient
```
bfirsh: I think this should just be "Client" -- it's the title of the documentation page about clients, not the specific object name.

In the same way below, it says "Creating a client", not "Creating a DockerClient".

docs/services.md
```diff
@@ -6,7 +6,7 @@ using the Docker Engine API. Note that the engine needs to be part of a
```
bfirsh: Oops – I think this file is redundant now. The content has been moved to docstrings and `api.rst`. There is also a file called `services.rst` which will be displayed instead of this file.

We can just delete it, I think.

shin-: There's a "user guide" part that we'd like to preserve I think - let me move the relevant bits to a different folder and we can figure out converting to RST / where to include at a later time.

shin-:
@bfirsh 
- Moved the user guide part of services.md to `docs/user_guides/swarm_services.md` (To be converted / amended at a later date)
- Moved `docker.types.services` docs to docstrings

PTAL!

bfirsh:
LGTM

@dnephin @aanand any better name ideas?

aanand:
`DockerClient` is fine by me. It has the advantage that inspecting the object makes it clear what it is (it's very easy to imagine an application holding onto several "client" objects from different libraries, so being able to see instantly that this is a `<DockerClient ...>` object could save people a lot of debugging time.

(The only nit is that, by this logic, `APIClient` should be `DockerAPIClient`. Thoughts?)

bfirsh:
The point of this is to disambiguate `DockerClient` and `APIClient`, so continuing that and your logic, they should be `DockerSomethingClient` and `DockerAPIClient`. (This will end in madness.)

aanand:
Should it have `Client` in the name at all?

A `DockerClient` object is part of a graph of objects that represent different things managed by the Docker Engine - `ContainerCollection`, `Container`, `Swarm`. If `DockerClient` has `Client` on the end, shouldn't each of them have `Client` on the end too? That would be tedious.

Maybe the right name is `Engine`.

TomasTomecek:
So... No backwards compatibility? I have also noticed that `AutoVersionClient` is gone.

aanand:
@TomasTomecek Correct. It's a 2.0 release, so a rare opportunity to break backwards compatibility.

TomasTomecek:
Okay then.
This puss is not happy about the compatibility breakage.

In Fedora lands, we will likely keep both versions for some time (to give everyone time to migrate their code bases).

shin-:
@TomasTomecek People will still be able to use 1.10.6 which is stable, and we're still considering whether or not to do some 1.x backports.

shin-:
> A `DockerClient` object is part of a graph of objects that represent different things managed by the Docker Engine

@aanand I'm not sure I follow how `DockerClient` is managed by the engine - the way I see it, it's an interface to the engine, but the engine doesn't have awareness of it as an entity.

bfirsh:
I think his point is that it _represents_ the Engine. But, perhaps we also `Client` just for disambiguation between the client and server halves of the Engine, and to make it clear that it is connecting to something over the network.

It does work kinda neatly, though:

```
import docker
engine = docker.from_env()
engine.containers.run(...)
```

But perhaps not neatly enough to justify the change from the common pattern of these things being called "clients".

Now it doesn't matter, here's another bike shed: `EngineClient`. ;)

PR #1313, opened 2016-11-27
Keep README in ASCII

glensc:
This avoids Python3 build errors when ran as non-UTF8 locale (C locale):

```
+ python3 setup.py build '--build-base=build-3'
  File "setup.py", line 36, in <module>
    long_description = readme_rst.read()
  File "/usr/lib64/python3.5/encodings/ascii.py", line 26, in decode
    return codecs.ascii_decode(input, self.errors)[0]
UnicodeDecodeError: 'ascii' codec can't decode byte 0xe2 in position 152: ordinal not in range(128)
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,7 +4,7 @@
 
 **Warning:** This readme is for the development version of docker-py, which is significantly different to the stable version. [Documentation for the stable version is here.](https://docker-py.readthedocs.io/)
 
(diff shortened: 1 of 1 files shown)


bfirsh:
Oh, gawd.

A better solution would be to use `--`, which is the ASCII equivalent of an em-dash. An even better solution would be for it to not break when reading utf-8, which will trip is again in the future. Does https://github.com/docker/docker-py/pull/1314 solve this issue for you?

glensc:
Yep, #1314 solves as well.

PR #1314, opened 2016-11-28
Specify encoding when loading readme

bfirsh:
Loading readme fails when system locale is not utf-8.

Potentially replaces #1313


shin-:
LGTM

PR #1319, opened 2016-11-30
Move config type creation from docker.utils functions to classes in docker.types

shin-:
Mostly internal API changes, but some of those are breaking (public) changes 
- `create_host_config` for some users who have not migrated to using the method through `APIClient`
- `create_ipam_config` and `create_ipam_pool`

Those functions now raise a `docker.errors.DeprecatedMethod` error pointing to the new classes.

This also fixes some egregious circular imports between `docker.types` and `docker.utils`


bfirsh:
:+1:

Test failures, though.

docker/api/network.py
```diff
@@ -47,7 +47,7 @@ def create_network(self, name, driver=None, options=None, ipam=None,
             driver (str): Name of the driver used to create the network
             options (dict): Driver options as a key-value dictionary
             ipam (dict): Optional custom IP scheme for the network.
-                Created with :py:meth:`~docker.utils.create_ipam_config`.
+                Created with :py:class:`~docker.types.IPAMConfig`.
```
bfirsh: This won't actually link to anywhere unless they are represented in a docs file somewhere.

docs/api.rst
```diff
@@ -49,15 +49,6 @@ Networks
   :members:
   :undoc-members:
 
-Utilities
```
bfirsh: Perhaps we could replace this with a "types" section to actually document them? (See my other comments.) Ditto the `HostConfig` object, and so on.

bfirsh:
Looks good apart from docs!

shin-:
Thanks Ben! Updated a lot of doc-related stuff and rebased, should look good now.

docker/types/services.py
```diff
@@ -10,19 +11,15 @@ class TaskTemplate(dict):
...
-      the service. See the :py:class:`~docker.types.services.DriverConfig`
-      class for details.
-    * resources (dict): Resource requirements which apply to each individual
-      container created as part of the service. See the
-      :py:class:`~docker.types.services.Resources` class for details.
-    * restart_policy (dict): Specification for the restart policy which applies
-      to containers created as part of this service. See the
-      :py:class:`~docker.types.services.RestartPolicy` class for details.
-    * placement (list): A list of constraints.
+        container_spec (ContainerSpec): Container settings for containers
```
bfirsh: A tip, but not important: these types can be links, like `` :py:class:`~docker.types.services.ContainerSpec` ``.

bfirsh:
LGTM! :tada:

PR #1322, opened 2016-11-30
replace http+docker special URIs with http:// URLs

graingert:
be compatable with requests 2.12.2


graingert:
in https://github.com/kennethreitz/requests/issues/3734 @lukasa recommends using "http" schemes rather than "http+<something" schemes.

shin-:
Let's hold off on that - https://github.com/kennethreitz/requests/issues/3735 suggests it's not resolved yet and I don't want to make hasty changes.

PR #1332, opened 2016-12-02
Rename all meaningful occurrences of docker-py to docker-sdk-python

shin-:
## DO NOT MERGE

This should only be merged at release time, and once the GitHub repo has been renamed.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,7 +1,7 @@
 # Contributing guidelines
 
 See the [Docker contributing guidelines](https://github.com/docker/docker/blob/master/CONTRIBUTING.md).
-The following is specific to docker-py.
+The following is specific to Docker SDK for Python.
 
 Thank you for your interest in the project. We look forward to your
 contribution. In order to make the process as fast and streamlined as possible,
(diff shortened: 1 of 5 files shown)


CONTRIBUTING.md
```diff
@@ -1,7 +1,7 @@
 # Contributing guidelines
 
 See the [Docker contributing guidelines](https://github.com/docker/docker/blob/master/CONTRIBUTING.md).
-The following is specific to docker-py.
+The following is specific to the Docker SDK for Python.
```
bfirsh: It's a proper noun, so should be just "Docker SDK for Python".

CONTRIBUTING.md
```diff
@@ -87,19 +88,18 @@ to reach out and ask questions. We will do our best to answer and help out.
 
 ## Development environment
 
-If you're looking contribute to docker-py but are new to the project or Python,
-here are the steps to get you started.
+If you're looking contribute to the Docker SDK for Python but are new to the
```
bfirsh: same - s/the//

shin-:
I have done a partial revert of my previous changes. Moving the GitHub repo and docs is too disruptive right now and we need to move forward with the release. There will still be time to make those renames after the dust has settled a bit.

TomasTomecek:
So the project will be renamed, is that correct? (I'm assuming we will need to do a rename in downstream distributions)

PR #1333, opened 2016-12-03
mkbuildcontext() takes file and directory list.

recap:
The new mkbuildcontext() has the ability to take a list of files or directories. These are added to the context archive in a flat file structure. This will enable the docker client.build() to find the files.

### DIFF
--- a/docker/utils/utils.py
+++ b/docker/utils/utils.py
@@ -11,12 +11,15 @@ import warnings
 from distutils.version import StrictVersion
 from datetime import datetime
 from fnmatch import fnmatch
+from os import listdir
 
 import requests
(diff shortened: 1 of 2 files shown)


ijc:
PR grooming: Thank you for your PR sorry it didn't get looked at for so long. Looking at the diff it appears to include a bunch of apparently unrelated changed and/or code motion which made it rather hard to review. However given the age of this PR rather than dig in to far I'm going to assume you are no longer interested and/or the functionality is no longer needed. In case I'm wrong please do reopen without the extra changes, it would also be best to explain the use case here since it isn't immediately apparent to me at least.

PR #1337, opened 2016-12-05
Add attachable field to network create

### LINKED ISSUE #1336: Support Attachable flag to network create

dnephin:
Fixes #1336

### DIFF
--- a/docker/api/network.py
+++ b/docker/api/network.py
@@ -38,7 +38,7 @@ class NetworkApiMixin(object):
     @minimum_version('1.21')
     def create_network(self, name, driver=None, options=None, ipam=None,
                        check_duplicate=None, internal=False, labels=None,
-                       enable_ipv6=False):
+                       enable_ipv6=False, attachable=None):
         """
         Create a network. Similar to the ``docker network create``.
 
@@ -54,6 +54,9 @@ class NetworkApiMixin(object):
             labels (dict): Map of labels to set on the network. Default
                 ``None``.
             enable_ipv6 (bool): Enable IPv6 on the network. Default ``False``.
+            attachable (bool): If enabled, and the network is in the global
+                scope,  non-service containers on worker nodes will be able to
+                connect to the network.
 
         Returns:
             (dict): The created network reference object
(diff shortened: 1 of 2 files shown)


docker/api/network.py
```diff
@@ -38,7 +38,7 @@ def networks(self, names=None, ids=None):
     @minimum_version('1.21')
     def create_network(self, name, driver=None, options=None, ipam=None,
                        check_duplicate=None, internal=False, labels=None,
-                       enable_ipv6=False):
+                       enable_ipv6=False, attachable=False):
```
shin-: Maybe we should have `attachable=None` here, and only accept either value (True/False) if API version >= 1.24?

docker/api/network.py
```diff
@@ -91,7 +94,8 @@ def create_network(self, name, driver=None, options=None, ipam=None,
             'Driver': driver,
             'Options': options,
             'IPAM': ipam,
-            'CheckDuplicate': check_duplicate
+            'CheckDuplicate': check_duplicate,
+            'Attachable': attachable,
```
shin-: Only add the Attachable field to data if it's != None and API version >= 1.24

shin-:
Couple comments based on my current understanding that this option was introduced with API 1.24.

dnephin:
Only the CS release had it in 1.24 I think, but I guess that's still more correct than no check.

docker/api/network.py
```diff
@@ -116,6 +119,12 @@ def create_network(self, name, driver=None, options=None, ipam=None,
                                      'supported in API version < 1.22')
             data['Internal'] = True
 
+        if attachable is not None
```
shin-: Missing a colon

shin-:
Hmm, I don't know what's the cause, but it doesn't look like the `Attachable` option is being registered by my Engine (1.13.0-rc2)

```
In [10]: net_id = c.create_network(name='helloworld', attachable=True)
In [12]: c.inspect_network('helloworld')['Attachable']
Out[12]: False
```

PR #1339, opened 2016-12-06
Minimum version 1.24 -> 1.21

shin-:
Provides a more reasonable range of backward compatibility.
This also matches the lowest version Compose is using (for v1 file format)


bfirsh:
LGTM

PR #1340, opened 2016-12-06
Add attachable option in create_network

### LINKED ISSUE #1336: Support Attachable flag to network create

shin-:
Picking up @dnephin 's work in #1337 

Resolves #1336

### DIFF
--- a/tests/helpers.py
+++ b/tests/helpers.py
@@ -73,4 +73,4 @@ def force_leave_swarm(client):
             if e.explanation == "context deadline exceeded":
                 continue
             else:
-                raise
+                return
--- a/tests/integration/models_swarm_test.py
+++ b/tests/integration/models_swarm_test.py
@@ -19,4 +19,9 @@ class SwarmTest(unittest.TestCase):
         assert client.swarm.leave(force=True)
         with self.assertRaises(docker.errors.APIError) as cm:
             client.swarm.reload()
-        assert cm.exception.response.status_code == 406
+        assert (
+            # FIXME: test for both until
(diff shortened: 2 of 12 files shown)


shin-:
@dnephin If you have a minute can you do a quick review of the changes here?

tests/helpers.py
```diff
@@ -73,4 +73,4 @@ def force_leave_swarm(client):
             if e.explanation == "context deadline exceeded":
                 continue
             else:
-                raise
+                return
```
dnephin: What's this change about?

tests/integration/models_swarm_test.py
```diff
@@ -19,4 +19,7 @@ def test_init_update_leave(self):
         assert client.swarm.leave(force=True)
         with self.assertRaises(docker.errors.APIError) as cm:
             client.swarm.reload()
-        assert cm.exception.response.status_code == 406
+        assert (
+            cm.exception.response.status_code == 406 or
+            cm.exception.response.status_code == 503
```
dnephin: Comment why it would be one or the other?

tests/unit/api_network_test.py
```diff
@@ -104,6 +104,10 @@ def test_create_network(self):
                     }
                 })
 
+    @requires_api_version('1.24')
+    def test_create_network_with_attachable(self):
+        pass
```
dnephin: Sorry I added this empty test stub. I think we can remove it now that there are other tests.

tests/helpers.py
shin-: It was overzealous - we don't really care if there's an error during the `tearDown`

PR #1341, opened 2016-12-06
Scope is added in volume after docker 1.12

pacoxu:
volume scope is included after docker 1.12.

According to remote API 1.24, the return of creating volume is like below:

> HTTP/1.1 201 Created
Content-Type: application/json
{
  "Name": "tardis",
  "Driver": "custom",
  "Mountpoint": "/var/lib/docker/volumes/tardis",
  "Status": {
    "hello": "world"
  },
  "Labels": {
    "com.example.some-label": "some-value",
    "com.example.some-other-label": "some-other-value"
  },
  "Scope": "local"
}

### DIFF
--- a/tests/unit/api_volume_test.py
+++ b/tests/unit/api_volume_test.py
@@ -89,6 +89,16 @@ class VolumeTest(BaseAPIClientTest):
                 'perfectcherryblossom', driver_opts=''
             )
 
+    @requires_api_version('1.24')
+    def test_create_volume_with_no_specified_name(self):
+        result = self.client.create_volume(name=None)
+        self.assertIn('Name', result)
+        self.assertNotEqual(result['Name'], None)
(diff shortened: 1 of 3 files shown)


pacoxu:
ut not passed, so close it and try to fix it

tests/unit/api_volume_test.py
```diff
@@ -89,6 +89,17 @@ def test_create_volume_invalid_opts_type(self):
                 'perfectcherryblossom', driver_opts=''
             )
 
+    @requires_api_version('1.24')
+    def test_create_volume_with_no_specified_name(self):
+        no_specified_name = None
```
shin-: You can just use `None` directly.

pacoxu:
@shin-  commits are squashed and use 'None' directly.

PR #1342, opened 2016-12-07
Add Jenkinsfile for integration tests matrix

shin-:
@mikedougherty Can you take a look at this when you have some time? I'm not too familiar with the syntax so there might be some dumb things.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -4,7 +4,7 @@ def imageNameBase = "dockerbuildbot/docker-py"
 def imageNamePy2
 def imageNamePy3
 def images = [:]
-def dockerVersions = ["1.12.3", "1.13.0-rc3"]
+def dockerVersions = ["1.12.0", "1.13.0-rc3"]
 
 def buildImage = { name, buildargs, pyTag ->
   img = docker.image(name)
@@ -33,6 +33,7 @@ def buildImages = { ->
 
 def runTests = { Map settings ->
   def dockerVersion = settings.get("dockerVersion", null)
+  def pythonVersion = settings.get("pythonVersion", null)
   def testImage = settings.get("testImage", null)
 
   if (!testImage) {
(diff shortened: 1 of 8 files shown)


jenkinsfile
```diff
@@ -0,0 +1,56 @@
...
+
+
+buildImage()
+parallel(
+  failFast: true,
+  py2_docker_1_13: runTests(pythonVersion: "py2", dockerVersion: "1.13.0-rc3"),
+  py3_docker_1_13: runTests(pythonVersion: "py3", dockerVersion: "1.13.0-rc3"),
+  py2_docker_1_12: runTests(pythonVersion: "py2", dockerVersion: "1.12.3"),
+  py3_docker_1_12: runTests(pythonVersion: "py3", dockerVersion: "1.12.3"),
+)
```
mikedougherty: nothing wrong with this as is, but long term it's probably better written with 2 lists of `pyVersion` and `dockerVersion` and 2x `for` loops

jenkinsfile
```diff
@@ -0,0 +1,56 @@
...
+  }
+  if (!dockerVersion) {
+    throw new Exception("Need Docker version to test, e.g.: `runTests(pythonVersion: 'py3')`")
+  }
+
+  { ->
+    wrappedNode(label: "ubuntu && !zfs", cleanWorkspace: true) {
+      stage("test python=${pythonVersion} / docker=${dockerVersion}") {
+        checkout(scm)
+        sh """docker run -d --name dpy-dind-$BUILD_NUMBER -v /tmp --privileged \\
```
mikedougherty: also you can merge the two `sh` calls into one long string if you want.

jenkinsfile
```diff
@@ -0,0 +1,56 @@
...
+          dockerswarm/dind:${dockerVersion} docker daemon -H tcp://0.0.0.0:2375
+        """
+        sh """docker run \\
```
mikedougherty: kind of external to the Jenkinsfile syntax/semantics; but it could be an improvement to have a docker-compose file to manage the container inter-dependencies instead with `--link`; etc. \</sidebar>

jenkinsfile
```diff
@@ -0,0 +1,56 @@
+#!groovy
+
+def imageNameBase = "dockerbuildbot/docker-py"
+def imageNamePy2 = "${imageNameBase}:py2-${gitCommit()}"
+def imageNamePy3 = "${imageNameBase}:py3-${gitCommit()}"
+
+
+def buildImage = { ->
+  wrappedNode(label: "ubuntu && !zfs", cleanWorkspace: true) {
```
shin-: dind doesn't start properly when inheriting ZFS from the host, IIRC.

shin-:
@mikedougherty Thanks! Took some time to read up on Groovy this morning too. Does this look decent?

jenkinsfile
```diff
@@ -0,0 +1,71 @@
...
+}
+
+
+buildImages()
+
+def testMatrix = [failFast: true]
+
+for (img in images) {
+  for (version in dockerVersions) {
+    testMatrix["${img.key}_${version}"] = [testImage: img.value, dockerVersion: version]
```
mikedougherty: i think the right side of this might be missing `runTests(...)`

jenkinsfile
```diff
@@ -0,0 +1,71 @@
...
+        sh """docker run -d --name dpy-dind-\$BUILD_NUMBER -v /tmp --privileged \\
+          dockerswarm/dind:${dockerVersion} docker daemon -H tcp://0.0.0.0:2375
+        """
+        sh """docker run \\
+          --name dpy-tests-\$BUILD_NUMBER --volumes-from dpy-dind-\$BUILD_NUMBER \\
+          -e 'DOCKER_HOST=tcp://docker:2375' \\
+          --link=dpy-dind-\$BUILD_NUMBER:docker \\
+          ${testImage.id} \\
+          py.test -rxs tests/integration
+        """
```
mikedougherty: there should be some kind of cleanup script to remove the containers at the end. i would recommend a `try { .. } finally { ... }` with the two existing `sh` steps in the try, and the necessary `docker stop` / `docker rm` calls in the finally block.

shin-:
Thanks - comments addressed :+1:

PR #1344, opened 2016-12-07
Dramatically increase get logs performance

psviderski:
The `APIClient` does a lot of unnecessary bytes object copies while demultiplexing logs buffer with `stream=False`: https://github.com/docker/docker-py/blob/master/docker/api/client.py#L310-L312. This PR dramatically improves the performance of `_multiplexed_buffer_helper`.


shin-:
LGTM - Thank you!

PR #1355, opened 2016-12-15
Return back something useful during load_image.

huzhengchuan:
In docker 1.12.0, During a `docker load` there will say something
useful back(https://github.com/docker/docker/pull/23377/commits).
In docker-py, this is also needed because I have lots of images
and after a recent docker load I had no idea what image
I just imported.

Signed-off-by: Zhengchuan.hu &lt;zhengchuan.hu@easystack.cn>

### DIFF
--- a/docker/api/image.py
+++ b/docker/api/image.py
@@ -271,8 +271,8 @@ class ImageApiMixin(object):
         Args:
             data (binary): Image data to be loaded.
         """
-        res = self._post(self._url("/images/load"), data=data)
-        self._raise_for_status(res)
+        return self._result(
+            self._post(self._url("/images/load"), data=data))
 
     def pull(self, repository, tag=None, stream=False,
              insecure_registry=False, auth_config=None, decode=False):


docker/api/image.py
```diff
@@ -271,8 +271,8 @@ def load_image(self, data):
         Args:
             data (binary): Image data to be loaded.
         """
```
shin-: Docstring should be updated with a `Returns:` section.

docker/api/image.py
```diff
@@ -271,8 +271,8 @@ def load_image(self, data):
...
-        res = self._post(self._url("/images/load"), data=data)
-        self._raise_for_status(res)
+        return self._result(
```
shin-: We may not want to do this for API < 1.24 depending on what the content is befor engine 1.12.

docker/api/image.py
huzhengchuan: Thank you very much, I will update the Docstring and integration tests.

shin-:
Closing in favor of #1657

PR #1362, opened 2016-12-19
Add filters option to NetworkApiMixin.networks

qazbnm456:
I've noticed that Docker Remote API has filters option to `/networks` [here](https://docs.docker.com/engine/reference/api/docker_remote_api_v1.24/#/networks), but docker-py doesn't support it [here](https://github.com/docker/docker-py/blob/master/docker/api/network.py#L10), however. Therefore, I add such a filters option in this PR.

Would be great if we can have this 😄

### DIFF
--- a/docker/api/network.py
+++ b/docker/api/network.py
@@ -1,19 +1,23 @@
-import json
-
 from ..errors import InvalidVersion
 from ..utils import check_resource, minimum_version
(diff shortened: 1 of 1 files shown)


shin-:
Thanks! Can you squash those commits together?

qazbnm456:
@shin- Sure!

PR #1363, opened 2016-12-19
Add cpuset_mems to HostConfig

qazbnm456:
I've noticed that Docker Remote API has CpusetMems option to /containers/create [here](https://docs.docker.com/engine/reference/api/docker_remote_api_v1.24/#/create-a-container) since API version 1.19, but docker-py doesn't support it [here](https://github.com/docker/docker-py/blob/master/docker/types/containers.py#L103), however. Therefore, I add such an option in this PR.

Would be great if we can have this. 😄


qazbnm456:
@shin- Ok, I will add a test later. 😃

delcypher:
This can probably be closed. #1583 implemented this and has already been merged into `master`.

Issue #1371, opened 2016-12-29
mount ReadOnly option is ignored by service.create


dzimine:
It says it all: code is missed to assign it.
```
docker.types.Mount('foo','bar', read_only=True)
{'Source': 'bar', 'Target': 'foo', 'Type': 'volume'}
m['ReadOnly']
KeyError: 'ReadOnly'
```

I'll submit a patch.

PR #1372, opened 2016-12-29
Fix readonly in mounts.

dzimine:
Fixes: #1371 - properly pass `ro` and `rw` to mount parameters for swarm services.

Signed-off-by: Dmitri Zimine <dz@stackstorm.com>

### DIFF
--- a/tests/unit/types.py
+++ b/tests/unit/types.py
@@ -1,0 +1,16 @@
+import unittest
+from docker.types.services import Mount
+
+
+class TestMounts(unittest.TestCase):
+    def test_parse_mount_string_docker(self):
+        mount = Mount.parse_mount_string("foo/bar:/buz:ro")
+        self.assertEqual(mount['Source'], "foo/bar")
+        self.assertEqual(mount['Target'], "/buz")
+        self.assertEqual(mount['ReadOnly'], True)
+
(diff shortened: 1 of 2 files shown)


dzimine:
Hey @GordonTheTurtle the robot, commit is signed and amended, as requested.

dzimine:
Hi folks, happy new year; 
how can I remove 'dco-signed' label now that the commit is signed? 
Anything more I have to do to have this reviewed?

tests/unit/types.py
```diff
@@ -0,0 +1,16 @@
+import unittest
+from docker.types.services import Mount
+
+
+class TestMounts(unittest.TestCase):
```
shin-: Can you add this to `dockertypes_test.py` instead?

tests/unit/types.py
```diff
@@ -0,0 +1,16 @@
...
+    def test_parse_mount_string_docker(self):
```
shin-: Please make this into 3 separate tests (3 methods) instead

dzimine:
Sorry I missed the memo. Will update in a separate pull-request (I want to add a new `service` mount syntax support parser so I'll fix the tests once I get to it). Thanks!

PR #1393, opened 2017-01-16
Fix #1351

### LINKED ISSUE #1351: Exception retrieving untagged images on api >= 1.24
Docker API >= 1.24 will return a null object if image tags DNE instead of not including it in the response. This makes the dict.get fail to catch the null case and the list comprehension to iterate over a non-iterable.

```
  File "docker/models/images.py", line 16, in __repr__
    return "<%s: '%s'>" % (self.__class__.__name__, "', '".join(self.tags))
  File "docker/models/images.py", line 34, in tags
    tag for tag in self.attrs.get('RepoTags', [])
TypeError: 'NoneType' object is not iterable
```

This is similar to an issue seen in [salt](https://github.com/saltstack/salt/pull/35447/commits/b833b5f9587534d3b843a026ef91abc4ec929d0f)

Was able to get things working with a pretty quick change:
```
diff --git a/docker/models/images.py b/docker/models/images.py
index 32068e6..39a640d 100644
--- a/docker/models/images.py
+++ b/docker/models/images.py
@@ -30,9 +30,11 @@ class Image(Model):
         """
         The image's tags.
         """
+        tags = self.attrs.get('RepoTags', [])
+        if tags is None:
+            return []
         return [
-            tag for tag in self.attrs.get('RepoTags', [])
-            if tag != '<none>:<none>'
+            tag for tag in tags if tag != '<none>:<none>'
         ]
 
     def history(self):
```

aebm:
* Fix TypeError when getting the tags property from an image that has
  no tags. Ex: An image pulled by cryptohash. It is handled like when
  the image doesn't have defined the RepoTags member.

Signed-off-by: Alejandro E. Brito Monedero <alejandro.monedero@gmail.com>

#1351

### DIFF
--- a/docker/models/images.py
+++ b/docker/models/images.py
@@ -30,10 +30,10 @@ class Image(Model):
         """
         The image's tags.
         """
-        return [
-            tag for tag in self.attrs.get('RepoTags', [])
-            if tag != '<none>:<none>'
-        ]
+        tags = self.attrs.get('RepoTags')
+        if tags is None:
+            tags = []
+        return [tag for tag in tags if tag != '<none>:<none>']
 
     def history(self):
         """
--- a/tests/unit/models_images_test.py
+++ b/tests/unit/models_images_test.py
@@ -83,6 +83,11 @@ class ImageTest(unittest.TestCase):
         })
         assert image.tags == []
 
+        image = Image(attrs={
+            'RepoTags': None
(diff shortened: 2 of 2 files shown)


bfirsh:
LGTM

shin-:
Thank you!

PR #1404, opened 2017-01-22
allow ipv6 :: notation in split_port

### LINKED ISSUE #1361: IPv6 port binding error with docker composer
Hi,

There is a issue by using IPv6 by port/publishing in docker-compose file see for more info also:[#2663](https://github.com/docker/compose/issues/2663)

How reproduce:
1. Create docker-compose with for example:
```
version: '2.1'

services:
  web:
    image: 'tutum/hello-world'
    ports:
      - '[2a01:7c8:fffb:31::1]:80:80'
```
2. docker-compose up
And you get the follow error:
```
ERROR: The Compose file './docker-compose.yml' is invalid because:
services.web.ports is invalid: Invalid port "[2a01:7c8:fffb:31::1]:80:80", should be [[remote_ip:]remote_port[-remote_port]:]port[/protocol]
```

Version information:
```
docker-compose version 1.9.0, build 2585387
docker-py version: 1.10.6
CPython version: 2.7.9
OpenSSL version: OpenSSL 1.0.1t  3 May 2016
```

chrisv2:
fixes #1361 and docker/compose#2663, and allows to use IPv6 in compose port mappings.

### DIFF
--- a/docker/utils/ports.py
+++ b/docker/utils/ports.py
@@ -1,4 +1,7 @@
 
+import re
+
+
 def add_port_mapping(port_bindings, internal_port, external):
     if internal_port in port_bindings:
         port_bindings[internal_port].append(external)
@@ -57,7 +60,11 @@ def _raise_invalid_port(port):
 
 
 def split_port(port):
+    # mask : inside [] with ^ (ipv6 address)
+    port = re.sub(r'\[[^]]*\]', lambda x: x.group(0).replace(':', '^'), port)
     parts = str(port).split(':')
+    # unmask parts
+    parts = [part.replace('^', ':') for part in parts]
 
     if not 1 <= len(parts) <= 3:
         _raise_invalid_port(port)
--- a/docker/utils/ports.py
+++ b/docker/utils/ports.py
@@ -1,6 +1,4 @@
 
-import re
-
 
 def add_port_mapping(port_bindings, internal_port, external):
     if internal_port in port_bindings:
@@ -60,11 +58,7 @@ def _raise_invalid_port(port):
 
 
 def split_port(port):
-    # mask : inside [] with ^ (ipv6 address)
-    port = re.sub(r'\[[^]]*\]', lambda x: x.group(0).replace(':', '^'), port)
-    parts = str(port).split(':')
-    # unmask parts
-    parts = [part.replace('^', ':') for part in parts]
+    parts = str(port).rsplit(':',2)
 
     if not 1 <= len(parts) <= 3:
         _raise_invalid_port(port)
--- a/tests/unit/utils_test.py
+++ b/tests/unit/utils_test.py
@@ -547,6 +547,12 @@ class PortsTest(unittest.TestCase):
         self.assertEqual(external_port,
                          [("127.0.0.1", "1000"), ("127.0.0.1", "1001")])
 
+    def test_split_port_with_ipv6_address(self):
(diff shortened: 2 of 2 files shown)


chrisv2:
updated the PR with a much simpler solution (which happens to work, actually)

chrisv2:
okay, that breaks `PortsTest.test_split_port_invalid`:

```

______________________ PortsTest.test_split_port_invalid _______________________

tests/unit/utils_test.py:558: in test_split_port_invalid

    lambda: split_port("0.0.0.0:1000:2000:tcp"))

E   AssertionError: ValueError not raised

========== 1 failed, 451 passed, 3 skipped, 1 xfailed in 2.82 seconds ==========

ERROR: InvocationError: '/home/travis/build/docker/docker-py/.tox/py27/bin/py.test --cov=docker tests/unit'
```

I think we should change the test in this case, as it tests for an odd corner case which would be somewhat difficult to satisfy, with no apparent advantage. Opinions?

lenovouser:
Maybe use the `[1234:1234:1234:1234::]:22:22` format?

lenovouser:
Any progress?

lenovouser:
Hello? 😞

chrisv2:
@lenovouser We can't pass v6 IP in square bracket ([x:y::z]) notation to docker (while the engine accepts it, it still listens on 0.0.0.0). So we had to remove them in our code, which would of course be possible, but lead to Python API using a different notation than docker itself, which seems a bad idea to me.

I still believe that the unit test should be changed.

lenovouser:
@chrisv2 I understand and don't mind that in particular. I rather want this to be merged so that I can use docker-compose with IPv6 😆

lenovouser:
Can this be merged?

lenovouser:
😞

lenovouser:
😞 @shin- do you know who needs to review this? I really need https://github.com/docker/compose/issues/2663

SethMichaelLarson:
@lenovouser Remember that most people in Open Source are volunteers and we need to respect their time. There are failing tests that haven't been fixed in the CI. If you want to expedite the merge of this change perhaps you can open your own PR with this change.

lenovouser:
@SethMichaelLarson there are questions to be resolved by the maintainers before any further changes can be made. @chrisv2 explicitly stated that:

> I think we should change the test in this case, as it tests for an odd corner case which would be somewhat difficult to satisfy, with no apparent advantage. Opinions?

SethMichaelLarson:
I'm aware that that is the test being mentioned. I'm suggesting that it may be more productive for you to open your own PR with the changes rather than pinging the same issue over and over if the change is really important or time-sensitive to you.

lenovouser:
@SethMichaelLarson I understand what you mean and I would do that if there was something to change and then create a PR. But the info if or what to change needs to come from the maintainers. I can't read peoples minds - my guess is that they just haven't even seen this PR and it got lost in all the others.

kaiyou:
After trying to wrap my head around the commits and why the tests fail, it appears to me that `split_port` and `to_port_range` could (should ?) be written differently. Generating a port range should somehow be at a lower level than deciding if ports in that range are TCP or UDP.

As I understand it, this is only used when creating containers, wouldn't it be simpler and safer to user regular expressions in that case?

shin-:
I'm closing this in favor of #1632, but still want to thank you for submitting this PR!

Issue #1400, opened 2017-01-19
Docker build output is not iterable


appendjeff:
```
from io import BytesIO
import docker
dockerfile = '''
FROM ubuntu
RUN apt-get update
'''
f = BytesIO(dockerfile.encode('utf-8'))
client = docker.from_env()
for line in client.images.build(fileobj=f, rm=True, tag='test', nocache=True, stream=True):
    print(line)
```

From the documentation, it seems like I client.images.build should return a generator. I get the following traceback:
```
Traceback (most recent call last):
  File "issue_no_logs.py", line 9, in <module>
    for line in client.images.build(fileobj=f, rm=True, tag='test', nocache=True, stream=True):
TypeError: 'Image' object is not iterable
```

```
pip freeze | grep docker && python --version && docker version
docker==2.0.1
docker-pycreds==0.2.1
Python 2.7.12
Client:
 Version:      1.12.3
 API version:  1.24
 Go version:   go1.6.2
 Git commit:   6b644ec
 Built:        Mon, 19 Dec 2016 09:20:48 +1300
 OS/Arch:      linux/amd64
Server:
 Version:      1.12.3
 API version:  1.24
 Go version:   go1.6.2
 Git commit:   6b644ec
 Built:        Mon, 19 Dec 2016 09:20:48 +1300
 OS/Arch:      linux/amd64

```

```
cat /etc/os-release

NAME="Ubuntu"
VERSION="16.04.1 LTS (Xenial Xerus)"
ID=ubuntu
ID_LIKE=debian
PRETTY_NAME="Ubuntu 16.04.1 LTS"
VERSION_ID="16.04"
HOME_URL="http://www.ubuntu.com/"
SUPPORT_URL="http://help.ubuntu.com/"
BUG_REPORT_URL="http://bugs.launchpad.net/ubuntu/"
VERSION_CODENAME=xenial
UBUNTU_CODENAME=xenial
```

shin-:
The returned type `Image` matches the [documentation](http://docker-py.readthedocs.io/en/stable/images.html#docker.models.images.ImageCollection.build), as far as I can tell.  

Where did you see that `client.images.build` returns a generator? Might be an overlook on our part in another section of the docs.

appendjeff:
@shin- I appreciate the quick response. From the documentation, I saw stream as a kwarg.
> stream (bool) – Deprecated for API version > 1.8 (always True). Return a blocking generator you can iterate > over to retrieve build output as it happens


[...]

PR #1420, opened 2017-01-26
Improve docs for streaming build logs

appendjeff:
Signed-off-by: Jeff Mandell <jeffamandell@gmail.com>

Fixes https://github.com/docker/docker-py/issues/1400


appendjeff:
@shin- can you review this tiny change?

shin-:
Actually, the direction I'd rather see us go in is to
1. Remove the mention of the `stream` param in that docstring, since it's a no-op.
2. In another update, add something to the `DockerClient` that lets one retrieve build logs (either a param when calling build or maybe a different method

Issue #1422, opened 2017-01-27
put_archive - must specify all data options


vmindru:
put_archive is using requests.Session.put in the background, the doc must state all options that can be passed as data.

from http://docs.python-requests.org/en/master/api/

`data -- (optional) Dictionary, bytes, or file-like object to send in the body of the Request.`

ijc:
Looks like this proposal was rejected in https://github.com/docker/docker-py/pull/1423#issuecomment-275771000, closing.

PR #1423, opened 2017-01-27
1422: amend docs for put_archive, provide missing data details

vmindru:
#1422 pr for

### DIFF
--- a/docker/api/container.py
+++ b/docker/api/container.py
@@ -900,7 +900,7 @@ class ContainerApiMixin(object):
             container (str): The container where the file(s) will be extracted
             path (str): Path inside the container where the file(s) will be
                 extracted. Must exist.
-            data (bytes): tar data to be extracted
+            data: Dictionary, bytes, or file-like object to be extracted
 
         Returns:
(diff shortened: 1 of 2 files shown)


vmindru:
@GordonTheTurtle  what is the dco/no label?

vmindru:
I think Jenkins failure is a bit strange, give both the error message and the type of changes this patch brings. 

VM

shin-:
Thank you for your contribution! However, I feel like that change will make our documentation more confusing, whereas the current version at least provides a clear, supported way to use this method.

As far as the test failures are concerned, I'm investigating, but I agree that they're obviously not a result of your changes :)

vmindru:
Hold on :) one would have no idea that he can pass an iterable object instead of simple bytes. One would have to check the source code find out you use requests in the background and figure out that he can pass an iterable object, maybe you folks give it a second thought before you close this?

ns-cweber:
I came here specifically to figure out why `put_archive()` claims that it only supports `bytes` when it apparently supports iterable objects too?

PR #1424, opened 2017-01-28
Add new (1.25) service params

shin-:
- `max_failure_ratio`, `monitor` in `UpdateConfig`
- `force_update` in `TaskTemplate`


bfirsh:
Documentation for the model needs updating too, doesn't it?

shin-:
@bfirsh Those params aren't referenced directly by the models, so documentation only needs to be updated on the classes themselves

Issue #1380, opened 2017-01-07
docker-py v2.0.0  Can't create container data volumes using client.containers.run & create


pommetjehorlepiep:
[https://docker-py.readthedocs.io/en/latest/containers.html](url): **run** method: for **volumes** it says:
```
volumes (dict or list) –

A dictionary to configure volumes mounted inside the container. The key is either the host path or a volume name, and the value is a dictionary with the keys:

    bind The path to mount the volume inside the container
    mode Either rw to mount the volume read/write, or ro to mount it read-only.

For example:

{'/home/user1/': {'bind': '/mnt/vol2', 'mode': 'rw'},
 '/var/www': {'bind': '/mnt/vol1', 'mode': 'ro'}}
```

This effectively means that you can't create container data volumes, since you either have to create a volume 1st using **docker create volume** or use a host directory.
There is no option to just specify a container folder.

The  low-level API ([https://docker-py.readthedocs.io/en/latest/api.html#module-docker.api.container](url)) **create_container** still allows you to do this (since I was able to do this using the v1 of docker-py

From the **docker run** documentation [https://docs.docker.com/engine/reference/commandline/run/](url):
```
  -v, --volume value                Bind mount a volume (default []). The format
                                    is `[host-src:]container-dest[:<options>]`.
                                    The comma-delimited `options` are [rw|ro],
                                    [z|Z], [[r]shared|[r]slave|[r]private], and
                                    [nocopy]. The 'host-src' is an absolute path
                                    or a name value.
```

Am I correct here that logic to support container data volumes is missing from **client.containers.run & create**?
Or is there another way how to do this? Or are container data volumes being replaced by data volumes?

shin-:
Thank you for the report. It looks like this is an issue we overlooked.

cc @bfirsh It seems like we're passing the `volumes` data to `host_config` no matter what, so even using the list syntax (`volumes=['/hello/world:rw']`) will result in an error from the engine because we're passing an invalid bind mount spec.

RobbieClarken:
I believe there is also a bug with named volumes not being supported by `DockerClient.containers.run`:

```
>>> print(docker.__version__)
2.0.1
>>> client.containers.run('alpine', 'ls', volumes={'my-named-volume': {'bind': '/data', 'mode': 'ro'}})
Traceback (most recent call last):
[...]
docker.errors.APIError: 500 Server Error: Internal Server Error for url: http+docker://localunixsocket/v1.24/containers/create
[...]

PR #1439, opened 2017-02-07
Fix various volume bugs

bfirsh:
Fixes both issues mentioned in #1380.

Technically we shouldn't be passing volumes without a host path as binds, but the daemon doesn't seem to mind.

### DIFF
--- a/docker/api/container.py
+++ b/docker/api/container.py
@@ -313,9 +313,10 @@ class ContainerApiMixin(object):
 
         **Using volumes**
 
-        Volume declaration is done in two parts. Provide a list of mountpoints
-        to the with the ``volumes`` parameter, and declare mappings in the
-        ``host_config`` section.
+        Volume declaration is done in two parts. Provide a list of
+        paths to use as mountpoints inside the container with the
+        ``volumes`` parameter, and declare mappings from paths on the host
+        in the ``host_config`` section.
 
         .. code-block:: python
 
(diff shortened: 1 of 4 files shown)


docker/api/container.py
```diff
@@ -392,7 +393,8 @@ def create_container(self, image, command=None, hostname=None, user=None,
                 version 1.10. Use ``host_config`` instead.
             dns_opt (:py:class:`list`): Additional options to be added to the
                 container's ``resolv.conf`` file
-            volumes (str or list):
+            volumes (str or list): List of paths inside the container to use
```
shin-: I don't think `str` is a valid type here, is it?

docker/models/containers.py
```diff
@@ -885,5 +885,15 @@ def _create_container_args(kwargs):
                                   for p in sorted(port_bindings.keys())]
     binds = create_kwargs['host_config'].get('Binds')
     if binds:
-        create_kwargs['volumes'] = [v.split(':')[0] for v in binds]
+        create_kwargs['volumes'] = [_host_volume_from_bind(v) for v in binds]
     return create_kwargs
+
+
+def _host_volume_from_bind(bind):
+    bits = bind.split(':')
```
shin-: This is broken in the case of Windows host paths

docker/api/container.py
shin-: ah, nevermind, we just add it to a one-element list if it's a string. Fine by me.

bfirsh:
In retrospect I think we should have dropped support for the string format in 2.0. ;)

Perhaps we can phase it out over a few releases...

PR #1445, opened 2017-02-08
Remove wrong url in docker/transport/npipeconn.py

xulike666:
well,no repo called docker-sdk-python existed
Signed-off-by: Aaron.L.Xu <likexu@harmonycloud.cn>


shin-:
We will eventually rename the `docker-py` github repo. This was a pre-emptive change before we decided to delay the move. It's not critical, so I'll leave it like this for now.

PR #1458, opened 2017-02-16
Add init parameter to container HostConfig

tmad:
Docker introduced "init". This allows to run an init inside the container that forwards signals and reaps processes.

It would be nice if docker-py supported this one.

### DIFF
--- a/docker/types/containers.py
+++ b/docker/types/containers.py
@@ -118,7 +118,7 @@ class HostConfig(dict):
                  tmpfs=None, oom_score_adj=None, dns_opt=None, cpu_shares=None,
                  cpuset_cpus=None, userns_mode=None, pids_limit=None,
                  isolation=None, auto_remove=False, storage_opt=None,
-                 init=None):
+                 init=None, init_path=None):
 
         if mem_limit is not None:
             self['Memory'] = parse_bytes(mem_limit)
@@ -422,6 +422,11 @@ class HostConfig(dict):
             if version_lt(version, '1.25'):
                 raise host_config_version_error('init', '1.25')
             self['Init'] = init
+
+        if init_path:
+            if version_lt(version, '1.25'):
+                raise host_config_version_error('init', '1.25')
+            self['InitPath'] = init_path
 
 
 def host_config_type_error(param, param_value, expected):
--- a/docker/types/containers.py
+++ b/docker/types/containers.py
@@ -117,7 +117,8 @@ class HostConfig(dict):
                  oom_kill_disable=False, shm_size=None, sysctls=None,
                  tmpfs=None, oom_score_adj=None, dns_opt=None, cpu_shares=None,
                  cpuset_cpus=None, userns_mode=None, pids_limit=None,
-                 isolation=None, auto_remove=False, storage_opt=None):
+                 isolation=None, auto_remove=False, storage_opt=None,
+                 init=None):
 
         if mem_limit is not None:
             self['Memory'] = parse_bytes(mem_limit)
@@ -416,6 +417,11 @@ class HostConfig(dict):
             if version_lt(version, '1.24'):
                 raise host_config_version_error('storage_opt', '1.24')
             self['StorageOpt'] = storage_opt
+
+        if init is not None:
+            if version_lt(version, '1.25'):
+                raise host_config_version_error('init', '1.25')
+            self['Init'] = init
 
 
 def host_config_type_error(param, param_value, expected):
--- a/tests/integration/api_container_test.py
+++ b/tests/integration/api_container_test.py
@@ -450,6 +450,18 @@ class CreateContainerTest(BaseAPIIntegrationTest):
         config = self.client.inspect_container(ctnr)
         assert config['HostConfig']['Init'] is True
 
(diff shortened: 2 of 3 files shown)


docker/api/container.py
```diff
@@ -238,7 +238,7 @@ def create_container(self, image, command=None, hostname=None, user=None,
                          memswap_limit=None, cpuset=None, host_config=None,
                          mac_address=None, labels=None, volume_driver=None,
                          stop_signal=None, networking_config=None,
-                         healthcheck=None, stop_timeout=None):
+                         healthcheck=None, stop_timeout=None, init=None):
```
shin-: I believe this needs to be an argument in `create_host_config`, not `create_container`

shin-: I believe this needs to be an argument in `create_host_config`, not `create_container`

tests/integration/api_container_test.py
```diff
@@ -422,6 +422,23 @@ def test_create_with_stop_timeout(self):
         config = self.client.inspect_container(container)
         assert config['Config']['StopTimeout'] == 25
 
+    @requires_api_version('1.26')
```
shin-: Should be `requires_api_version('1.25')`

docker/types/containers.py
```diff
@@ -412,6 +412,11 @@ def __init__(self, version, binds=None, port_bindings=None,
                 raise host_config_version_error('auto_remove', '1.25')
             self['AutoRemove'] = auto_remove
 
+        if init:
+            if version_lt(version, '1.26'):
```
shin-: Minimum version is `1.25`

docker/types/containers.py
```diff
@@ -438,7 +443,7 @@ def __init__(
         working_dir=None, domainname=None, memswap_limit=None, cpuset=None,
         host_config=None, mac_address=None, labels=None, volume_driver=None,
         stop_signal=None, networking_config=None, healthcheck=None,
-        stop_timeout=None
+        stop_timeout=None, init=None
```
shin-: This needs to be removed, Init is not part of ContainerConfig

tests/integration/api_container_test.py
```diff
@@ -422,6 +422,23 @@ def test_create_with_stop_timeout(self):
...
+    def test_create_with_init(self):
+        ctnr = self.client.create_container(
+            BUSYBOX, 'true',
+            host_config=self.client.create_host_config(
+                init='true'
```
shin-: 'true' is not a valid boolean. This test would fail, but the decorator prevents it from being run in the first place.

tests/integration/api_container_test.py
```diff
@@ -422,6 +422,23 @@ def test_create_with_stop_timeout(self):
...
+            )
+        )
+        self.assertIn('Id', ctnr)
```
shin-: Unnecessary

tests/integration/api_container_test.py
```diff
@@ -422,6 +422,23 @@ def test_create_with_stop_timeout(self):
...
+        self.tmp_containers.append(ctnr['Id'])
+        self.client.start(ctnr)
+        inspect = self.client.inspect_container(ctnr)
+
+        self.assertIn('HostConfig', inspect)
```
shin-: Unnecessary

tests/integration/api_container_test.py
```diff
@@ -422,6 +422,23 @@ def test_create_with_stop_timeout(self):
...
+        host_config = inspect['HostConfig']
+        self.assertIn('init', host_config)
```
shin-: It's not enough to test the presence of the item,  you need to test the value you provide is the same as the value you're getting back.

shin-: Additionally, the name of the field is `Init`, capitalized.

docker/api/container.py
tmad: Fixed, this was unnecessary.

docker/types/containers.py
tmad: Done

tests/integration/api_container_test.py
tmad: My bad.

tests/integration/api_container_test.py
tmad: I've changed the test, could you verify now?

docker/types/containers.py
```diff
@@ -417,6 +418,11 @@ def __init__(self, version, binds=None, port_bindings=None,
                 raise host_config_version_error('storage_opt', '1.24')
             self['StorageOpt'] = storage_opt
 
+        if init:
```
shin-: Minor: this should be `if init is not None:`. If the user explicitly specifies `init=False`, we still want to enter that branch.

tmad: Sure, fixed

tmad:
I've fixed issues, that you've pointed out.
I've also added support for init_path parameter, as suggested.

shin-:
Rebased and merged manually - 13b5f785a7ab459960aae82fae00e4245e391387 8c6534d7be9130c7889164cd5ec054cb1e051569

Thank you!

PR #1467, opened 2017-02-19
Allow port range like 8000-8010:80

crierr:
Docker recognizes (host port range):(container port) syntax and determines available port in range randomly.
docker-py raised the error 'Port ranges don't match in length'.
It would be very useful to allow it, especially for using docker-compose.


shin-:
Thank you!

PR #1473, opened 2017-02-21
Allow events daemon command to read config.json

### LINKED ISSUE #1472: DaemonApiMixin.events does not propagate HttpHeaders from config.json
The docker.api.daemon.DaemonApiMixin.events does not make use of the config.json, which could have custom HTTP headers to pass to the server.

Anvil:
Events units test have also been updated/fixed.
This should fix #1472


Anvil:
You're welcome. :]

PR #1490, opened 2017-03-01
Update assert in test_create_with_restart_policy

ripcurld0:
In https://github.com/docker/docker/pull/30870 a new error message is displayed if the container is restarting.

To make `test_create_with_restart_policy` pass against the above change,
the test checks that the error message contains `You cannot remove ` instead of
`You cannot remove a running container`


thaJeztah:
ping @shin- ptal 👍

Issue #1369, opened 2016-12-22
support request: testing client code that uses docker(-py) 2


funkyfuture:
after updating to the new sdk, i haven't figured out how to test client code that makes use of the new api, in particular how to mock the low-level `APIClient` properly.

here's a reduced example how i tested my code before the update. note, that this uses `pytest-mocker`, a wrapper around the `mock` library. (addendum: i also tried using `mock` directly, resulting in the same issues)

```python
def test_options(value, result, mocker):
    labels = {'deck-chores.options': value}
    mocker.patch('deck_chores.config.Client.inspect_container',
                 return_value={'Image': '', 'Config': {'Labels': labels}})
    assert parse_labels('container_id') == result
```

the `deck_chores.config` module is where the docker client is instantiated that is used all around the client code.

i tried different things to mock `DockerClient.api.inspect_container` resp. `APIClient.inspect_container`, but none worked. looking at the unittests of the sdk itself, it doesn't seem to be trivial. are there any guidelines how to update tests or how client code should be tested?

funkyfuture:
to elaborate a little more.

to get the labels to parse (within `parse_labels` above), the code called `client.inspect_container`, now it calls `client.api.inspect_container`.

my first approach was to mock like this: 

```python
mocker.patch('deck_chores.config.DockerClient.api.inspect_container',
             return_value={'Image': '', 'Config': {'Labels': labels}})
```

which yields this error:

```
thing = <class 'docker.client.DockerClient'>, comp = 'api'
import_path = 'deck_chores.config.DockerClient.api'

    def _dot_lookup(thing, comp, import_path):
        try:
            return getattr(thing, comp)
        except AttributeError:
>           __import__(import_path)
E           ImportError: No module named 'deck_chores.config.DockerClient'; 'deck_chores.config' is not a package
```

indeed, that is a module (where `DockerClient` is imported), not a package. so, i don't get the error message, my guess is this is due to the fact that the `api` attribute is bound when the client is initialized.

the next try is this what follows the paradigm 'mock where it is used' that worked so far for me.

```python
mocker.patch('docker.api.client.ContainerApiMixin.inspect_container',
             return_value={'Image': '', 'Config': {'Labels': labels}})
```

which evokes this when `client.api.inspect_container` is called:

```
self = <docker.client.DockerClient object at 0x7f9fe4441390>, name = 'api'

    def __getattr__(self, name):
        s = ["'DockerClient' object has no attribute '{}'".format(name)]
[...]

PR #1493, opened 2017-03-04
Adds a few model properties

funkyfuture:
this adds the properties `image` and `labels` for container objects, and `labels` for image objects. this would solve #1369 in a way.

alongside is a patch that made a docs build result actually available outside the container.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -13,5 +13,4 @@ html/*
 __pycache__
 
 # Compiled Documentation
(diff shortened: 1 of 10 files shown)


funkyfuture:
@shin- do these features have a chance of getting merged somehow?

shin-:
Yes, those are pretty useful additions. Thank you!

PR #1494, opened 2017-03-06
Add docker-pycreds to test-requirements.txt

ripcurld0:
To run integration tests docker-pycreds is required.
Before running the integration tests conftest.py is
loaded which results in loads auth.py that
imports the docker-pycreds module.

### DIFF
--- a/test-requirements.txt
+++ b/test-requirements.txt
@@ -3,3 +3,4 @@ pytest==2.9.1
 coverage==3.7.1
 pytest-cov==2.1.0
 flake8==2.4.1
+docker-pycreds==0.2.1


shin-:
Technically, everything that's in `requirements.txt` is also required to run the tests. Why are we making an exception by adding `docker-pycreds` is `test-requirements.txt`?

ripcurld0:
I didn't know that.
I thought it should be there because when I run [this Dockerfile](https://github.com/docker/docker/blob/master/Dockerfile#L198), it only `pip-install`s the modules from test-requirements.txt.

Thanks for the clarification. I am closing this PR.
:man_dancing:

PR #1495, opened 2017-03-07
Fix #1381 - Provide a High Level interface to get exec_run exit_code

### LINKED ISSUE #1381: How to get exec_run exit_code?
Hi,

I create an ubuntu container and then running ```exec_run``` later. But I can not get the exit_code. ```exec_run``` just return output bytecode. I don't know whether the executing result is successful or not.

```
>>> import docker
>>> container = client.containers.run('ubuntu', command=['/bin/bash'], tty=True, detach=True)
>>> container.exec_run(['not_exist'])
b'rpc error: code = 2 desc = oci runtime error: exec failed: exec: "not_exist": executable file not found in $PATH\r\n'
>>>
```

I can use ```APIClient.exec_create(), APIClient.exec_start() and APIClient.exec_inspect()``` to achieve the goal. Is there has any high level api to use?

Thanks for your reply.

roidelapluie:
Signed-off-by: Julien Pivotto <roidelapluie@inuits.eu>

### DIFF
--- a/docker/models/containers.py
+++ b/docker/models/containers.py
@@ -140,6 +140,43 @@ class Container(Model):
             resp['Id'], detach=detach, tty=tty, stream=stream, socket=socket
         )
 
+    def exec_run2(self, cmd, stdout=True, stderr=True, stdin=False, tty=False,
+                  privileged=False, user='', detach=False,
+                  socket=False):
+        """
+        Run a command inside this container and returns its output and exit
+        code. Similar to ``docker exec``.
+
+        Args:
+            cmd (str or list): Command to be executed
+            stdout (bool): Attach to stdout. Default: ``True``
+            stderr (bool): Attach to stderr. Default: ``True``
+            stdin (bool): Attach to stdin. Default: ``False``
+            tty (bool): Allocate a pseudo-TTY. Default: False
+            privileged (bool): Run as privileged.
+            user (str): User to execute command as. Default: root
+            detach (bool): If true, detach from the exec command.
+                Default: False
+
+        Returns:
+            tuple:
+                str: A string containing response data otherwise.
+                int: Return code of the exec
+
+        Raises:
+            :py:class:`docker.errors.APIError`
+                If the server returns an error.
+        """
+        resp = self.client.api.exec_create(
+            self.id, cmd, stdout=stdout, stderr=stderr, stdin=stdin, tty=tty,
+            privileged=privileged, user=user
+        )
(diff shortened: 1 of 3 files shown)


docker/models/containers.py
```diff
@@ -140,6 +140,43 @@ def exec_run(self, cmd, stdout=True, stderr=True, stdin=False, tty=False,
...
+        """
+        resp = self.client.api.exec_create(
+            self.id, cmd, stdout=stdout, stderr=stderr, stdin=stdin, tty=tty,
+            privileged=privileged, user=user
+        )
+        exec_output = self.client.api.exec_start(
+            resp['Id'], detach=detach, tty=tty, stream=False, socket=socket
+        )
+        exec_inspect = self.client.api.exec_inspect(resp['Id'])
+        return (exec_output, exec_inspect['ExitCode'])
```
funkyfuture: there's no need to wrap the return values in brackets. also, i'd say that returning the exit code as first values seems more appropriate.

docker/models/containers.py
```diff
@@ -140,6 +140,43 @@ def exec_run(self, cmd, stdout=True, stderr=True, stdin=False, tty=False,
             resp['Id'], detach=detach, tty=tty, stream=stream, socket=socket
         )
 
+    def exec_run2(self, cmd, stdout=True, stderr=True, stdin=False, tty=False,
```
funkyfuture: `run2` is certainly not a descriptive name. what about a switch like `stream` that controls the return value?

docker/models/containers.py
```diff
@@ -140,6 +140,43 @@ def exec_run(self, cmd, stdout=True, stderr=True, stdin=False, tty=False,
...
+            stderr (bool): Attach to stderr. Default: ``True``
+            stdin (bool): Attach to stdin. Default: ``False``
+            tty (bool): Allocate a pseudo-TTY. Default: False
+            privileged (bool): Run as privileged.
+            user (str): User to execute command as. Default: root
+            detach (bool): If true, detach from the exec command.
+                Default: False
+
+        Returns:
+            tuple:
```
dimaspivak: It might be a good idea to make this return a namedtuple (with `output` and `exit_code`) since then a user could do:
```command = container.exec_run(...)
if command.exit_code == 0:
  ...
```
instead of having to deal with indices.

hnq90:
@roidelapluie Are you still working on this PR? If not, I will continue your work.

roidelapluie:
I am not. Feel free to move forward :)

Issue #1477, opened 2017-02-23
The options 'detach' and 'remove' cannot be used together


al3xdelarge:
The options 'detach' and 'remove' are not conflicting in current versions of docker.

Lines in  docker-py/docker/models/containers.py can be removed.
```
        if detach and remove:
            raise RuntimeError("The options 'detach' and 'remove' cannot be "
"used together.")
```

nsteinmetz:
I would agree but noticed that for old version of docker (like the one in Debian), I think it's still the case. You can't have both.

With recent versions, I agree you can use both.

ztlpn:
Workaround for this and #1579 (checked with docker-py=2.2.1 and docker API v.1.27)
```python
client = docker.from_env()
image = client.images.get('hello-world')
host_config = client.api.create_host_config(auto_remove=True)
container_id = client.api.create_container(image.id, detach=True, host_config=host_config)['Id']
container = client.containers.get(container_id)
container.start()
```

shin-:
Closing via #1545

PR #1497, opened 2017-03-07
Fix documentation for containers.run (auto_remove->remove)

gabrielpjordao:


nsteinmetz:
Hi,

There are other occurences of "auto_remove" accross the code (being code or doc about the code).

Is this PR enough to solve the issue ?

shin-:
Ideally the `DockerClient` would use `auto_remove` for API versions where it is available when `remove=True` is passed as an argument. That would also let us allow using `detach` and `remove` together for sufficient API versions, which would fix #1477 as well.

shin-:
Closing via #1545

PR #1501, opened 2017-03-09
fix: Missing exception handling in split_port when no container port

max8899:
"localhost:host_port:" case will raise TypeError exception directly

Catch the "TypeError" and give proper error message

* docker/utils/ports.py

Signed-off-by: Lei Gong <xue177125184@gmail.com>


shin-:
Thanks!

Issue #980, opened 2016-03-08
Add utility to create a build context from a git repository.


apeace:
The reason is because it won't work if you pass a private git repository as the build URL.

See docker/compose#2856 for more information, including a simple test-case by @zkf.

Instead, `docker-py` should follow the same steps as `docker`: if the build URL is a git repository, it should clone the repo locally and pass as a tarball to the daemon.

I believe the problem is caused by how `docker-py` performs a build ([here](https://github.com/docker/docker-py/blob/master/docker/api/build.py#L15)) instead of following the same steps as `docker`.

Please let me know if I'm misunderstanding the problem or if there's any other way I can help.

Here is my version information:

```
OS X El Capitan 10.11.2 (15C50)
docker-py==1.8.0.dev0
Python 2.7.9
Client:
 Version:      1.11.0-dev
 API version:  1.23
 Go version:   go1.6
 Git commit:   bc730f3-unsupported
 Built:        Tue Mar  8 19:01:29 2016
 OS/Arch:      linux/amd64

Server:
 Version:      1.11.0-dev
 API version:  1.23
 Go version:   go1.6
 Git commit:   bc730f3-unsupported
 Built:        Tue Mar  8 19:01:29 2016
 OS/Arch:      linux/amd64
```

shin-:
I think both options should exist. For public repositories, just passing the git URL is definitely faster.

Feel free to submit a patch to add that feature.

apeace:
Thanks @shin- 

I can work on a patch for this, though I'm sure it will take me some time :) I have a couple questions for you:

1. Are you sure this wouldn't be considered a bug? I believe (though have not yet tested) that this would also not work for public repositories if you use SSH syntax (`git@github.com:user/repo.git`). Essentially I think git SSH URLs will not work. Perhaps the bug should be clarified to say "git SSH URLs" instead of any "git URL"?

2. If indeed this is not a bug, could you give me some guidance on how a `docker-py` client might enable this feature? For example, maybe a client like `docker-compose` might pass a flag (e.g. to [this call](https://github.com/docker/compose/blob/master/compose/service.py#L677))? Any guidance on that would be appreciated--I will incorporate it into my patch.

shin-:
[...]

PR #1508, opened 2017-03-11
Added util for creating build context from git repo

jcm99:
This adds a package to docker.utils for dealing with building directly from git repos.

This feature mostly mirrors how `docker` implements cloning in the client, so it supports the same git protocols, branch checkouts, and subfolders. One notable exception is that it does NOT automatically detect git URLs and clone the repo in the client by default like `docker` does.  As currently implemented, this would require the end user to pass the result as a custom_context to build(). Ref: #980

### DIFF
--- a/docker/errors.py
+++ b/docker/errors.py
@@ -140,6 +140,10 @@ class BuildError(Exception):
     pass
 
 
+class GitError(BuildError):
+    pass
+
+
 def create_unexpected_kwargs_error(name, kwargs):
     quoted_kwargs = ["'{}'".format(k) for k in sorted(kwargs)]
     text = ["{}() ".format(name)]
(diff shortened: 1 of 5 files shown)


iamanikeev:
Hi! Any chance this would be merged soon? Also, there are utils for making a context from git repo, but they are not used in the build method itself. I'll give this a shot today, see if it helps to fix #980

jcm99:
@iamanikeev  I didn't directly integrate into the build() command itself based on the preferences expressed in #980, though it would be trivial to make that change if that was the desired behavior. 

I wasn't sure if there was still interest in adding this feature, so I haven't really tracked whether or not this works in most recent versions of docker-py.

If the project maintainers would like to see this feature added, I'd be happy to take another look and make any updates based on their recommendations.

shin-:
Sorry for the (very) late response. While I'm sure this is useful to some, I don't think it belongs in the Docker SDK. I realize that's not something I expressed clearly in #980 and I apologize for that.
I believe it would make sense as a complementary library though!

PR #1524, opened 2017-03-22
Fix APIError status_code property for client/server errors

alex-dr:
requests.Response objects evaluate as falsy when the status_code
attribute is in the 400-500 range. Therefore we are assured that
prior to this change, APIError would show `is_server_error() == False`
when generated with a 500-level response and `is_client_error() == False`
when generated with a 400-level response. This is not desirable.

Added some seemingly dry (not DRY) unit tests to ensure nothing silly
slips back in here, and of course to verify the problem/fix

### DIFF
--- a/docker/errors.py
+++ b/docker/errors.py
@@ -59,7 +59,7 @@ class APIError(requests.exceptions.HTTPError, DockerException):
 
     @property
     def status_code(self):
-        if self.response:
+        if self.response is not None:
             return self.response.status_code
 
     def is_client_error(self):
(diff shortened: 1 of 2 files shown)


alex-dr:
Is something up with Janky PR tests? It seems like all recent PR's to the project are stuck waiting for this build to be scheduled

shin-:
LGTM, thanks! Such a weird behavior for requests to have!

> Is something up with Janky PR tests? It seems like all recent PR's to the project are stuck waiting for this build to be scheduled

It's obsolete, we need to remove the hook - I just haven't taken the time yet!

PR #1545, opened 2017-04-04
Fix #1477: Allow detach and remove for api versions >= 1.25

### LINKED ISSUE #1477: The options 'detach' and 'remove' cannot be used together

davidsteines:
This pull request addresses the exception that is currently raised when a container attempts to start with "detach" and "remove" both set. The current behavior is to throw a RuntimeError without considering the version of the API.

This pull request implements the following behavior:

1) For API versions >= 1.25, when "detach" and "remove" are both set, do not throw the RuntimeError. Assume that the caller intends for the container to be automatically removed when the container's process exits and set the "auto_remove" flag to True.

2) For API versions < 1.25 continue to throw the RuntimeError.

Although suggested in the discussion thread for issue #1497, this pull request does not automatically convert remove=True to auto_remove=True when the api version is >= 1.25. The reason for this decision is that the remove flag in the run function is tied to logic that captures the logs of the container before removing it. If "remove" were automatically converted to "auto_remove," log capture cannot be guaranteed before the container is removed in cases where detach=False.

Signed-off-by: David Steines <d.steines@gmail.com>

### DIFF
--- a/docker/models/containers.py
+++ b/docker/models/containers.py
@@ -4,6 +4,7 @@ from ..api import APIClient
 from ..errors import (ContainerError, ImageNotFound,
                       create_unexpected_kwargs_error)
 from ..types import HostConfig
+from ..utils import compare_version
 from .images import Image
 from .resource import Collection, Model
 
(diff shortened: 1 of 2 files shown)


illera88:
Nice! Thank you for the pull requests.

Hope someone that has write access can approve it.  (@shin-)
Cheers

illera88:
@shin- is there anything wrong with this pull request?

annatisch:
@shin- We would really be interested in this fix - is it likely to be merged soon? :)

Alphasite:
Not to pester you, but its been a couple of months, has there been any progress made in merging this? Its quite a nuisance to work around.

shin-:
Merged in e17a545aa5b17a2aa8de486a2e2363a6274ea6a7

Thank you!

Issue #1551, opened 2017-04-06
Why does the 'cpuset_mems' only can be set when update?


qq690388648:
I find that the `cpuset_mems` only exist in the parameter of `update` method(https://github.com/docker/docker-py/blob/830b9edce8691440e571270488e7bfa803e62d2d/docker/models/containers.py#L381, https://github.com/docker/docker-py/blob/830b9edce8691440e571270488e7bfa803e62d2d/docker/api/container.py#L1170), but does not exist in the parameter of `HostConfig`'s constructor(https://github.com/docker/docker-py/blob/830b9edce8691440e571270488e7bfa803e62d2d/docker/types/containers.py#L119).

PS: The `HostConfig`'s constructor has `cpuset_cpus`, so I don't think the `cpuset_mems` is missing unintentionally.

shin-:
Closing via #1583

PR #1553, opened 2017-04-06
Add cpuset_mems as the parameter of HostConfig.

qq690388648:
See #1551


shin-:
Thank you!

Do you mind squashing those commits into one?

shin-:
Closing via #1583

PR #1554, opened 2017-04-06
Fix docs container cpu opts

peter-slovak:
Fixed some inconsistencies in documentation, namely:
- cpu_group renamed to cpu_period
- cpu_period renamed to cpu_quota
- duplicate mem_limit description in container create + type (float, str) changed to (int, str) to be consistent with rest of the docs


shin-:
LGTM, thanks!

PR #1556, opened 2017-04-07
Add 'verbose' option for network inspect api

sanimej:
Signed-off-by: Santhosh Manohar <santhosh@docker.com>

### DIFF
--- a/docker/api/network.py
+++ b/docker/api/network.py
@@ -168,14 +168,23 @@ class NetworkApiMixin(object):
         self._raise_for_status(res)
 
     @minimum_version('1.21')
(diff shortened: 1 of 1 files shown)


shin-:
Thanks Santhosh! Do you know which API version introduced that option? We probably want to do a client-side check to see if it's available to the user.

sanimej:
@shin- `verbose` option was introduced in v1.28. I will add a version check and an error message similar to what is done for other optional flags. Thanks.

sanimej:
@shin- Added the version check. PTAL.

shin-:
Merged via #1556

PR #1560, opened 2017-04-12
WIP: Adding nvidia-docker support

andyneff:
Currently just a proof of concept for `nvidia-docker` support, as discussed in docker/compose#4657

## TL;DR;

It's working at least. Automatically used `nvidia-docker` mounts when starting a container with GPU support

---

There's a number of things I'm not 100% happy with yet.

1. Where I had to put the code (at the very least I should put it in its own function)
2. It's pretty fragile right now.
    - If you don't have nvidia on your host, it'll probably fail because devices don't exist
    - If you specify a bad GPU with `NV_GPU` it will fail. But this is the same behavior as `nvidia-docker`, this this is perhaps ok
3. If you want to disable the `nvidia-docker` support feature, there's nothing to do that with.

I'd like some guidance to these issues and what direction you would like me to go in.


shin-:
I've only glanced at this, but maybe you could subclass `APIClient` into `NvidiaAPIClient` instead.

More thoughts when I have some time!

andyneff:
@shin- I've cleaned it up how I think you were suggesting it. I tried a Mixin at first, but everything was static or needed the container api, so it ended up being moved to either the `NvidiaAPIClient` class or `docker.utils.nvidia`. When you get the chance, can you look through it and let me know if anything is unacceptable?

I didn't look at an integration test yet because I didn't know if any of the CI systems have GPU support. Is there something you want to do about that?

andyneff:
@shin- Any idea why the travis-ci says it failed? When I look at the Details it says the Exit code should be zero.

scott-vsi:
This looks promising. When is it going to be available?

docker/api/client.py
```diff
@@ -433,3 +438,73 @@ def get_adapter(self, url):
     @property
     def api_version(self):
         return self._version
+
+
+class NvidiaAPIClient(APIClient):
+    """
+    Version of APIClient that uses the nvidia-docker API to support nvidia GPUs
+
+    Requires `nvidia-docker` to be installed in order to function correctly
```
TomasTomecek: Will this be a hard requirement if this PR gets merged?

TomasTomecek:
I personally don't understand why this code should be part of docker-py, can you enlighten me? What's wrong with having it inside separate project?

docker/api/client.py
andyneff: That is up for debate. But this requirement is *only* for the nvidia support. All other use of docker-py will not require this of course. This is captured in the `nvidia_docker_compatible` function. So I'm open to suggestions to changing the logic of that function. Right now it checks three things

1. Is this linux. Nvidia-docker only works on non-virtual machine instances of linux dockers. So this is limited to Linux hosts only
2. Is the nvidia-driver installed. If you aren't using an nvidia card or the nvidia driver (nouveau for example) then there is use for this either
3. Is the `nvidia-docker` executable found. This would indicate the ["plugin"](https://github.com/NVIDIA/nvidia-docker) is installed

The `nvidia-docker` plugin does two things.

1. Create a special `nvidia-docker` mount called `nvidia_driver_{version_number}`. This includes all the .so files and binaries necessary to use nvidia/cuda in a docker. This is not a simple process and is all coded up in go in the [nvidia-docker](https://github.com/NVIDIA/nvidia-docker/blob/3422bca272d64600ca15874d96e4e511b669c7c4/src/nvidia/volumes.go#L104-L173) project. Due to the complexity and superior knowledge nvidia will have and the chance that they will be updating this, I choose to not even consider reproducing this.
2. Mounting the necessary devices and a special nvidia driver library mount into the container.

I'm reproducing the second of these two in python. The first is accomplished by calling `nvidia-docker` on a dummy container to trigger it in `create_nvidia_driver`. This is why I choose to make `nvidia-docker` a requirement for using the `NvidiaAPIClient`.

andyneff:
@TomasTomecek I am trying to get nvidia support into docker compose. It started with the docker/compose#4657 issue.

`docker-compose` is a popular tool that I would like to use. Unfortunately I deal with GPU work (Cuda, opengl, and opencl). But there is no way for me to use `docker-compose` in its current state.

A work around of "Just add the devices and mounts" myself in the docker compose yaml will only "work on my computer", as enough varies from computer to computer (version numbers, which particular devices get mounted in, etc...) For this reason I feel it is a very useful ability to leverage some of what `nvidia-docker` has created for us.

Since `create_container_config` has to add a few things to accomplish mounting in the correct devices and volumes, I was following @shin- 's advice of creating a `NvidiaAPIClient`. One of the the important things to note is that `APIClient` is unaffected by these changes. But now other apps such as `docker-compose` can use the `NvidiaAPIClient` which will automatically set up nvidia drivers for nvidia enabled docker images (where the label `com.nvidia.volumes.needed` is `nvidia_driver`)

TomasTomecek:
I think this PR raises a more generic issue: what is the preferred way of dealing with plugins which are not inside upstream dockerd and need support in docker-compose. This PR suggests that it is to add the needed code which supports these plugins inside official python API client and docker-compose.

My worry is that if every vendor with their own dockerd plugin adds support to docker-py and docker-compose, we'll end up with code which will be hard to maintain & understand, potentially getting new dependencies which are only needed by users of the feature and finally complicating the API & maybe even cryptic bugs. This is the reason why I still think that either:

1. the code should part of dedicated project
2. `docker-compose` should have plugin system on its own

But hey, I'm no maintainer of this project, I'm just a contributor with an opinion on my own.

andyneff:
@TomasTomecek any ideas for a plugin architecture that would be welcomed in the docker-py/compose projects?

I believe what I've done can be converted to:

1. A mixin for the `APIClient` in docker-py
2. A way to add my own error handling to `handle_connection_errors` in `compose.cli.errors`
3. Running my tests. It would be nice to have a way of to run both my test and the existing test in the current test harness. Of course I would be running it on my own computer or my own CI system to test my plugin, not expecting the main docker-py/compose projects to run tests for plugins.

I'm sure there are other endpoints that would be useful in a plugin system, but these three were sufficient for my case.

TomasTomecek:
> @TomasTomecek any ideas for a plugin architecture that would be welcomed in the docker-py/compose projects?

@andyneff As I said, I'm not maintainer of these projects. It's up to maintainers to figure out if and how they want to merge this code (it's been a month since last comment from a maintainer). I just expressed my personal concerns with the currently proposed PR.

Figuring out plugin architecture for these projects is out of scope of my free time. Sorry.

andyneff:
No long necessary with nvidia-docker2

petergerten:
@andyneff I still don't understand what the solution would be to start a nvidia-docker container with docker-py. Could you provide a simple example?

andyneff:
@petergerten I'm not 100% sure on the docker-py call (as I use `docker-compose` which uses `docker-py` to do this) but I believe when you call `create_container`, you need to specify `runtime="nvidia"`

The new nvidia-docker2 adds a runtime to docker daemon, so using the nvidia-docker2 plugin is as simple as specifying the nvidia runtime. For example, using the docker-cli:

```
docker run -it --rm nvidia/cuda nvidia-smi
```

**Fails**

But

```
docker run --runtime=nvidia -it --rm nvidia/cuda nvidia-smi
```

**Succeeds!**

If this docker-cli test doesn't work for you, then I believe you don't have nvidia-docker2 installed successfully.

---

## Background

The `nvidia-docker` cli used to be a shim that sets up the nvidia specific settings for you (volumes, devices, etc...) if and only if a [label existed](https://gitlab.com/nvidia/cuda/blob/f31ece4bc9195a446f1dd6b9c8479c71fb295dfa/8.0/runtime/Dockerfile#L30). This is what this PR was originally reproducing. Now the `nvidia-docker2` cli is a script that just adds the `--runtime=nvidia` flag for you. This new runtime is a modified version of the docker default runtime, that will add the nvidia specific volumes, devices, etc if and only if an [environment variable](https://gitlab.com/nvidia/cuda/blob/f31ece4bc9195a446f1dd6b9c8479c71fb295dfa/8.0/runtime/Dockerfile#L40-42) is set.

So because `runtime` is part of the `docker`, `docker-py`, and `docker-compose` API, there is no need for this PR.

petergerten:
@andyneff thanks.
I know how to use the docker cli but currently try to use docker-py to start an nvidia runtime container.
I indeed tried to use `runtime="nvidia" ` as well as several variations of it but always get
`
docker.errors.APIError: 400 Client Error: Bad Request ("Unknown runtime specified nvidia")`

andyneff:
### My Environemnt

```bash
pipenv install docker==3.3.0
pipenv run python
```

Running Docker 18.03.1-ce on Fedora 27.

---

```python
import docker
c=docker.from_env()

# Failes
c.containers.run('nvidia/cuda', command='nvidia-smi')

# Succeeds
c.containers.run('nvidia/cuda', command='nvidia-smi', runtime='nvidia')
```

andyneff:
@petergerten If this does not work, can you please verify the contents of `/etc/docker/daemon.json` ???

petergerten:
Running Docker 18.04.0-ce on Arch.
Ok, this might not be docker-py related:
@andyneff 

`nvidia-docker run --runtime=nvidia -it --rm nvidia/cuda nvidia-smi`
works, but
`docker run --runtime=nvidia -it --rm nvidia/cuda nvidia-smi`
does not work. (Same "Unknown runtime" error)

```
[pg@x240 ~]$ docker --version
Docker version 18.04.0-ce, build 3d479c0af6
[pg@x240 ~]$ nvidia-docker --version
Docker version 18.04.0-ce, build 3d479c0af6
```
I do not have a `/etc/docker/daemon.json` file on my system.

petergerten:
@andyneff 
thanks a lot! I got it working now. It was my fault - I was indeed still using nvidia-docker version 1. But the `--version`  output confused me. `version` without dashes showed that I was using v1.

I switched to nvidia-docker2 from arch AUR and now have it working (even with the non stable release):
```
nvidia-docker version
NVIDIA Docker: 2.0.3
Client:
 Version:	18.04.0-ce
 API version:	1.37
 Go version:	go1.10.1
 Git commit:	3d479c0af6
 Built:	Tue Apr 17 22:46:17 2018
 OS/Arch:	linux/amd64
 Experimental:	false
 Orchestrator:	swarm

Server:
 Engine:
  Version:	18.04.0-ce
  API version:	1.37 (minimum version 1.12)
  Go version:	go1.10.1
  Git commit:	3d479c0af6
  Built:	Tue Apr 17 22:46:48 2018
  OS/Arch:	linux/amd64
  Experimental:	false
```

andyneff:
@petergerten That's great! It's possible the arch package maker does release for ever Edge build, which would be awesome. But if that is not the case, I *think* you need to be careful if it's really the "docker 18.03 nvidia-container-runtime" running in docker 18.04. This may work perfectly, or may not. Just keep that in mind if anything crazy starts happening. (The only way I know of to check this is to ask the package manage what version of nvidia-container-runtime is installed, in my case it says `2.0.0-1.docker18.03.1`)

PR #1561, opened 2017-04-12
add service update_preserve method

ornoone:
the Service.update method did not preserve non given attribute.
this add a method named Service.update_preserve to add this
feature.

calling Service.update_preserve(image='alpine') will only update the image, preserving all other definition, as the ``docker`` cli command do. 

the Service.update(**) method did not behave like the ``docker`` cli, is it don't pass all old definitions along with new one.


fmobus:
Hi,

Do you have any updates on this pull request?

ornoone:
since no test seem to work for no apparent reason, and noone in the project seem to care about this, I copied my code in my solution, and it work in production.

PR #1563, opened 2017-04-14
Use `requires[security]` in the requirements instead of just `requires`

cyli:
This installs a few more packages, but guarantees that at least on Mac OS a newer version of openssl is used instead of the default, which doesn't support TLS 1.2.  I think the openssl is provided for Windows wheels too.

See https://stackoverflow.com/questions/31811949/pip-install-requestssecurity-vs-pip-install-requests-difference.

An alternative is to specify a "security" option similar to requests:  see https://github.com/kennethreitz/requests/blob/master/setup.py#L98.

Also since the `requirements.txt` is pinning specific versions, I just pinned everything (including deps of deps) for a consistent build.

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,6 +10,10 @@ The latest stable version [is available on PyPI](https://pypi.python.org/pypi/do
 
     pip install docker
 
+If you are intending to connect to a docker host via TLS, add `docker[tls]` to your requirements instead, or install with pip:
+
+    pip install docker[tls]
+
 ## Usage
 
 Connect to Docker using the default socket or the configuration in your environment:
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -3,7 +3,7 @@ version: '{branch}-{build}'
 install:
   - "SET PATH=C:\\Python27-x64;C:\\Python27-x64\\Scripts;%PATH%"
   - "python --version"
-  - "pip install tox==2.1.1 virtualenv==13.1.2"
+  - "pip install tox==2.7.0 virtualenv==15.1.0"
 
 # Build the binary after tests
 build: false
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,6 +1,16 @@
-requests[security]==2.11.1
-six>=1.4.0
-websocket-client==0.32.0
-backports.ssl_match_hostname>=3.5 ; python_version < '3.5'
-ipaddress==1.0.16 ; python_version < '3.3'
+appdirs==1.4.3
+asn1crypto==0.22.0
+backports.ssl-match-hostname==3.5.0.1
+cffi==1.10.0
+cryptography==1.9
 docker-pycreds==0.2.1
+enum34==1.1.6
+idna==2.5
+ipaddress==1.0.18
+packaging==16.8
+pycparser==2.17
+pyOpenSSL==17.0.0
+pyparsing==2.2.0
+requests==2.14.2
+six==1.10.0
+websocket-client==0.40.0
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,4 @@
-requests==2.11.1
+requests[security]==2.11.1
 six>=1.4.0
 websocket-client==0.32.0
 backports.ssl_match_hostname>=3.5 ; python_version < '3.5'
--- a/setup.py
+++ b/setup.py
@@ -36,10 +36,10 @@ extras_require = {
     # ServerAltname: https://pypi.python.org/pypi/backports.ssl_match_hostname
     ':python_version < "3.3"': 'ipaddress >= 1.0.16',
 
-    # If using docker-py over TLS, highly recommend this option is pip-installed
-    # or pinned.
+    # If using docker-py over TLS, highly recommend this option is
+    # pip-installed or pinned.
 
-    # TODO: if pip installign both "requests" and "requests[security]", the
+    # TODO: if pip installing both "requests" and "requests[security]", the
     # extra package from the "security" option are not installed (see
     # https://github.com/pypa/pip/issues/4391).  Once that's fixed, instead of
     # installing the extra dependencies, install the following instead:
--- a/setup.py
+++ b/setup.py
@@ -35,6 +35,16 @@ extras_require = {
     # ssl_match_hostname to verify hosts match with certificates via
     # ServerAltname: https://pypi.python.org/pypi/backports.ssl_match_hostname
     ':python_version < "3.3"': 'ipaddress >= 1.0.16',
+
+    # If using docker-py over TLS, highly recommend this option is pip-installed
+    # or pinned.
+
+    # TODO: if pip installign both "requests" and "requests[security]", the
+    # extra package from the "security" option are not installed (see
+    # https://github.com/pypa/pip/issues/4391).  Once that's fixed, instead of
+    # installing the extra dependencies, install the following instead:
+    # 'requests[security] >= 2.5.2, != 2.11.0, != 2.12.2'
+    'tls': ['pyOpenSSL>=0.14', 'cryptography>=1.3.4', 'idna>=2.0.0'],
 }
 
 version = None


shin-:
I think we should leave it up to users to figure out if they need those additional packages. A sizable amount of users don't require TLS at all when communicating with the engine over a UNIX socket / Windows npipe.

cyli:
@shin How would you feel about a `tls` option in the `setup.py`?  So people would have to pin or `pip install docker[tls]`?

cyli:
(I'm not sure what is causing the windows failure)

shin-:
Looks like building the `cryptography` module on AppVeyor fails because `opensslv.h` is missing.

```
Failed to build cryptography
Installing collected packages: mock, colorama, py, pytest, coverage, pytest-cov, mccabe, pyflakes, pep8, flake8, appdirs, asn1crypto, backports.ssl-match-hostname, pycparser, cffi, idna, six, pyparsing, packaging, cryptography, docker-pycreds, enum34, ipaddress, pyOpenSSL, requests, websocket-client
  Running setup.py install for cryptography
    Complete output from command c:\projects\docker-py\.tox\py35\scripts\python.exe -c "import setuptools, tokenize;__file__='C:\\Users\\appveyor\\AppData\\Local\\Temp\\1\\pip-build-mitiwy39\\cryptography\\setup.py';exec(compile(getattr(tokenize, 'open', open)(__file__).read().replace('\r\n', '\n'), __file__, 'exec'))" install --record C:\Users\appveyor\AppData\Local\Temp\1\pip-ltw8vmfe-record\install-record.txt --single-version-externally-managed --compile --install-headers c:\projects\docker-py\.tox\py35\include\site\python3.5\cryptography:
    running install
    running build
    running build_py
    running egg_info
    writing dependency_links to src\cryptography.egg-info\dependency_links.txt
    writing requirements to src\cryptography.egg-info\requires.txt
    writing top-level names to src\cryptography.egg-info\top_level.txt
    writing src\cryptography.egg-info\PKG-INFO
    writing entry points to src\cryptography.egg-info\entry_points.txt
    warning: manifest_maker: standard file '-c' not found
    
    reading manifest file 'src\cryptography.egg-info\SOURCES.txt'
    reading manifest template 'MANIFEST.in'
    no previously-included directories found matching 'docs\_build'
    warning: no previously-included files matching '*' found under directory 'vectors'
    writing manifest file 'src\cryptography.egg-info\SOURCES.txt'
    running build_ext
    generating cffi module 'build\\temp.win32-3.5\\Release\\_padding.c'
    already up-to-date
    generating cffi module 'build\\temp.win32-3.5\\Release\\_constant_time.c'
    already up-to-date
    generating cffi module 'build\\temp.win32-3.5\\Release\\_openssl.c'
    already up-to-date
    building '_openssl' extension
    C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\BIN\cl.exe /c /nologo /Ox /W3 /GL /DNDEBUG /MD -Ic:\python35\include -Ic:\python35\include "-IC:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\INCLUDE" "-IC:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\ATLMFC\INCLUDE" "-IC:\Program Files (x86)\Windows Kits\10\include\10.0.14393.0\ucrt" "-IC:\Program Files (x86)\Windows Kits\NETFXSDK\4.6.1\include\um" "-IC:\Program Files (x86)\Windows Kits\10\include\10.0.14393.0\shared" "-IC:\Program Files (x86)\Windows Kits\10\include\10.0.14393.0\um" "-IC:\Program Files (x86)\Windows Kits\10\include\10.0.14393.0\winrt" /Tcbuild\temp.win32-3.5\Release\_openssl.c /Fobuild\temp.win32-3.5\Release\build\temp.win32-3.5\Release\_openssl.obj
    _openssl.c
    build\temp.win32-3.5\Release\_openssl.c(434): fatal error C1083: Cannot open include file: 'openssl/opensslv.h': No such file or directory
    error: command 'C:\\Program Files (x86)\\Microsoft Visual Studio 14.0\\VC\\BIN\\cl.exe' failed with exit status 2
```

Do we need those changes to `requirements.txt`?

shin-:
https://github.com/pyca/cryptography/issues/3028

Not sure what version of `pip` AppVeyor is using.

shin-:
`pip` is up to date on Appveyor, seems like the solution is to force `pip` to install the wheel using `--only-binary`.

This should work: https://github.com/shin-/docker-py/commit/eeb29b92ad609cf8818854930d7f1b7740d55650

cyli:
@shin- Ah ok, thank you!  Will update.

shin-:
```
Collecting cryptography==1.8.1 (from -r C:\projects\docker-py/requirements.txt (line 5))
  Could not find a version that satisfies the requirement cryptography==1.8.1 (from -r C:\projects\docker-py/requirements.txt (line 5)) (from versions: 0.2, 0.2.1, 0.2.2, 0.3, 0.4, 0.5, 0.5.1, 0.5.2, 0.5.3, 0.5.4, 0.6, 0.6.1, 0.7, 0.7.1, 0.7.2, 0.8, 0.8.1, 0.8.2, 0.9, 0.9.1, 0.9.2, 0.9.3, 1.0, 1.0.1, 1.0.2, 1.1, 1.1.1, 1.1.2, 1.2, 1.2.1, 1.2.2, 1.2.3, 1.3, 1.3.1, 1.3.2, 1.3.3, 1.3.4, 1.4)
No matching distribution found for cryptography==1.8.1 (from -r C:\projects\docker-py/requirements.txt (line 5))
```

I guess 1.8.1 is only available on UNIX platforms?

cyli:
@shin- I believe there a wheels for windows:  https://pypi.org/project/cryptography/1.8.1/#files - I'm not sure what's up

tox.ini
```diff
@@ -1,15 +1,22 @@
...
-envlist = py27, py33, py34, py35, flake8
+envlist = {py27,py33,py34,py35}-{windows,linux,darwin},flake8
 skipsdist=True
 
 [testenv]
 usedevelop=True
 commands =
     py.test -v --cov=docker {posargs:tests/unit}
+platform =
+    windows: windows
```
moshez: This should be 
 
```
{py27,py33,py34,py35}-windows: windows
```

and ditto for the other relevant lines.

If you're asking whether I feel bad already for giving you this advice: yes but I have no better advice :(

tox.ini
```diff
@@ -1,15 +1,22 @@
...
+    linux: linux
+    darwin: darwin
+install_command = 
+    {py27,py33,py34,py35}-windows: pip install --only-binary=cryptography {opts} {packages}
```
moshez: you probably want to put an install command here too, because otherwise things will break :(

```
{py27,...}-{darwin,linux}: pip install {opts} {packages}
```

tox.ini
cyli: ah, good point, thanks!

cyli:
@shin- Looks like we were using an older version of tox/virtualenv in appveyor, which installed an older version of pip, which for some reason did not find a matching wheel.

cyli:
> Do we need those changes to requirements.txt?

I wasn't sure if this was in reference to the appveyor failure or a more general question about the extra packages in the requirements.txt in this particular PR.  If the latter, I have mainly been following the advice in https://caremad.io/posts/2013/07/setup-vs-requirement/ and https://packaging.python.org/discussions/install-requires-vs-requirements/#requirements-files, where the `setup.py` has the dependencies required to install the package, generally, and `requirements.txt` pins the exact versions of both dependencies and dependencies of dependencies, to make sure that CI is repeatable.

shin-:
Sorry for neglecting this for so long, I thought I had merged it. Thank you!

cyli:
No worries, thank you!

PR #1568, opened 2017-04-18
Fix Issue1567

antoineVerlant:
Here is how I fixed the parsing.
Rather than append element and remove element in the list on which the loop is, I created a new list wherre I append the parsed_mount.
Then I replace it.

### DIFF
--- a/docker/types/services.py
+++ b/docker/types/services.py
@@ -107,11 +107,14 @@ class ContainerSpec(dict):
         if labels is not None:
             self['Labels'] = labels
         if mounts is not None:
+            parsed_mounts = []
             for mount in mounts:
                 if isinstance(mount, six.string_types):
-                    mounts.append(Mount.parse_mount_string(mount))
-                    mounts.remove(mount)
-            self['Mounts'] = mounts
+                    parsed_mounts.append(Mount.parse_mount_string(mount))
+                else:
+                    # If mount already parsed
+                    parsed_mounts.append(mount)
+            self['Mounts'] = parsed_mounts
         if stop_grace_period is not None:
             self['StopGracePeriod'] = stop_grace_period
 


antoineVerlant:
Sry this is my first Pull Request ever, not use to it. Is there anything else I can do for you ?

antoineVerlant:
Any news ?

shin-:
Hi Antoine,

Thank you for taking the time to write and submit a PR! Sorry for not responding earlier, I've been busy with other tasks lately. There's a few things I need before I can accept this PR:

1. [Please sign your commits](https://github.com/docker/docker-py/pull/1568#issuecomment-294865101)
2. Make sure the [`test_create_with_volume_mount`](https://github.com/docker/docker-py/blob/master/tests/integration/api_service_test.py#L136-L156) passes. You can verify this locally by running `make test` in the project's root folder.
3. As a bonus, if you were able to add a test checking that the behaviour that used to fail doesn't do so anymore, that would be really great. If that seems too daunting, let me know and I can add that myself.

antoineVerlant:
Hey again, Thank for your answer.
Shoud I sign the 2 first commit too ?
I admit I don't realy got how those test work and why it failed. Can you help me with this.
To use the debugger I used these test in my own project:
   ```
 def test_create_service_with_volume_mount(self):
        vol_name = 'banane'
        print vol_name
        container_spec = docker.types.ContainerSpec(
            'nib:oo', ['ls'],
            mounts=[
                docker.types.Mount(target='/test', source=vol_name)
            ]
        )
        # self.tmp_volumes.append(vol_name)
        task_tmpl = docker.types.TaskTemplate(container_spec)
        name = 'pomme'
        client = docker.APIClient(base_url='unix://var/run/docker.sock')
        svc_id = client.create_service(task_tmpl, name=name)
        svc_info = client.inspect_service(svc_id)
        print 'ContainerSpec' in svc_info['Spec']['TaskTemplate']
        cspec = svc_info['Spec']['TaskTemplate']['ContainerSpec']
        print 'Mounts' in cspec
        print len(cspec['Mounts']) == 1
        mount = cspec['Mounts'][0]
        print mount['Target'] == '/test'
        print mount['Source'] == vol_name
        print mount['Type'] == 'volume'
```
Every print is true (exept the first one wich is banane).
So the test should work right ?
In the original code you don't test if the mount element is already parsed into a dictionary but shoud I change my else into an elif isintance ...

Thanks for your answer and time.

shin-:
Yes, every commit should be signed. Alternatively, you can squash them all into a single commit.

Changes look good now. Thanks!

antoineVerlant:
Here it is.
Thanks for your support

PR #1576, opened 2017-04-27
Fix if replicas is set to 0, Fixes #1572

### LINKED ISSUE #1572: Can't Create/Update an replicated service with replicas = 0, bug finded
I try to update a service with the following code:
```
 service.update(
            name=service.name,
            mode=docker.types.ServiceMode('replicated', replicas=0),
            networks=networks_list,
        )
```
First, I want to update the replicas only but I have to put name and network or the api think I want to change them (?) then send me an error.
Second, the ServiceMode object created here don't work and looks like this :  `{'replicated': {}}`
The reason for this is https://github.com/docker/docker-py/blob/master/docker/types/services.py#L410 where the if should be an if exist.
I would love to do a pull request but i'm a newbie and I'm already lost with my first one : #1568 
Thx for everything

robkooper:
This will fix the issue in #1572.

Added a test to make sure that if replicas is set to 0 it will really be 0.

Following code would return false in case of replicas is 0, this would prevent from updating a service to have 0 replicas.

```
if replicas:
  do something
```

### DIFF
--- a/docker/types/services.py
+++ b/docker/types/services.py
@@ -407,7 +407,7 @@ class ServiceMode(dict):
                 'replicas can only be used for replicated mode'
             )
         self[mode] = {}
-        if replicas:
+        if replicas is not None:
             self[mode]['Replicas'] = replicas
 
     @property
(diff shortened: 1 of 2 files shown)


shin-:
Thank you!

PR #1581, opened 2017-04-28
Handle multiple success messages during image building.

### LINKED ISSUE #1577: images.build fails when using tag argument.
When using the `tag` kwarg on `client.images.build`, it will always throw a Build Error.

The api.build output looks like this

```
['{"stream":"Step 1/4 : FROM scratch\\n"}\r\n',
 '{"stream":" ---\\u003e \\n"}\r\n',
 '{"stream":"Step 2/4 : LABEL com.nvidia.volumes.needed \\"nvidia_driver\\"\\n"}\r\n',
 '{"stream":" ---\\u003e Using cache\\n"}\r\n',
 '{"stream":" ---\\u003e 36ec3942c5f5\\n"}\r\n',
 '{"stream":"Step 3/4 : SHELL /usr/local/nvidia/bin/nvidia-smi\\n"}\r\n',
 '{"stream":" ---\\u003e Running in f875d54529eb\\n"}\r\n',
 '{"stream":" ---\\u003e b750cf87aed6\\n"}\r\n',
 '{"stream":"Step 4/4 : CMD -L\\n"}\r\n',
 '{"stream":" ---\\u003e Running in f3f9e21b8171\\n"}\r\n',
 '{"stream":" ---\\u003e 61c46a80da73\\n"}\r\n',
 '{"stream":"Successfully built 61c46a80da73\\n"}\r\n',
 '{"stream":"Successfully tagged 5b3bc129-d296-4b7d-872c-bc7117d4f327:latest\\n"}\r\n']
```

The problem can be tracked down to [here](https://github.com/docker/docker-py/blob/2.2.1-release/docker/models/images.py#L164-L167). The code is assuming the "Successfully built" comes last, however in the tag case, this is not true.

Two ideas are to either 

1. Search the last two lines
2. Or in case that is not enough, search all the lines for "Successfully built"

I'm using docker server 17.05.0-ce-rc1

aaronthebaron:
Fixes #1577

Signed-off-by: Aaron Cowdin <aaron.cowdin@gmail.com>

### DIFF
--- a/docker/models/images.py
+++ b/docker/models/images.py
@@ -169,14 +169,15 @@ class ImageCollection(Collection):
         events = list(json_stream(resp))
         if not events:
             return BuildError('Unknown')
-        event = events[-1]
-        if 'stream' in event:
-            match = re.search(r'(Successfully built |sha256:)([0-9a-f]+)',
-                              event.get('stream', ''))
-            if match:
-                image_id = match.group(2)
-                return self.get(image_id)
+        for event in events:
+            if 'stream' in event:
+                match = re.search(r'(Successfully built |sha256:)([0-9a-f]+)',
+                                  event.get('stream', ''))
+                if match:
+                    image_id = match.group(2)
+                    return self.get(image_id)
 
+        event = events[-1]
         raise BuildError(event.get('error') or event)
 
     def get(self, name):
--- a/docker/models/images.py
+++ b/docker/models/images.py
@@ -166,19 +166,18 @@ class ImageCollection(Collection):
         resp = self.client.api.build(**kwargs)
         if isinstance(resp, six.string_types):
             return self.get(resp)
-        events = list(json_stream(resp))
-        if not events:
-            return BuildError('Unknown')
-        for event in events:
-            if 'stream' in event:
+        for chunk in json_stream(resp):
+            if 'error' in chunk:
+                raise BuildError(chunk['error'])
+                break
+            if 'stream' in chunk:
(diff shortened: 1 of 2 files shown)


docker/models/images.py
```diff
@@ -169,14 +169,15 @@ def build(self, **kwargs):
...
         if not events:
             return BuildError('Unknown')
-        event = events[-1]
-        if 'stream' in event:
-            match = re.search(r'(Successfully built |sha256:)([0-9a-f]+)',
-                              event.get('stream', ''))
-            if match:
-                image_id = match.group(2)
-                return self.get(image_id)
+        for event in events:
```
shin-: If we're going through all events anyway, the conversion to list we do above is wasteful. We should iterate on `json_stream(resp)` directly.

docker/models/images.py
```diff
@@ -169,14 +169,15 @@ def build(self, **kwargs):
...
+            if 'stream' in event:
+                match = re.search(r'(Successfully built |sha256:)([0-9a-f]+)',
+                                  event.get('stream', ''))
+                if match:
+                    image_id = match.group(2)
+                    return self.get(image_id)
 
+        event = events[-1]
```
shin-: maybe we could interrupt the loop on any `event` that has an `error` key.

aaronthebaron:
@shin let me know if these changes fit the bill. Thanks.

shin-:
LGTM, thank you =)

PR #1582, opened 2017-04-30
Clear uname/gname when creating tar archive

### LINKED ISSUE #998: `Client.build` and `docker build` don't share a build cache
*Copied from https://github.com/docker/compose/issues/3148.*

I discovered a few days ago that when building images, a cache generated from `docker build` will not be used by `docker-compose build` or vice versa. This happens both on my Mac laptop (Docker Machine, Docker v1.10.3 installed through Homebrew) and on an Ubuntu 14.04 server (Docker v1.10.2 installed through `apt-get`).

Version info for my laptop:
```
docker-compose version 1.6.2, build unknown
docker-py version: 1.7.2
CPython version: 2.7.10
OpenSSL version: OpenSSL 0.9.8zg 14 July 2015
```

Version info on my server:
```
docker-compose version 1.6.2, build 4d72027
docker-py version: 1.7.2
CPython version: 2.7.9
OpenSSL version: OpenSSL 1.0.1e 11 Feb 2013
```

It is simple to reproduce. The directory *tiny-app* in https://github.com/SamirTalwar/docker-build-weirdness shows the issue:

```
$ docker build --tag=tiny-app .
Sending build context to Docker daemon 4.096 kB
Step 1 : FROM ruby
 ---> 0f58cbcb8dce
Step 2 : WORKDIR /app
 ---> Running in be80cb8a6324
 ---> 2972973d66d4
Removing intermediate container be80cb8a6324
Step 3 : COPY script ./
 ---> 77bc9f2b16eb
Removing intermediate container dc9beae186af
Step 4 : CMD ./script
 ---> Running in ea5f91334ea4
 ---> d69db9b758e0
Removing intermediate container ea5f91334ea4
Successfully built d69db9b758e0

$ docker-compose build
Building tiny-app
Step 1 : FROM ruby
 ---> 0f58cbcb8dce
Step 2 : WORKDIR /app
 ---> Using cache
 ---> 2972973d66d4
Step 3 : COPY script ./
 ---> abb5045392d4
Removing intermediate container 1b1450dd66eb
Step 4 : CMD ./script
 ---> Running in a0c7889ca68a
 ---> 354c8fcf4876
Removing intermediate container a0c7889ca68a
Successfully built 354c8fcf4876
```

From the `COPY` operation onwards, it no longer uses the cache.

I don't know, but I am pretty sure, that this is because the tarballs sent to the Docker server are different. Also in that repo are the uploaded tarballs and hex dumps (`hexdump -C`), of the same from the CLI and Compose (through docker-py), captured through a fake HTTP server, *server.py* in the repo. The only real differences seem to be the inclusion of the owner user and group names in the latter, and a flag in the file mode header (grep for `0100644` vs. `0000644`) which I cannot find the life of me find documentation on.

I could, of course, be way off. And even if I'm on track, I can't be sure it'll be considered a bug here as opposed to the server.

Whatever the reason, I'd love it if we could reconcile these so the cache is used. Cheers!

Larsjep:
To ensure that the cache from docker command client
can be reused, we must clear group/user name as the docker
client does.

Together with golang/go#20150 this fixes #998

Signed-off-by: Lars Jeppesen <lj@linel.dk>

### DIFF
--- a/docker/utils/utils.py
+++ b/docker/utils/utils.py
@@ -103,6 +103,10 @@ def create_archive(root, files=None, fileobj=None, gzip=False):
             # ignore it and proceed.
             continue
 
+        # Clear uname/gname as the docker client does.
+        i.uname = ''
+        i.gname = ''
+
         if constants.IS_WINDOWS_PLATFORM:
             # Windows doesn't keep track of the execute bit, so we make files
             # and directories executable by default.


Larsjep:
Anyone knows what this "janky" check is and why nothing happens with it ?

tomyan:
Hi Larsjep

Thanks for working on a fix for this. Any idea when it's likely to be merged?

Thanks

Tom

Larsjep:
Hi @tomyan,

I'm sorry but I'm also waiting for a response from the maintainers.
They might be busy with other stuff, because I can see several other pull requests waiting.

Br
  Lars

shin-:
Thank you @Larsjep !

Sorry about the late response - my understanding is that this patch will not do anything unless it is used with a Docker Engine built with Go 1.9+, of which there has been none so far. Once one such Engine has been released, we can test those changes to ensure they have the expected result, and if confirmed, merge the PR.

edmorley:
I may be misreading, but it looks like moby/moby#33935 has re-added the very bits that were intentionally removed in go1.9, in an attempt to fix the breakage seen in moby/moby#33892?

oskarpearson:
Hi @shin- and @Larsjep

Docker CE 17.12.0 has just been released, and includes "Bump Go to 1.9.2" for both moby and the cli. See the Runtime section of https://github.com/docker/docker-ce/releases/tag/v17.12.0-ce

Is there a way I can test this patch out?

Thanks for your help!

Larsjep:
Unfortunately, the docker team decided to reintroduce the incorrect bits to avoid breaking the cache.
They said that they already planned to break the cache when introducing buildkit and then correct the bits at that point. But I haven't been able to get an answer on when buildkit would be implemented and that is 4 months ago :(.
See moby/moby#33935

kilianc:
Any updates on this?

AlJohri:
Bump, also running into this issue

Larsjep:
> Bump, also running into this issue

Have you tried using buildkit?. That should have been include since Docker 18.06. The docker team promised that they would fix the bits when switching to buildkit. If that is the case this PR should fix the problem.
I'm no longer working on the project where I discovered this issue, so I will not have time to test it myself.

AlJohri:
No, I haven't tried using buildkit yet. Just the default docker and docker-compose that comes with Docker for Mac (19.03.5).

Larsjep:
> No, I haven't tried using buildkit yet. Just the default docker and docker-compose that comes with Docker for Mac (19.03.5).
It should be included with that build, but you will need to enable it by setting DOCKER_BUILDKIT=1.
See here: [https://docs.docker.com/develop/develop-images/build_enhancements/]()

But I'm quite sure that you will also need the patch from this PR.

PR #1583, opened 2017-05-01
CPUset* parameter fixes

delcypher:
This PR addresses two issues

* `CpusetMems` not being supported in `create_host_config()` and in `Container.run()`  (see #1551). The commit I've done here is based on #1553 which seems to have been abandoned.
* Fix a typo `s/CpuSetCpus/CpusetCpus/`

I tried running the tests (via `make test`) but they fail for me from the point where I branched from `master` so I don't really know if I've broken anything.


delcypher:
@shin- You expressed interest in #1553 so I'd appreciate any help you can offer in getting this merged. It would also be great if a new release can be made so I can grab a version of `docker-py` that has the fixes in this PR.

PR #1587, opened 2017-05-02
Documentation fixes for login func

terminalmage:
This makes a small tweak to the grammar of the documentation for the
reauth argument, and also updates the dockercfg_path docs for accuracy.


shin-:
Thank you!

PR #1592, opened 2017-05-06
Use collection's get method inside ImageCollection's list method

adrianliaw:
- Replicate the same thing as in `ContainerCollection`
- Without doing this, `Image.attrs` will not be the result of `docker inspect`, unless you run `Image.reload` method once
- Originally, getting `Image.labels` property raises an error because there's no `Labels` field in `Configs`


shin-:
Thank you!

PR #1616, opened 2017-05-17
Fix type checking for nano_cpus

AlexeyRokhin:
Fixes for https://github.com/docker/docker-py/pull/1570
int replaced with six.integer_types.

PR #1619, opened 2017-05-17
fix type checking for nano_cpus

shin-:
Rebased #1616 

Thanks @AlexeyRokhin


AlexeyRokhin:
@shin- 
Is it possible to include this fix into docker-py 2.3.1 in order to release it before compose 1.14?
Without this fix compose will throw error when nano_cpus >2^32 (cpus >= ~2.2).

PR #1631, opened 2017-06-01
Add support for ``runtime`` config

madhuri-rai07:


madhuri-rai07:
Hi, This is my first commit to docker repo. Please let me know if I need to change anything. Thanks!

docker/types/containers.py
```diff
@@ -473,6 +473,9 @@ def __init__(self, version, binds=None, port_bindings=None,
 
             self['NanoCpus'] = nano_cpus
 
+        if runtime:
```
shin-: What's the minimum API version for this feature? Please add a version check similar to the other options in this method.

docs/change-log.md
```diff
@@ -122,6 +122,8 @@ Change log
 * Added support for `force_update` in `TaskTemplate`
 * Made `name` parameter optional in `APIClient.create_volume` and
   `DockerClient.volumes.create`
+* Added support for `runtime` in `APIClient.create_container` and
+  `DockerClient.containers.run`
```
shin-: We only update the ChangeLog at release time. This should be removed.

tests/integration/api_container_test.py
```diff
@@ -1255,6 +1255,14 @@ def test_container_cpuset(self):
         inspect_data = self.client.inspect_container(container)
         self.assertEqual(inspect_data['HostConfig']['CpusetCpus'], cpuset_cpus)
 
+    def test_create_with_runtime(self):
```
shin-: Please add a `@requires_api_version` decorator with the minimum API version for this feature.

docker/types/containers.py
madhuri-rai07: @shin- Thank you for reviewing. I have updated the patch.

madhuri-rai07:
@shin- Hi Can you help merge this? Thanks.

hongbin:
LGTM

cascandaliato:
[`create_container`](https://github.com/docker/docker-py/blob/main/docker/api/container.py#L220) stores the `runtime` parameter in the Config dictionary while `containers.run/create` stores it in the HostConfig dictionary via [`_create_container_args`](https://github.com/docker/docker-py/blob/c38656dc7894363f32317affecc3e4279e1163f8/docker/models/containers.py#L1115)/[`RUN_HOST_CONFIG_KWARGS`](https://github.com/docker/docker-py/blob/c38656dc7894363f32317affecc3e4279e1163f8/docker/models/containers.py#L1100). Is this correct?

Also, the request gets posted to `/containers/create` but `runtime` doesn't appear in the official API documentation for [v1.43](https://docs.docker.com/engine/api/v1.43/#tag/Container/operation/ContainerCreate) (or [v1.25](https://docs.docker.com/engine/api/v1.25/#tag/Container/operation/ContainerCreate)). Does it get injected in a special way?

cascandaliato:
Nevermind, while it's true that the two commands are specifying the `Runtime` via different objects (`APIClient.create_container` -> `Config`, `containers.run` -> `HostConfig`) the Docker API seems to accept both. Out of curiosity, `docker run --runtime` uses `HostConfig.Runtime`.

PR #1632, opened 2017-06-05
allow ipv6 :: notation in split_port (using re)

kaiyou:
In the case of a defined format with specific parts, a regular
expression with named capturing bits makes reasoning about the
parts simpler than imlementing a parser from scratch.

This is strongly related to https://github.com/docker/docker-py/pull/1404 and simply provides a suggested implementation using regular expressions.

### DIFF
--- a/docker/utils/ports.py
+++ b/docker/utils/ports.py
@@ -1,3 +1,16 @@
+import re
+
+PORT_SPEC = re.compile(
+    "^"  # Match full string
+    "("  # External part
+    "((?P<host>[a-fA-F\d.:]+):)?"  # Address
+    "(?P<ext>[\d]*)(-(?P<ext_end>[\d]+))?:"  # External range
+    ")?"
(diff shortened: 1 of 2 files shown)


kaiyou:
Sorry about commit signatures, will fix that asap.

docker/utils/ports.py
```diff
@@ -24,81 +37,41 @@ def build_port_bindings(ports):
...
-                            for ex_port in external_range]
+    if host is None:
+        if external is not None and len(internal) != len(external):
+            raise ValueError('Port ranges don\'t match in length')
+        return internal, external
+    else:
+        if not external:
+            external = [None] * len(internal)
+        elif len(internal) != len(external):
+            raise ValueError('Port ranges don\'t match in length')
```
shin-: nitpick: It feels like there should be a way to have this `raise` statement only once, but it's not immediately obvious to me what is the best way to get there.

Issue #1374, opened 2017-01-02
exception ReadTimeout: UnixHTTPConnectionPool when calling container.stop()


pommetjehorlepiep:
OS: Debian 8 Jessie 64-bit

**pip show docker:**
Name: docker
Version: 2.0.0
Summary: A Python library for the Docker Engine API.
Home-page: https://github.com/docker/docker-py
Author: Joffrey F
Author-email: joffrey@docker.com
License: UNKNOWN
Location: /usr/local/lib/python2.7/dist-packages
Requires: backports.ssl-match-hostname, ipaddress, docker-pycreds, websocket-client, requests, six

**docker version:**
Client:
 Version:      1.12.5
 API version:  1.24
 Go version:   go1.6.4
 Git commit:   7392c3b
 Built:        Fri Dec 16 02:21:54 2016
 OS/Arch:      linux/amd64

Server:
 Version:      1.12.5
 API version:  1.24
 Go version:   go1.6.4
 Git commit:   7392c3b
 Built:        Fri Dec 16 02:21:54 2016
 OS/Arch:      linux/amd64

**ReadTimeout exception:**

  File "/usr/local/lib/python2.7/dist-packages/docker/models/containers.py", line 343, in stop
    return self.client.api.stop(self.id, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/docker/utils/decorators.py", line 21, in wrapped
    return f(self, resource_id, *args, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/docker/api/container.py", line 1084, in stop
    timeout=(timeout + (self.timeout or 0)))
  File "/usr/local/lib/python2.7/dist-packages/docker/utils/decorators.py", line 47, in inner
    return f(self, *args, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/docker/api/client.py", line 171, in _post
    return self.post(url, **self._set_request_timeout(kwargs))
  File "/usr/local/lib/python2.7/dist-packages/requests/sessions.py", line 535, in post
    return self.request('POST', url, data=data, json=json, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/requests/sessions.py", line 488, in request
    resp = self.send(prep, **send_kwargs)
  File "/usr/local/lib/python2.7/dist-packages/requests/sessions.py", line 609, in send
    r = adapter.send(request, **kwargs)
  File "/usr/local/lib/python2.7/dist-packages/requests/adapters.py", line 499, in send
[...]

PR #1634, opened 2017-06-08
Ensure default timeout is used by API Client

grahamlyons:
The `from_env` method on the `docker` module passed `None` as the value for the `timeout` keyword argument which overrode the default value in the initialiser, taken from `constants` module.

~This sets the default in the initialiser to `None` and adds logic to set that, in the same way that `version` is handled.~

This sets the default in `from_env` to the default defined in the `constants` module.

Addresses #1374

### DIFF
--- a/docker/api/client.py
+++ b/docker/api/client.py
@@ -83,8 +83,7 @@ class APIClient(
             configuration.
         user_agent (str): Set a custom user agent for requests to the server.
     """
-    def __init__(self, base_url=None, version=None,
-                 timeout=DEFAULT_TIMEOUT_SECONDS, tls=False,
+    def __init__(self, base_url=None, version=None, timeout=None, tls=False,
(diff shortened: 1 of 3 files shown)


grahamlyons:
Is the "janky" check configured correctly? It doesn't have a link to a Jenkins job and it hasn't completed on the last large handful of pull requests.

grahamlyons:
I'm not sure what the failure on Jenkins relates to; can anybody assist with that?

shin-:
Please ignore the `janky` test... It's an obsolete test, but we somehow haven't been able to disable the corresponding webhook.

grahamlyons:
@shin-, is that the kind of change you envisaged? Let me know if not.

shin-:
Yeah, I think that works! Thank you!

grahamlyons:
Awesome. Thanks, @shin-

PR #1641, opened 2017-06-13
join_swarm should never send empty ListenAddr

unitymind:
We should never send empty values for `ListenAddr` in `/swarm/join` api call.
 
Docker Engine could decide by self about handling default case. Otherwise:
- if we send empty `ListenAddr` - always fail with `400 Client Error: Bad Request ("invalid ListenAddr "": invalid empty address")`. Just block using default `0.0.0.0:2377` by Docker Engine.

And we not should provide default `0.0.0.0:2377`, because it's internal Docker Engine default not related to client api call specs.

### DIFF
--- a/docker/api/swarm.py
+++ b/docker/api/swarm.py
@@ -171,6 +171,12 @@ class SwarmApiMixin(object):
             "JoinToken": join_token,
             "AdvertiseAddr": advertise_addr,
         }
+        # We should never send empty value for ListenAddr.
+        # Docker Engine could decide by self about handling default case (use '0.0.0.0:2377')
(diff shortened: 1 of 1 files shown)


unitymind:
Build fails due `flake8`

shin-:
That is already handled in `APIClient._post_json`

unitymind:
@shin- Hmm.. seems like Docker bug, because...

`join_swarm(remote_addrs, join_token)` (with remain two default `None`) will **never** went through and never ends up successful, but **always** fires through `400 Client Error: Bad Request ("invalid ListenAddr "": invalid empty address")`.

Issue #1515, opened 2017-03-17
Add support for floats in "since" kwarg on logs().


mccalluc:
The [documentation](https://docker-py.readthedocs.io/en/stable/containers.html#docker.models.containers.Container.logs) is clear:
> since (datetime or int) – Show logs since a given datetime or integer epoch

but it's easy for someone to overlook that and pass in something based on a float from `time.time()` and `since` will just be ignored. Would a PR that supported any number or a datetime be welcome?

```
django-docker-engine==0.0.1
docker==2.1.0
docker-pycreds==0.2.1
Python 2.7.13 :: Continuum Analytics, Inc.
Client:
 Version:      17.03.0-ce
 API version:  1.26
 Go version:   go1.7.5
 Git commit:   60ccb22
 Built:        Thu Feb 23 10:40:59 2017
 OS/Arch:      darwin/amd64

Server:
 Version:      17.03.0-ce
 API version:  1.26 (minimum version 1.12)
 Go version:   go1.7.5
 Git commit:   3a232c8
 Built:        Tue Feb 28 07:52:04 2017
 OS/Arch:      linux/amd64
 Experimental: true
```

shin-:
Yes - I agree it shouldn't be ignored. Raising an error if a float is provided would be the appropriate course of action.

fopinappb:
I know this has been closed long time ago, but maybe it should be revisited as apparently the API does support floats (for added - and useful - precision).

https://github.com/moby/moby/issues/41784#issuecomment-822371593

PR #1648, opened 2017-06-15
Raising error in case of invalid value of since kwarg on Container.logs

ChrsMark:
Resolving https://github.com/docker/docker-py/issues/1515

### DIFF
--- a/docker/api/container.py
+++ b/docker/api/container.py
@@ -825,6 +825,11 @@ class ContainerApiMixin(object):
                         params['since'] = utils.datetime_to_timestamp(since)
                     elif (isinstance(since, int) and since > 0):
                         params['since'] = since
+                    else:
+                        raise errors.InvalidArgument(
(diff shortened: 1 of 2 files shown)


ChrsMark:
@chin could you review?

tests/integration/api_container_test.py
```diff
@@ -809,6 +810,25 @@ def test_logs_with_tail_0(self):
         logs = self.client.logs(id, tail=0)
         self.assertEqual(logs, ''.encode(encoding='ascii'))
 
+    def test_logs_with_since_kwarg(self):
```
shin-: This might make more sense as a unit test, since we really just want to check that the method raises an error when passed invalid arguments, creating an actual container or connecting to the Docker Engine is unnecessary.

ChrsMark:
@shin-, comments addressed, also rebased on top of the latest `upstream/master`

shin-:
Thanks, LGTM!

PR #1653, opened 2017-06-16
Closes #1588, image.tag does not return anything

### LINKED ISSUE #1588: image.tag does not return anything
There's no return statement in `images.tag`:
https://github.com/docker/docker-py/blob/master/docker/models/images.py#L99

[Readthedocs](https://docker-py.readthedocs.io/en/stable/images.html) (and the method comments) suggest it should return a bool for success.

I saw this running version 2.2.1 of the library
```
# pip freeze | grep docker
docker==2.2.1
docker-pycreds==0.2.1
```

**Repro code:**
```
import docker
def test_tag(id):
    client = docker.DockerClient()
    image = client.images.get(id)
    tag_result = image.tag('test_image', tag='test_tag')
    if tag_result is None:
        print('oops')
```

osallou:
This patch returns the check made against api when tagging an image as stated in documentation

Signed-off-by: Olivier Sallou <olivier.sallou@irisa.fr>

### DIFF
--- a/docker/models/images.py
+++ b/docker/models/images.py
@@ -96,7 +96,7 @@ class Image(Model):
         Returns:
             (bool): ``True`` if successful
         """
-        self.client.api.tag(self.id, repository, tag=tag, **kwargs)
+        return self.client.api.tag(self.id, repository, tag=tag, **kwargs)
 
 
 class ImageCollection(Collection):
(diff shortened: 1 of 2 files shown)


shin-:
Thanks!

PR #1657, opened 2017-06-16
Return generator for output of load_image endpoint 

shin-:
Fixes #1650

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -5,9 +5,8 @@ def imageNamePy2
 def imageNamePy3
 def images = [:]
 
-// Note: Swarm in dind seem notoriously flimsy with 1.12.1+, which is why we're
-// sticking with 1.12.0 for the 1.12 series
-def dockerVersions = ["1.13.1", "17.04.0-ce", "17.05.0-ce", "17.06.0-ce-rc5"]
+
+def dockerVersions = ["1.13.1", "17.04.0-ce", "17.05.0-ce", "17.06.0-ce"]
 
 def buildImage = { name, buildargs, pyTag ->
   img = docker.image(name)
--- a/docker/api/image.py
+++ b/docker/api/image.py
@@ -262,7 +262,7 @@ class ImageApiMixin(object):
             self._get(self._url("/images/{0}/json", image)), True
         )
 
-    def load_image(self, data):
+    def load_image(self, data, quiet=None):
         """
         Load an image that was previously saved using
         :py:meth:`~docker.api.image.ImageApiMixin.get_image` (or ``docker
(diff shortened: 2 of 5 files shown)


cecton:
@shin- sorry to bring you bad news, I think you broke the `load_image`. I just tried to load `b'000'` which obviously should fail and since the check of the status is not done, the call works but nothing is loaded. You should actually call `_raise_for_status` **before** returning with `_stream_helper`.

I also noticed in `_stream_helpers` that the behavior is not consistent:
 *  If the response is chunked, the status code is not checked.
 *  If the response is not chunked, the status code is checked.

You may also want to check #1496 because the behavior is even more a bit trickier if I recall.

@shin- I can make a PR to fix all of this if you want.

shin-:
@cecton Behavior is the same with `_raise_for_status` called before. If an error occurs while loading, the API has already started streaming the response, and the 200 status will already be in the header. I tested this locally just to be sure:

```diff
$ git diff
diff --git a/docker/api/image.py b/docker/api/image.py
index 41cc267..02b99b0 100644
--- a/docker/api/image.py
+++ b/docker/api/image.py
@@ -292,10 +292,13 @@ class ImageApiMixin(object):
         res = self._post(
             self._url("/images/load"), data=data, params=params, stream=True
         )
+
+        self._raise_for_status(res)
+
         if utils.version_gte(self._version, '1.23'):
             return self._stream_helper(res, decode=True)
 
-        self._raise_for_status(res)
+        return None
 
     @utils.minimum_version('1.25')
     def prune_images(self, filters=None):
```
```
$ python setup.py develop
/home/joffrey/.envs/pydocker/local/lib/python2.7/site-packages/setuptools/dist.py:341: UserWarning: Normalizing '2.6.0-dev' to '2.6.0.dev0'
  normalized_version,
running develop
running egg_info
[...]
Finished processing dependencies for docker==2.6.0.dev0
$ python
Python 2.7.9 (default, Apr  2 2015, 15:33:21) 
[GCC 4.9.2] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> import docker
>>> c = docker.APIClient()
>>> strm = c.load_image(b'000')
>>> for x in strm:
...   print(x)
... 
{u'errorDetail': {u'message': u'Error processing tar file(exit status 1): unexpected EOF'}, u'error': u'Error processing tar file(exit status 1): unexpected EOF'}
```

cecton:
@shin- thanks I confirm! I thought the status was varying because there was a `_raise_for_status` call. So apparently the `_raise_for_status` is completely useless. I don't see any 4xx possible error in the doc either: https://docs.docker.com/engine/api/v1.28/#operation/ImageLoad

PR #1659, opened 2017-06-20
Make sure data is written in prune test so space can be reclaimed

shin-:
https://github.com/moby/moby/pull/33735#issuecomment-309910176


jhowardmsft:
@shin It looks like this PR is making Janky and Experimental fail on moby/moby master. Can you investigate? From the name of the commit here, and the failures in moby/moby Jenkins, and the timing of this being merged, this seems to line up. https://github.com/moby/moby/pull/33241#issuecomment-309933669 for more info.

cc @crosbymichael

PR #1689, opened 2017-07-25
Do not try to get logs for not readable logging drivers

### LINKED ISSUE #1688: Running container with syslog log driver fails
Steps to reproduce:

```python
✗ python -i 
>>> import docker
>>> client.containers.run("busybox:latest", "echo hello world", log_config=dict(type='syslog'))
  File "/usr/local/lib/python2.7/site-packages/docker/models/containers.py", line 712, in run
    out = container.logs(stdout=stdout, stderr=stderr)
  File "/usr/local/lib/python2.7/site-packages/docker/models/containers.py", line 235, in logs
    return self.client.api.logs(self.id, **kwargs)
  File "/usr/local/lib/python2.7/site-packages/docker/utils/decorators.py", line 19, in wrapped
    return f(self, resource_id, *args, **kwargs)
  File "/usr/local/lib/python2.7/site-packages/docker/api/container.py", line 839, in logs
    return self._get_result(container, stream, res)
  File "/usr/local/lib/python2.7/site-packages/docker/api/client.py", line 403, in _get_result
    return self._get_result_tty(stream, res, cont['Config']['Tty'])
  File "/usr/local/lib/python2.7/site-packages/docker/api/client.py", line 417, in _get_result_tty
    self._raise_for_status(res)
  File "/usr/local/lib/python2.7/site-packages/docker/api/client.py", line 222, in _raise_for_status
    raise create_api_error_from_http_exception(e)
  File "/usr/local/lib/python2.7/site-packages/docker/errors.py", line 31, in create_api_error_from_http_exception
    raise cls(e, response=response, explanation=explanation)
docker.errors.APIError: 500 Server Error: Internal Server Error ("configured logging driver does not support reading")
```

Running the same command form bash works as expected:

```
✗ docker run --log-driver=syslog busybox:latest echo hello world
hello world
```

The problem may be reproduced in console as well:

```
✗ docker run --name='syslog-test' -d --log-driver=syslog busybox:latest  sleep 60
875d8c9619ea626efbd3a813dbd6550e752c2ff443ad2a7559eab2659e00c528
✗ docker logs syslog-test
Error response from daemon: configured logging driver does not support reading
```

It looks like we should not [try to obtain logs](https://github.com/docker/docker-py/blob/master/docker/models/containers.py#L712) if log driver is overridden. 

My environment:

```
✗ pip freeze | grep docker && python --version && docker version
docker==2.4.2
docker-pycreds==0.2.1
Python 2.7.13
Client:
 Version:      17.06.0-ce
 API version:  1.30
 Go version:   go1.8.3
 Git commit:   02c1d87
 Built:        Fri Jun 23 21:31:53 2017
 OS/Arch:      darwin/amd64

Server:
 Version:      17.06.0-ce
 API version:  1.30 (minimum version 1.12)
 Go version:   go1.8.3
 Git commit:   02c1d87
 Built:        Fri Jun 23 21:51:55 2017
 OS/Arch:      linux/amd64
 Experimental: true
```

OS: macOS Sierra 10.12.5 (16F73)

bolshakov:
Fixes #1688

According to the [documentation](https://docs.docker.com/engine/api/v1.30/#operation/ContainerLogs), logs may be read only if either `json-file` or `journald` logging driver used. However, `client.containers.run` does not respect this rule and tries to read logs after running container.

```python
✗ python -i 
>>> import docker
>>> client.containers.run("busybox:latest", "echo hello world", log_config=dict(type='syslog'))
.....
docker.errors.APIError: 500 Server Error: Internal Server Error ("configured logging driver does not support reading")
```

To overcome this issue, I added logging driver check to `client.containers.run`.

See for details https://github.com/docker/docker-py/issues/1688

After my fix it works as follows:

```python
```python
✗ python -i 
>>> import docker
>>> client.containers.run("busybox:latest", "echo hello world", log_config=dict(type='syslog'))
"Result logged using `syslog` driver"
```

I understand this is not the best solution, but it is definitely better than exception. Any ideas?

### DIFF
--- a/docker/errors.py
+++ b/docker/errors.py
@@ -127,8 +127,14 @@ class ContainerError(DockerException):
         self.command = command
         self.image = image
         self.stderr = stderr
-        msg = ("Command '{}' in image '{}' returned non-zero exit status {}: "
-               "{}").format(command, image, exit_status, stderr)
+
+        if stderr is None:
+            msg = ("Command '{}' in image '{}' returned non-zero exit "
+                   "status {}").format(command, image, exit_status, stderr)
+        else:
+            msg = ("Command '{}' in image '{}' returned non-zero exit "
+                   "status {}: {}").format(command, image, exit_status, stderr)
+
         super(ContainerError, self).__init__(msg)
 
 
--- a/docker/models/containers.py
+++ b/docker/models/containers.py
@@ -667,6 +667,13 @@ class ContainerCollection(Collection):
             The container logs, either ``STDOUT``, ``STDERR``, or both,
             depending on the value of the ``stdout`` and ``stderr`` arguments.
 
+            ``STDOUT`` and ``STDERR`` may be read only if either ``json-file``
+            or ``journald`` logging driver used. Thus, if you are using none of
+            these drivers, a ``None`` object is returned instead. See the
+            `Engine API documentation
+            <https://docs.docker.com/engine/api/v1.30/#operation/ContainerLogs/>`_
+            for full details.
+
             If ``detach`` is ``True``, a :py:class:`Container` object is
             returned instead.
 
@@ -709,7 +716,14 @@ class ContainerCollection(Collection):
         if exit_status != 0:
             stdout = False
             stderr = True
-        out = container.logs(stdout=stdout, stderr=stderr)
+
+        logging_driver = container.attrs['HostConfig']['LogConfig']['Type']
+
+        if logging_driver == 'json-file' or logging_driver == 'journald':
+            out = container.logs(stdout=stdout, stderr=stderr)
+        else:
+            out = None
+
         if remove:
             container.remove()
         if exit_status != 0:
--- a/tests/integration/models_containers_test.py
+++ b/tests/integration/models_containers_test.py
@@ -88,6 +88,24 @@ class ContainerCollectionTest(BaseIntegrationTest):
         assert 'Networks' in attrs['NetworkSettings']
         assert list(attrs['NetworkSettings']['Networks'].keys()) == [net_name]
 
+    def test_run_with_none_driver(self):
+        client = docker.from_env(version=TEST_API_VERSION)
+
+        out = client.containers.run(
+            "alpine", "echo hello",
+            log_config=dict(type='none')
+        )
+        self.assertEqual(out, None)
+
+    def test_run_with_json_file_driver(self):
+        client = docker.from_env(version=TEST_API_VERSION)
+
+        out = client.containers.run(
+            "alpine", "echo hello",
+            log_config=dict(type='json-file')
+        )
+        self.assertEqual(out, b'hello\n')
+
     def test_get(self):
         client = docker.from_env(version=TEST_API_VERSION)
         container = client.containers.run("alpine", "sleep 300", detach=True)
--- a/tests/unit/errors_test.py
+++ b/tests/unit/errors_test.py
@@ -2,8 +2,10 @@ import unittest
 
 import requests
 
-from docker.errors import (APIError, DockerException,
+from docker.errors import (APIError, ContainerError, DockerException,
                            create_unexpected_kwargs_error)
+from .fake_api import FAKE_CONTAINER_ID, FAKE_IMAGE_ID
+from .fake_api_client import make_fake_client
 
 
 class APIErrorTest(unittest.TestCase):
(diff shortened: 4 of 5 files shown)


docker/models/containers.py
```diff
@@ -709,7 +709,15 @@ def run(self, image, command=None, stdout=True, stderr=False,
...
             stderr = True
-        out = container.logs(stdout=stdout, stderr=stderr)
+
+        config = self.client.api.inspect_container(container.id)
+        logging_driver = config['HostConfig']['LogConfig']['Type']
+
+        if logging_driver == 'json-file' or logging_driver == 'journald':
+            out = container.logs(stdout=stdout, stderr=stderr)
+        else:
+            out = "Result logged using `%s` driver" % logging_driver
```
bolshakov: I'm not sure if using any "special" message is a good idea.

docker/models/containers.py
```diff
@@ -709,7 +709,15 @@ def run(self, image, command=None, stdout=True, stderr=False,
         if exit_status != 0:
             stdout = False
             stderr = True
-        out = container.logs(stdout=stdout, stderr=stderr)
+
+        config = self.client.api.inspect_container(container.id)
```
shin-: Rather than make an additional call to the API, we should be able to determine the driver from `kwargs['log_config']`

docker/models/containers.py
shin-: We should set this to `None` instead, and explain the behavior in the docs.

bolshakov:
@shin- Done, review please

docker/errors.py
```diff
@@ -127,8 +127,14 @@ def __init__(self, container, exit_status, command, image, stderr):
...
         self.stderr = stderr
-        msg = ("Command '{}' in image '{}' returned non-zero exit status {}: "
-               "{}").format(command, image, exit_status, stderr)
+
+        if stderr is None:
+            msg = ("Command '{}' in image '{}' returned non-zero exit "
+                   "status {}").format(command, image, exit_status, stderr)
+        else:
+            msg = ("Command '{}' in image '{}' returned non-zero exit "
+                   "status {}: {}").format(command, image, exit_status, stderr)
```
shin-: This change seems unrelated to the issue at hand. Did you run into an issue with this specifically? `format` should work fine with `None` values.

tests/integration/models_containers_test.py
```diff
@@ -88,6 +88,24 @@ def test_run_with_network(self):
...
         assert list(attrs['NetworkSettings']['Networks'].keys()) == [net_name]
 
+    def test_run_with_none_driver(self):
+        client = docker.from_env(version=TEST_API_VERSION)
+
+        out = client.containers.run(
+            "alpine", "echo hello",
+            log_config=dict(type='none')
+        )
+        self.assertEqual(out, 'Result logged using `none` driver')
```
shin-: This probably needs to be updated as a result of the changes.

docker/errors.py
bolshakov: I decided to check how it would be rendered with `None` and found it a bit unfriendly:

    Command 'echo Hello World' in image 'e9aa60c60128' returned non-zero exit status 42: None 

For me it is not clear what does `None` mean in this message.

tests/integration/models_containers_test.py
bolshakov: Good catch

docker/errors.py
bolshakov: @shin- what should I do -- remove it or keep it?

docker/models/containers.py
```diff
@@ -709,7 +716,14 @@ def run(self, image, command=None, stdout=True, stderr=False,
         if exit_status != 0:
             stdout = False
             stderr = True
-        out = container.logs(stdout=stdout, stderr=stderr)
+
+        logging_driver = container.attrs['HostConfig']['LogConfig']['Type']
+
+        if logging_driver == 'json-file' or logging_driver == 'journald':
```
thaJeztah: I don't think this is the right approach; docker now supports logging driver plugins, which may (or may not) support reading. Hard-coding these options here will block that option.

the `500` error returned does not look right to me though, I think that should be fixed

bolshakov: I agree that hardcoding logging drivers is not the best option, however Docker API doesn't provide information about logging driver abilities. Do you think catching `docker.errors.APIError` is better approach?

```python
try:
    out = container.logs(stdout=stdout, stderr=stderr)
except docker.errors.APIError:
    out = None
```

For me the only disadvantage is that we can distinguish different types of api errors only by error message.

thaJeztah: Question is if different logic is needed based on the error; running with docker directly;

```
$ docker run -dit --log-driver=none --name foo busybox

$ docker logs foo
Error response from daemon: configured logging driver does not support reading
```

bolshakov: I consider `client.containers.run` as replacement for `docker run` command. Docker run command does not fail if you run it with non readable driver. 

If you need different logic based on the error you are always able to run two commands:

```python
container = client.containers.run("busybox", detach=True, log_config=dict(type='none'))
container.logs()
.....
docker.errors.APIError: 500 Server Error: Internal Server Error ("configured logging driver does not support reading")
```

This pull request does not change `logs()` method behaviour, it affect only `run` method. Currently it is impossible to run attached container with non-readable logging driver.

thaJeztah: I wonder why it's using the logging endpoint, and not attaching to the container (which is what `docker run` does)

bolshakov: Folks, I'm pretty sure there is always the better way doing something. The point of this PR was to overcome the pain of using not readable logging driver. But I'm wonder how we come to discussing why it was written this way in the past?

This is my first contribution in python project. Moreover this is the first time I'm programming python at all. I'm does not feel anouth courage to rewrite the whole method at this point.

shin-: @thaJeztah It always comes down to the same issue - the API direly needs error codes if we want to be able to handle these things gracefully. Until then, I'm afraid hardcoded values are the best solution here. Especially in this case, `run`'s purpose is to be a convenience method that imitates `docker run` - for more fine-grained control, people will still be able to call `logs` on the container directly and handle the possible exception themselves.

As for `logs` over `attach`, the reason is 2-fold:
1. We don't have an interactive mode, so both are functionally equivalent
2. Historically `logs` works better than `attach`

docker/errors.py
shin-: No, it's all good, I was just curious. Thank you!

shin-:
Rebased and merged manually: b8fd8213364ae21c2981f3e51cee924738cb9b93

Thank you!

PR #1694, opened 2017-07-31
client.networks.create check_duplicates docs not reflective of behavior

### LINKED ISSUE #1693: client.networks.create check_duplicates docs not reflective of behavior
Docs say it does, but it's actually set to `None`.

dimaspivak:
Fixes #1693


shin-:
We want to defer the default value to the Engine if possible, so I think we should keep `None` here, and update the docs instead to reflect the actual behavior.

dimaspivak:
My concern is that, as it stands, duplicate networks get created (by default) whereas they're not created by default from the Docker CLI. But if that's the expected behavior, we can close this.

shin-:
I think that's fine - the definitive reference is the engine API, not the CLI.

shin-:
Thanks!

PR #1695, opened 2017-08-01
Return the result of the API when using remove_image and load_image

cecton:
Those calls return result that can be used by the developers.

Signed-off-by: Cecile Tonglet <cecile.tonglet@gmail.com>

### DIFF
--- a/docker/api/image.py
+++ b/docker/api/image.py
@@ -272,7 +272,7 @@ class ImageApiMixin(object):
             data (binary): Image data to be loaded.
         """
         res = self._post(self._url("/images/load"), data=data)
-        self._raise_for_status(res)
+        return self._result(res, True)
 
     @utils.minimum_version('1.25')
     def prune_images(self, filters=None):
@@ -455,7 +455,7 @@ class ImageApiMixin(object):
         """
         params = {'force': force, 'noprune': noprune}
         res = self._delete(self._url("/images/{0}", image), params=params)
-        self._raise_for_status(res)
+        return self._result(res, True)
 
     def search(self, term):
         """
--- a/tests/integration/api_image_test.py
+++ b/tests/integration/api_image_test.py
@@ -113,7 +113,8 @@ class RemoveImageTest(BaseAPIIntegrationTest):
         self.assertIn('Id', res)
         img_id = res['Id']
         self.tmp_imgs.append(img_id)
-        self.client.remove_image(img_id, force=True)
+        logs = self.client.remove_image(img_id, force=True)
+        self.assertIn({"Deleted": img_id}, logs)
         images = self.client.images(all=True)
         res = [x for x in images if x['Id'].startswith(img_id)]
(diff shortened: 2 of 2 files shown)


cecton:
The travis check is broken:
```
$ python --version
Python 3.5.3
(...)
py33 create: /home/travis/build/docker/docker-py/.tox/py33
ERROR: InterpreterNotFound: python3.3
```

docker/api/image.py
```diff
@@ -456,6 +457,7 @@ def remove_image(self, image, force=False, noprune=False):
         params = {'force': force, 'noprune': noprune}
         res = self._delete(self._url("/images/{0}", image), params=params)
         self._raise_for_status(res)
```
shin-: The `_result` method includes this check, so this line can be removed

docker/api/image.py
```diff
@@ -273,6 +273,7 @@ def load_image(self, data):
         """
         res = self._post(self._url("/images/load"), data=data)
         self._raise_for_status(res)
+        return self._result(res, True)
```
shin-: This will conflict with the changes in #1657, which is a more thorough fix

tests/integration/api_image_test.py
```diff
@@ -103,6 +103,29 @@ def test_commit_with_changes(self):
         assert img['Config']['Cmd'] == ['bash']
 
 
+class LoadImageTest(BaseAPIIntegrationTest):
```
shin-: This is also addressed in #1657 and can safely be removed.

tests/integration/api_image_test.py
```diff
@@ -113,7 +136,8 @@ def test_remove(self):
         self.assertIn('Id', res)
         img_id = res['Id']
         self.tmp_imgs.append(img_id)
-        self.client.remove_image(img_id, force=True)
+        logs = self.client.remove_image(img_id, force=True)
+        self.assertIsInstance(logs, list)
```
shin-: We may be able to add a check for an `{'Deleted': img_id}` entry in that list.

docker/api/image.py
cecton: Ah yeah actually I knew that, thanks!

docker/api/image.py
cecton: Cool!

cecton: Ah but in #1667 it doesn't return the result, it only raises (which is the point of my PR). Either you merge #1667 first and I fix the conflict or you can update #1667 by replacing the `self._raise_for_status` by the `self._result`.

cecton:
ping @shin- (look also here https://github.com/docker/docker-py/pull/1695#discussion_r131831655 because I didn't remove the change on `load_image` yet)

shin-:
LGTM! Just one last thing - do you mind squashing those commits together? Thanks!

cecton:
@shin- done!

shin-:
Thank you!

PR #1698, opened 2017-08-06
Added wait to the Container class documentation.

AndreasBackx:
The container class documentation did not automatically document the `Container.wait` method.

Signed-off-by: Andreas Backx <andreas@backx.org>


shin-:
Good catch, thank you!

PR #1699, opened 2017-08-07
Fix domainname documentation in create_container function

terminalmage:
It looks like this was probably originally copypasta'ed from `dns_search` and not edited afterward.

Signed-off-by: Erik Johnson <palehose@gmail.com>


shin-:
Thanks!

Issue #1571, opened 2017-04-21
Networks option does not support aliases in create service


hussam-almarzoq:

lyager:
I've added a pull-reguest for the issue here: https://github.com/docker/docker-py/pull/1573

hussam-almarzoq:
@lyager But this does not help with services .

lyager:
@hussam-almarzoq Could you maybe provide a codesnippet?

hussam-almarzoq:
@lyager The APIClient has a method called create_container. This method has a parameter named networking_config where I can attach the container to networks and add aliases. This parameter is not available to the high level containers api nor the swarm services api even in low level ones.

lyager:
What I did using the above patch was (without using the underlaying API):

Build the containers needed: client.build 
Create the containers: client.create
Attach containers to network, using aliases: network.connect(container, aliases=[])
And start the container: container.start()

Not a solution to you API request, I agree, but maybe as a workaround?

hussam-almarzoq:
@lyager Works good enough with containers. But what about swarm services ?

lyager:
@hussam-almarzoq Haven't used swarm services much. Sorry

overmike:
Is there any update on this issue?

to whom may interest, I circumvent by using the engine api networkattachment format
`docker_client.services.create(....., networks=[{'Target': 'data', 'Aliases': ['mysql']}])`

PR #1712, opened 2017-08-16
Fix - #1571 Network Alias needs to be populated in service JSON.

watbulb:
**Fix - #1571 If service in VIP mode, Network Aliases needs to be populated with hostname.**

When using the standard docker CLI utility to deploy a docker stack `(docker stack deploy)` a JSON is always generated with an Aliases argument in the ["TaskTemplate"]["Networks"] dictionary when deployed in VIP mode.

Example shown here of JSON output from `docker stack deploy` CLI deployment.
```
"TaskTemplate" : {
    "Networks": [
      {
        "Target": "wlydwxqc5hwprwpktofo8le68",
        "Aliases": [
          "manager"]
      }],
}
```

Example shown here using high-level docker-py service creation `(client.service.create)`

```
"TaskTemplate" : {
    "Networks": [
      {
        "Target": "wlydwxqc5hwprwpktofo8le68"
      }]
}
```

Without Aliases in VIP mode, `dockerd` internal DNS resolver will be unable to resolve adjacent services on adjacent nodes in the swarm configuration. This has been verified with packet capture.

If you would like to reproduce this for yourselves, deploy a two service stack using the command line on a overlay network driver, view the `service.attrs` and check the Networks within TaskTemplate.

Next, deploy this same stack using docker-py in VIP mode on a overlay network, Aliases will not appear in `service.attrs` and internal DNS resolution will fail.

This hack is pretty ugly, fixup as you wish, but please fix.

### DIFF
--- a/docker/api/build.py
+++ b/docker/api/build.py
@@ -18,7 +18,8 @@ class BuildApiMixin(object):
               custom_context=False, encoding=None, pull=False,
               forcerm=False, dockerfile=None, container_limits=None,
               decode=False, buildargs=None, gzip=False, shmsize=None,
-              labels=None, cache_from=None, target=None, network_mode=None):
+              labels=None, cache_from=None, target=None, network_mode=None,
+              squash=None):
         """
         Similar to the ``docker build`` command. Either ``path`` or ``fileobj``
         needs to be set. ``path`` can be a local path (to a directory
@@ -98,6 +99,8 @@ class BuildApiMixin(object):
                 Dockerfile
             network_mode (str): networking mode for the run commands during
                 build
+            squash (bool): Squash the resulting images layers into a
+                single layer.
 
         Returns:
             A generator for the build output.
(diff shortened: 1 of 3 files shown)


daytonpid:
I would also like to reference #1571 and #1769.

Dear maintainers, 3 Issues is a bit much spanning over 6 months :-1: for a critical usability bug.

ulyssessouza:
Hello @daytonpid, thank you for the PR.
Is this still pertinent?

watbulb:
@ulyssessouza 
Unfortunately last I checked 5 or so months ago, it still existed, but the fix was no longer useful on later docker API builds as a-lot of the codebase had changed in that area, so we back ported newer versions internally to fix this issue.

I tried so many times to get in touch with Docker and @shin-; and have heard so many complaints on this issue, that certain individuals just gave up. And to be honest, completely soured me to "Docker" as a product, as this was a critical usability bug, that nobody wanted to pay attention to, while @shin- still fixed other things, apparently this wasn't as important or would ever be to this project.

ifokeev:
@watbulb haha. Getting the problem with `Aliases` (which are `None`) and I see there is no hope for the fix

watbulb:
@ifokeev Yes unfortunately 😞, welcome to docker... your problems do not exist here...

PR #1713, opened 2017-08-17
Fix handling of non-multiplexed (TTY) streams over upgraded sockets

shin-:
Replaces #1666


daytonpid:
@shin- 

Thank you very much for this, was running into a wall here too. Now we wait :)

PR #1739, opened 2017-09-10
Return Image objects on image.load

hongbin:
In before, image.load returns what Docker API returns, which is
a text stream. This commits propose an improvement for returning
more useful information, which is a list of Image objects being
loaded.


shin-:
Thank you!

PR #1751, opened 2017-09-18
Adding swarm id_attribute to match docker output

SteggyLeggy:
Swarm id is returned in a attribute with the key ID and the current swarm model was using the default behaviour and looking for Id. This was causing swarm.id to be None and swarm.short_id to error.


shin-:
LGTM, thanks!

PR #1756, opened 2017-09-24
Add mac_address to connect_container_to_network

hongbin:


shin-:
Thanks! Do you mind adding the appropriate version check?

hongbin:
@shin- Done

shin-:
I tried adding a test for this, but I don't think this is supported by the engine yet as no matter what I do, I'm not able to set a different MAC address for the network. This is reinforced by the fact that there is currently no CLI option to achieve this: https://github.com/moby/moby/issues/33715

Until this actually works with the Docker Engine, I think we probably shouldn't merge this PR.

hongbin:
@shin- when you said "I'm not able to set a different MAC address for the network", could you provide the exact reproducing steps? I will follow up with that.

PR #1776, opened 2017-10-17
Fix simple documentation copy/paste error.

janLo:
It seems that the documentation of inspect_image was copy/pasted from some place where containers where expected.


shin-:
Thank you!

PR #1778, opened 2017-10-18
Make check_duplicate network flag true by default.

dimorinny:
In your docs said that checking duplicate for creating network is forced by default. (https://docker-py.readthedocs.io/en/stable/networks.html)
But actually, it isn't true. I fixed this by changing default param in api client.
Also I added functional and unit tests for this case.

Signed-off-by: Dmitry Merkurev <didika914@gmail.com>

### DIFF
--- a/docker/api/network.py
+++ b/docker/api/network.py
@@ -17,7 +17,7 @@ class NetworkApiMixin(object):
                 Available filters:
                 - ``driver=[<driver-name>]`` Matches a network's driver.
                 - ``label=[<key>]`` or ``label=[<key>=<value>]``.
-                - ``type=["custom"|"builtin"]`` Filters networks by type.
+                - ``type=['custom'|'builtin']`` Filters networks by type.
 
         Returns:
             (dict): List of network objects.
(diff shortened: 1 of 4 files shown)


shin-:
Thank you, but I think the docs should be updated to reflect the code's behavior, not the other way around.

And what's the deal with all the quote changes?

dimorinny:
Hi.

**About docs:** no problems, I will generate it.
**About quotes:** I found, that non consistent quotes are using on this project and  I decide to change that in all files, that I changed. If it is confused you, I can rollback this changes.

dimorinny:
**About docs again:** Hm, I didn't get it. What should I do for updating docs?

On this pull request I changed only the internal low level API for creating network (https://docker-py.readthedocs.io/en/stable/api.html#networks) and I changed it on doc string. This is not enough for updating docs?

shin-:
Ah sorry, I realize what's happened now ; the docstring was updated by https://github.com/docker/docker-py/pull/1694, but because we haven't released a new version since, the docs on the website still have the incorrect version.

Since #1694 has been merged, no additional change needs to be made.

Thank you!

Issue #1785, opened 2017-11-01
Missing doc for valid service-list filter keys


baldoalessandro:
The doc of both the [high-level](http://docker-py.readthedocs.io/en/2.5.1/services.html#docker.models.services.ServiceCollection.list) and the [low-level](http://docker-py.readthedocs.io/en/2.5.1/api.html#docker.api.service.ServiceApiMixin.services) API for listing docker services miss some valid filter options.
In particular both `docker.api.services` and `docker.services.list` document as valid filter keys only `id` and `name`.

The docker engine API documents, at least from [v1.24](https://docs.docker.com/engine/api/v1.24/#39-services) (which by the way was also the first version supporting _services_), that the available filters are:
- `id`
- `name`
- `label`

These filter keys are AFAIK are supported by each version from v1.24 to  the current version.

In  [docker engine v1.29](https://docs.docker.com/engine/api/v1.29/#tag/Service) another filter key was added: `mode`. This key is present from that version to the current.

I don't know why those filter key was not included in the docs but I've tested on a docker swarm instance and they actually works.

Here's my conf:
```bash
$ docker.version
> '2.3.0'
$ docker.api.version()
> {
>   'Version': '17.09.0-ce',
>   'ApiVersion': '1.32',
>   'MinAPIVersion': '1.12',
>   'GitCommit': 'afdb6d4',
>   'GoVersion': 'go1.8.3',
>   'Os': 'linux', 'Arch': 'amd64', 'KernelVersion': '4.9.49-moby',
>   'Experimental': True,
>   'BuildTime': '2017-09-26T22:45:38.000000000+00:00'
> }
```

PR #1786, opened 2017-11-01
Improve docs for service list filters

baldoalessandro:
Referencing issue #1785 

- add "label" to the list of available filter keys in the high-level service API
- add "label" to the list of available filter keys in the low-level service API
- add integration tests

Signed-off-by: Alessandro Baldo <git@baldoalessandro.net>

### DIFF
--- a/docker/api/service.py
+++ b/docker/api/service.py
@@ -201,7 +201,8 @@ class ServiceApiMixin(object):
 
         Args:
             filters (dict): Filters to process on the nodes list. Valid
-                filters: ``id`` and ``name``. Default: ``None``.
+                filters: ``id``, ``name`` , ``label`` and ``mode``.
+                Default: ``None``.
 
         Returns:
             A list of dictionaries containing data about each service.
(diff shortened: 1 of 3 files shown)


shin-:
Thanks! Do you mind adding the `mode` key to the docs as well? It doesn't need a test, just a mention in the docstring.

baldoalessandro:
Sure, no problem! 
I didn't add the `mode` key because I wasn't sure if it needs some kind of version check in the code.
In particular `docker.api.services` is defined like:

```python
@utils.minimum_version('1.24')
def services(self, filters=None):
     # [ ... ]
```
but the `mode` key in service filter is only supported from [1.29](https://docs.docker.com/engine/api/v1.29/#tag/Service). Should I add some version check like:
```python
if utils.version_lt(version, '1.29'):
  raise errors.InvalidVersion(
    'mode filter is not supported in API version < 1.29'
  )
```

Thanks for the project by the way! 👍

shin-:
Our policy has been to not make version checks on filters, since we're often a bit behind engine releases, it'd cause valid filters to be refused until we add them to the list of accepted parameters, making the experience worse for users in the long run.
So no, just a mention in the docstring will be sufficient :) Thank you for being thorough!

baldoalessandro:
Oh right, understood. I've added the `mode` key

PR #1788, opened 2017-11-02
 Prevent data loss when attaching to container

shin-:
Based on #1770 

@cjh1 Do you mind testing this PR and see if it solves your issue as well?

### DIFF
--- a/docker/transport/unixconn.py
+++ b/docker/transport/unixconn.py
@@ -18,7 +18,20 @@ except ImportError:
 RecentlyUsedContainer = urllib3._collections.RecentlyUsedContainer
 
 
+class UnixHTTPResponse(httplib.HTTPResponse, object):
+    def __init__(self, sock, *args, **kwargs):
+        disable_buffering = kwargs.pop('disable_buffering', False)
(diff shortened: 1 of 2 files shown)


docker/transport/unixconn.py
```diff
@@ -18,21 +18,46 @@
...
     def connect(self):
         sock = socket.socket(socket.AF_UNIX, socket.SOCK_STREAM)
         sock.settimeout(self.timeout)
         sock.connect(self.unix_socket)
         self.sock = sock
 
+    def putheader(self, header, *values):
+        super(UnixHTTPConnection, self).putheader(header, *values)
+        if header == 'Connection' and 'Upgrade' in values:
+            self.disable_buffering = True
```
cjh1: This is a much nicer approach, thanks.

cjh1:
@shin- I can confirm that this also resolves me issue.

shin-:
Great, thank you for confirming!

cjh1:
@shin- Any idea when the next release will be? Thanks.

shin-:
Should come out in the next couple of weeks 👍

cjh1:
Awesome, thanks for helping me get this in.

PR #1797, opened 2017-11-07
Fix #1381: Add exit code to exec_run

### LINKED ISSUE #1381: How to get exec_run exit_code?

hnq90:
As described in https://github.com/docker/docker-py/issues/1381, we need to get the exit code after executing a command by `exec_run`.
So I change the returning of `exec_run` to return a dictionary which has `output` and `exit_code`.

Please help me to review this PR. Thank you in advance.

### DIFF
--- a/docker/models/containers.py
+++ b/docker/models/containers.py
@@ -149,10 +149,14 @@ class Container(Model):
                 ``{"PASSWORD": "xxx"}``.
 
         Returns:
-            (generator or str):
-                If ``stream=True``, a generator yielding response chunks.
-                If ``socket=True``, a socket object for the connection.
-                A string containing response data otherwise.
+            dict:
+                output: (generator or str):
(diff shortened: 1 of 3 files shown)


shin-:
Thank you!

PR #1804, opened 2017-11-10
Fix #1660 : Move Networks to TaskTemplate

### LINKED ISSUE #1660: Unable to subsequently update Docker Service when using 'networks'.
Since a few weeks it's no longer possible to perform a call to update() when networks are being defined. For instance:

service.update( name = service.name
                         , image = updated_image
                         , networks = networks
                         #  ... omitted other parameters.
                         )

Previously would refresh the service, only modifying the image. Now it throws the following error:
"rpc error: code = 12 desc = networks must be migrated to TaskSpec before being changed"

Please note, that the first update() works, this only occurs after performing an update() again on a service which was previously updated using the update() function.

For us this is critical, as currently rolling deployments (nightly builds) are no longer being deployed automatically.

rycus86:
This change should fix #1660 to avoid the `networks must be migrated to TaskSpec before being changed` error.
I have changed the API client to put the `Networks` on the `TaskTemplate` object and also added it as a parameter to the `TaskTemplate` class.
The new test should cover each of these changes.
I'm actually switching networks in the test and that seems to need version 1.29 of the API at least.

Thanks!

### DIFF
--- a/docker/api/service.py
+++ b/docker/api/service.py
@@ -368,7 +368,11 @@ class ServiceApiMixin(object):
             data['UpdateConfig'] = update_config
 
         if networks is not None:
-            data['Networks'] = utils.convert_service_networks(networks)
+            if task_template is None:
+                service_spec = self.inspect_service(service)['Spec']
+                data['TaskTemplate'] = service_spec['TaskTemplate']
+            api_networks = utils.convert_service_networks(networks)
+            data['TaskTemplate']['Networks'] = api_networks
         if endpoint_spec is not None:
             data['EndpointSpec'] = endpoint_spec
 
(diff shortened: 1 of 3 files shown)


rycus86:
Hi @shin- ,

Sorry, had some trouble with the tests but should be OK now hopefully.

One thing I'm unsure of: this change technically enables specifying networks on both the TaskTemplate and on the API client call. Should I check this in the `_check_api_features`?

In all honesty, if this PR goes through, I'd like to have a look at the `Service.update` method to allow only specifying attributes we want to change and just read the rest from the previous spec - if you think that's a good idea. If yes, I could just do this check there along with some others perhaps.

Thanks!

rycus86:
This change become part of https://github.com/docker/docker-py/pull/1807, please have a look at that one instead.
Thanks!

PR #1807, opened 2017-11-14
Update service from prev spec

### LINKED ISSUE #1660: Unable to subsequently update Docker Service when using 'networks'.

rycus86:
Allowing the service update to reuse attributes from the current service spec.

With this change, the service model update would use this by default but not the API update.
Any arguments coming as `None` should be replaced with values from the previous spec.
To delete a previously defined one, you'd need to pass in the default value (`{}` for labels for example).

This change should also fix #1660 to avoid the `networks must be migrated to TaskSpec before being changed` error.
Because of this, the PR also makes https://github.com/docker/docker-py/pull/1804 unnecessary as this one has all its changes.

This change technically enables specifying networks on both the TaskTemplate and on the API client call. I wonder if I should check this in the _check_api_features?

Please let me know if this change is any good/useful.

Thanks!

### DIFF
--- a/docker/api/service.py
+++ b/docker/api/service.py
@@ -428,7 +428,8 @@ class ServiceApiMixin(object):
                     if 'ContainerSpec' not in merged:
                         merged['ContainerSpec'] = {}
                     for cs_key, cs_value in override['ContainerSpec'].items():
-                        merged['ContainerSpec'][cs_key] = cs_value
-                else:
+                        if cs_value is not None:
+                            merged['ContainerSpec'][cs_key] = cs_value
+                elif ts_value is not None:
                     merged[ts_key] = ts_value
         return merged
--- a/docker/api/service.py
+++ b/docker/api/service.py
@@ -62,6 +62,21 @@ def _check_api_features(version, task_template, update_config):
                     raise_version_error('ContainerSpec.privileges', '1.30')
 
 
+def _merge_task_template(current, override):
+    merged = current.copy()
+    if override is not None:
+        for ts_key, ts_value in override.items():
+            if ts_key == 'ContainerSpec':
+                if 'ContainerSpec' not in merged:
+                    merged['ContainerSpec'] = {}
+                for cs_key, cs_value in override['ContainerSpec'].items():
+                    if cs_value is not None:
+                        merged['ContainerSpec'][cs_key] = cs_value
+            elif ts_value is not None:
+                merged[ts_key] = ts_value
+    return merged
+
+
 class ServiceApiMixin(object):
     @utils.minimum_version('1.24')
     def create_service(
@@ -306,7 +321,7 @@ class ServiceApiMixin(object):
     def update_service(self, service, version, task_template=None, name=None,
                        labels=None, mode=None, update_config=None,
                        networks=None, endpoint_config=None,
-                       endpoint_spec=None, use_current_spec=False):
+                       endpoint_spec=None, fetch_current_spec=False):
         """
         Update a service.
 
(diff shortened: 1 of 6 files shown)


kleptog:
I'm really glad this issue is getting some attention.

docker/api/service.py
```diff
@@ -306,7 +306,7 @@ def tasks(self, filters=None):
     def update_service(self, service, version, task_template=None, name=None,
                        labels=None, mode=None, update_config=None,
                        networks=None, endpoint_config=None,
-                       endpoint_spec=None):
+                       endpoint_spec=None, use_current_spec=False):
```
shin-: Can we rename this to `fetch_current_spec`, so it's clearer that this will result in an additional roundtrip to retrieve the spec?

docker/api/service.py
```diff
@@ -328,6 +328,8 @@ def update_service(self, service, version, task_template=None, name=None,
                 the service to. Default: ``None``.
             endpoint_spec (EndpointSpec): Properties that can be configured to
                 access and load balance a service. Default: ``None``.
+            use_current_spec (boolean): Use the undefined settings from the
+                previous specification of the service. Default: ``False``
```
shin-: I'd suggest `s/previous/current` to keep the terminology consistent here.

docker/api/service.py
```diff
@@ -345,35 +347,89 @@ def update_service(self, service, version, task_template=None, name=None,
 
         _check_api_features(self._version, task_template, update_config)
 
+        if use_current_spec:
+            if utils.version_lt(self._version, '1.29'):
+                inspect_defaults = None
+            else:
+                inspect_defaults = True
```
shin-: nit:
```python
inspect_defaults = None
if utils.version_lt(self._version, '1.29'):
  inspect_defaults = True
```

docker/api/service.py
```diff
@@ -345,35 +347,89 @@ def update_service(self, service, version, task_template=None, name=None,
...
+            current = {}
+
         url = self._url('/services/{0}/update', service)
         data = {}
         headers = {}
+
         if name is not None:
             data['Name'] = name
+        else:
+            data['Name'] = current.get('Name')
```
shin-: We can also shorten these statements to
```python
data['Name'] = current.get('Name') if name is None else name
```

docker/api/service.py
```diff
@@ -345,35 +347,89 @@ def update_service(self, service, version, task_template=None, name=None,
...
+            data['EndpointSpec'] = current.get('EndpointSpec')
 
         resp = self._post_json(
             url, data=data, params={'version': version}, headers=headers
         )
         self._raise_for_status(resp)
         return True
+
+    @staticmethod
+    def _merge_task_template(current, override):
```
shin-: I'd like this to be a module-level function like `check_api_features`, rather than a static method.

docker/api/service.py
rycus86: I think you meant this, right?
```python
inspect_defaults = True
if utils.version_lt(self._version, '1.29'):
  inspect_defaults = None
```

rycus86:
Thank you very much for the review @shin- !

shin-:
Awesome, thank you!

Issue #1364, opened 2016-12-20
Add support to disable openssl verification for request get 


fjhaveri:
Add support to disable openssl verification for request get response.

shin-:
This is a very late answer, but this can already be donen thusly (if anyone else stumbles upon this issue):

```python
from docker.tls import TLSConfig
from docker import DockerClient

tls_cfg = TLSConfig(verify=False)
client = DockerClient(tls=tls_cfg)
```

PR #1810, opened 2017-11-16
Add 'verify' parameter to APIClient

withsmilo:
'verify' parameter is used for requests to control whether we verify the server's TLS certificate.
If you want to enable TLS, but to disable SSL verification,
```
client = DockerClient(tls=True, verify=False)
```
This pull request resolved [#1364](https://github.com/docker/docker-py/issues/1364).

### DIFF
--- a/docker/api/client.py
+++ b/docker/api/client.py
@@ -83,6 +83,9 @@ class APIClient(
             ``True`` to enable it with default options, or pass a
             :py:class:`~docker.tls.TLSConfig` object to use custom
             configuration.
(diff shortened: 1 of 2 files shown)


shin-:
Thank you for your contribution, but no change is needed - this can already be done:

```python
from docker.tls import TLSConfig
from docker import DockerClient

tls_cfg = TLSConfig(verify=False)
client = DockerClient(tls=tls_cfg)
```

withsmilo:
@shin- : Great! Thank you!

PR #1811, opened 2017-11-16
Adding properties on the Service model and related types

rycus86:
Hi,

This change adds properties to the `Service` model and its related types mirroring the fields in the specification.
Hope this makes it easier to work with services.

Thanks!

### DIFF
--- a/docker/models/services.py
+++ b/docker/models/services.py
@@ -1,6 +1,8 @@
 import copy
 from docker.errors import create_unexpected_kwargs_error
-from docker.types import TaskTemplate, ContainerSpec
+from docker.types import (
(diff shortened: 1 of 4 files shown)


shin-:

I'm not sure how to feel about this to be honest. It's going to be a hassle to maintain as more properties constantly get added, and I'm not sure the payoff is significant enough to justify it.

rycus86:
Thank you for looking at it!

I see your point, it might be a bit painful to keep these in sync with the API changes. I have a couple of projects using docker-py and this would have come in handy I thought. :)

One alternative I could think of would be making it easier to add/maintain the properties using some Python-magic class modification so you'd only have a mapping of some sort to update.
I'm happy to code it up to see how that would work, otherwise I'm OK to get this closed - it's more of a convenience for users than something they'd actually need.

Issue #1809, opened 2017-11-15
Cannot pickle objects when using unix socket


andreycizov:
One of my use-cases is loading a list of containers in one process and then pushing them to be processed in another process, using a standard python queue.

While this method works well with the `tcp://` socket, when using the `unix://` socket the pickler will throw the following exception:
```
    queue.put((QUEUE_OBJ, obj), block=block)
  File "<string>", line 2, in put
  File "/usr/local/lib/python3.6/multiprocessing/managers.py", line 756, in _callmethod
    conn.send((self._id, methodname, args, kwds))
  File "/usr/local/lib/python3.6/multiprocessing/connection.py", line 206, in send
    self._send_bytes(_ForkingPickler.dumps(obj))
  File "/usr/local/lib/python3.6/multiprocessing/reduction.py", line 51, in dumps
    cls(buf, protocol).dump(obj)
AttributeError: Can't pickle local object 'UnixAdapter.__init__.<locals>.<lambda>'
```

On one hand I could just push the IDs of the objects downstream, but I need some of the attributes belonging to the Containers as well, and it makes for quite an inconvenient way of sharing them.

I am not sure if that is the expected behaviour; but the TCP protocol works well enough and I can't see a reason why UNIX should not.

shin-:
Can you provide an example I could run locally that exhibits the bug? When you say "a list of containers, do you mean [container objects](http://docker-py.readthedocs.io/en/stable/containers.html#container-objects)? If so, those should definitely not be assumed to be serializable / pickleable. That they currently are in some situations is an accident and not by design.

Please consider using the data dictionaries instead (accessible through the `.attrs` property on `Container` objects)

andreycizov:
All of the docker-py API is based around accessing the methods associated with the objects. How do I use the Container object down the stream in an another process, then? That goes with saying that there are ways to implement that using the docker-py API, but it's not documented.

Another point: urllib3 provides `__setstate__` and `__getstate__` for HTTPAdapter that is used to de/serialize the ConnectionPool used for accessing the API. So serialization does work if you use the S/HTTP, but not if using UnixAdapter.

```python
import docker
from multiprocessing import Manager, Pool

# This will work with https:// but not with unix://
ADDR = ''

def container_list_down_the_line(queue):
    cli = docker.DockerClient(base_url=ADDR)

    for c in cli.containers.list():
         queue.put(c)

def execute_in_subprocess(queue):
    while True:
        c = queue.get()
        c.remove()

[...]

PR #1814, opened 2017-11-20
Fixing issue #1809

andreycizov:
Have fixed the issue in #1809.


stasfilin:
@andreycizov Hi, you need to update commit message here: https://github.com/docker/docker-py/pull/1814/commits/e9c6c86e4dee33408ad2a8cec9ac93f14d663b6b

andreycizov:
@stasfilin Sorry - didn't get the fact that all of the commits I made need to be signed.

andreycizov:
The maintainers of this project object to this pull request as they assume the docker-py objects should not be serializable: https://github.com/docker/docker-py/issues/1809

andreycizov:
@MatthiasLohr I have to admit that I completely disagree with their logic as the only reason we can't do that is the implementation of the UNIX socket adapter. Which is easily fixable (given this pull request).

Let me be clear about that: what's the point of supplying me with the deserialized Container objects in the library, if I still need to copy them to another Container object in my code? You're making us do the job twice instead. 

You could have made the API much simpler by just allowing us to work directly with the dictionaries. But since you preferred to go the way of abstracting them from us - then why do you then actively try to stop us from doing things?

@feliperuhland Given your reply in #1809 I again disagree with the answer since my specific use case was loading a list of objects from the API in bulk, and then passing them down the queue for processing. Docker-py forced me to simply copy the attributes of these objects, which is not optimal.

PR #1817, opened 2017-11-20
Add started_at property.

stasfilin:
Good evening,
For my project, I need started_at information. So, I think it's a good time to add this property.

Best Regards,
Stanislav


shin-:

I think this is too situationally useful to warrant adding it as a model property. The data can easily be obtained otherwise through `.attrs`.

PR #1819, opened 2017-11-21
Add missing call to string format in log message

### LINKED ISSUE #1818: Log message not replacing string format placeholder
[This line](https://github.com/docker/docker-py/blob/e9fab1432b974ceaa888b371e382dfcf2f6556e4/docker/auth.py#L205) in the `parse_auth` function in `auth.py` is using a string format placeholder but is not calling `format` on the string to replace it.

alexvy86:
Fixes #1818 
Signed-off-by: Alejandro Villarreal <alex@alexvy86.com>


shin-:
Thanks!

PR #1825, opened 2017-11-30
Fix common issues with build context creation: inaccessible files and FIFOs

shin-:
Fixes docker/compose#4623
Fixes docker/compose#4396


STRML:
👍  Please merge. This cost us a bunch of time today. Imagine my surprise when I finally find the bug... and the first comment on it (#4623) is my own. 🙄

PR #1826, opened 2017-12-01
Use config.json for detachKeys

mtsmfm:
ref: https://github.com/docker/compose/issues/3311

I want to use another `detachKeys` for docker-compose.

docker-compose has already read ~/.docker/config.json for credential.
I think we should use `detachKeys` as well.

I read CONTRIBUTING.md and I know I should add tests and documents for this change.
But I'd like to make sure this design is acceptable before working.

### TODO

- [x] Make sure this design is acceptable
- [x] Add tests
- [x] Write documents

### DIFF
--- a/docker/api/client.py
+++ b/docker/api/client.py
@@ -32,7 +32,7 @@ from ..errors import (
 )
 from ..tls import TLSConfig
 from ..transport import SSLAdapter, UnixAdapter
-from ..utils import utils, check_resource, update_headers
+from ..utils import utils, check_resource, update_headers, config
 from ..utils.socket import frames_iter, socket_raw_iter
 from ..utils.json_stream import json_stream
 try:
@@ -106,6 +106,7 @@ class APIClient(
         self.headers['User-Agent'] = user_agent
 
         self._auth_configs = auth.load_config()
+        self._general_configs = config.load_general_config()
 
         base_url = utils.parse_host(
             base_url, IS_WINDOWS_PLATFORM, tls=bool(tls)
--- a/docker/api/container.py
+++ b/docker/api/container.py
@@ -66,6 +66,7 @@ class ContainerApiMixin(object):
             container (str): The container to attach to.
             params (dict): Dictionary of request parameters (e.g. ``stdout``,
                 ``stderr``, ``stream``).
+                For ``detachKeys``, ~/.docker/config.json is used by default.
             ws (bool): Use websockets instead of raw HTTP.
 
         Raises:
@@ -78,6 +79,11 @@ class ContainerApiMixin(object):
                 'stderr': 1,
                 'stream': 1
             }
+
+        if 'detachKeys' not in params \
+                and 'detachKeys' in self._general_configs:
+
+            params['detachKeys'] = self._general_configs['detachKeys']
 
         if ws:
             return self._attach_websocket(container, params)
--- a/docker/api/exec_api.py
+++ b/docker/api/exec_api.py
@@ -9,7 +9,7 @@ class ExecApiMixin(object):
     @utils.check_resource('container')
     def exec_create(self, container, cmd, stdout=True, stderr=True,
                     stdin=False, tty=False, privileged=False, user='',
-                    environment=None):
+                    environment=None, detach_keys=None):
         """
         Sets up an exec instance in a running container.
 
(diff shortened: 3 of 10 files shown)


mtsmfm:
Here is a URL for my patched docker-compose:

https://bintray.com/mtsmfm/docker/compose#files

mtsmfm:
@shin- Could you review?

mtsmfm:
> Should we add a detach_keys param to start, attach and exec_start?

The following is my opinion:

👍  : implemented
🤔  : not implemented and I'm not sure it should be implemented
👎  : not implemented and I think it shouldn't be implemented

| method | read config.json | add `detach_keys` keyword argument | note |
| ------- | -------- | ------------- | ------ |
| start | 🤔  | 👎    | [it doesn't accept params](https://github.com/docker/docker-py/blob/bf699c7ef78239f6285276b014dcd451bf029636/docker/api/container.py#L1056-L1059) |
| attach | 🤔  | 👎   | [it doesn't return socket](https://github.com/docker/docker-py/blob/bf699c7ef78239f6285276b014dcd451bf029636/docker/api/container.py#L55) |
| attach_socket | 👍  | 🤔  | [it's better to pass via `params` than to add keyword arg because of current interface](https://github.com/docker/docker-py/blob/bf699c7ef78239f6285276b014dcd451bf029636/docker/api/container.py#L60) |
| exec_start | 👍  | 👍  | |

mtsmfm:
> Is there a way we can test the detachKeys feature programmatically?

I found a way to test and it works on my machine.

https://github.com/docker/docker-py/pull/1826/files#diff-51581c0c732dcf484a455168fe36c731R1212

But I'm not sure it works well on CI and I noticed integration test doesn't run on travis.
How can I find the results of integration test?

mtsmfm:
> Can we move the config.py file into the utils/ subfolder?

✅

shin-:
Hi @mtsmfm , thank you for the update!

The integration tests are being run on Jenkins. You can see the output here: https://jenkins.dockerproject.org/job/docker/job/docker-py/job/PR-1826/2/consoleFull

mtsmfm:
@shin- I hope I finished all todos!
Could you review again?

Somehow jenkins doesn't start build for this change yet though(probably it searches PRs once a day?).

mtsmfm:
I noticed jenkins still failed 🙇 

https://jenkins.dockerproject.org/job/docker/job/docker-py/job/PR-1826/3/console

mtsmfm:
@shin- Hmm, integration test works well on my machine...
I want to debug on Jenkins.
Can I log in?

shin-:
Hi @mtsmfm - sorry about the delay. I can't give you access to Jenkins, but let me take a look at it and see what I can find. I'll update as soon as I find out what the snag is.

shin-:
So, I'm able to reproduce locally as well. Your assumption is that sending the detach sequence to the socket will close the socket (which you test with `assert_socket_closed_with_keys`), but that doesn't actually happen. As far as I can tell, all it does is tell the daemon "stop sending data through that socket". The client side is still responsible for closing the connection.

However, I was able to trigger a `socket.error` (Broken pipe) using this code snippet:

```python
import docker

c = docker.APIClient()
ctnr = c.create_container('busybox', 'cat', detach=True, stdin_open=True, tty=True)
c.start(ctnr)
sock = c.attach_socket({'stdin': True, 'stdout': True, 'stream': True, 'detachKeys': 'ctrl-x'})
sock.send('abc')
print(sock.recv(3))  # should read 'abc' back
s.send(chr(24))  # ctrl-x
s.send('abc')  # error: [Errno 32] Broken pipe
```

You should be able to adapt your tests to use that instead - let me know if you need more help with it. 

By the way, you should be able to run the same tests as Jenkins by running `make test` in the project folder, which will let you check the behavior faster.

mtsmfm:
I found `assert_socket_closed_with_keys` with dockerd booted by dind.

> However, I was able to trigger a socket.error (Broken pipe) using this code snippet:

I can't reproduce.
I guess at that time you didn't use dind.

mtsmfm:
I think the test works well if we use bind-mounting docker socket instead of dind.

shin-:
Sorry for the delay, I had to take care of the Compose release and didn't have much time for SDK issues. I'll take another look at this this week.

shin-:
Merged in #1879 

Thank you so much for your help and patience with this PR!

PR #1828, opened 2017-12-02
fixes create_api_error_from_http_exception()

pkit:
`create_api_error_from_http_exception()` is never tested in the original code
and will fail miserably when fed with empty `HTTPError` object
see fixes in requests for this behaviour: https://github.com/requests/requests/pull/3179

Signed-off-by: Constantine Peresypkin <pconstantine@gmail.com>

### DIFF
--- a/tests/unit/errors_test.py
+++ b/tests/unit/errors_test.py
@@ -3,7 +3,8 @@ import unittest
 import requests
 
 from docker.errors import (APIError, ContainerError, DockerException,
-                           create_unexpected_kwargs_error)
+                           create_unexpected_kwargs_error,
+                           create_api_error_from_http_exception)
 from .fake_api import FAKE_CONTAINER_ID, FAKE_IMAGE_ID
 from .fake_api_client import make_fake_client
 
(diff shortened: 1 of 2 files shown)


pkit:
Spurious test failures??

tests/unit/errors_test.py
```diff
@@ -117,3 +118,18 @@ def test_create_unexpected_kwargs_error_single(self):
     def test_create_unexpected_kwargs_error_multiple(self):
         e = create_unexpected_kwargs_error('f', {'foo': 'bar', 'baz': 'bosh'})
         assert str(e) == "f() got unexpected keyword arguments 'baz', 'foo'"
+
+
+class CreateErrorFromException(unittest.TestCase):
```
shin-: Let's add this test to the `APIErrorTest` class instead. :)

shin-:
The test failure should clear up after you re-push.

tests/unit/errors_test.py
pkit: Done

feliperuhland:
Hi, I believe there was a network problem here: ``docker.errors.APIError: 500 Server Error: Internal Server Error ("Get https://registry-1.docker.io/v2/library/hello-world/manifests/latest: Get https://auth.docker.io/token?scope=repository%3Alibrary%2Fhello-world%3Apull&service=registry.docker.io: net/http: TLS handshake timeout")``

PR #1829, opened 2017-12-05
Fix swarm integration test to set name when updating cluster spec.

anshulpundir:
With https://github.com/docker/swarmkit/commit/0b078d10da563bd8c1d4f3a795a81a5d2dbf5232 swarmkit doesn't allow updating cluster spec name.

### DIFF
--- a/tests/integration/api_swarm_test.py
+++ b/tests/integration/api_swarm_test.py
@@ -110,10 +110,12 @@ class SwarmTest(BaseAPIIntegrationTest):
     def test_update_swarm(self):
         assert self.init_swarm()
         swarm_info_1 = self.client.inspect_swarm()
-        spec = self.client.create_swarm_spec(
-            snapshot_interval=5000, log_entries_for_slow_followers=1200,
-            node_cert_expiry=7776000000000000
-        )
+        # Clients should do a read-modify-write when updating swarmkit objects.
+        spec = swarm_info_1['Spec']
+        spec['Raft']['SnapshotInterval'] = 5000
+        spec['Raft']['log_entries_for_slow_followers'] = 1200
+        spec['CAConfig']['NodeCertExpiry'] = 7776000000000000
+
         assert self.client.update_swarm(
             version=swarm_info_1['Version']['Index'],
             swarm_spec=spec, rotate_worker_token=True


anshulpundir:
@shin- @dperny for review

dperny:
IMHOIRL, instead of changing the test, we should change the code so that it inserts the `default` name.

anshulpundir:
>IMHOIRL, instead of changing the test, we should change the code so that it inserts the default name.

Which code ? @dperny

shin-:
This feels a bit like a stupid API trap - why do we have a _mandatory_ parameter that only accepts a single value in the first place?

dperny:
@shin- you're not wrong.

dperny:
@anshulpundir the `create_swarm_spec` method should be changed to autofill the name with `default`. Really, we should alter swarmkit to automatically fill in that parameter on service create...

anshulpundir:
It was designed to accept any value allowed to be changed. But this is where we're at now. @shin- 

I did consider autofill in swarmkit, but I wanted to avoid hidden behavior. Explicit is always better IMO. Also the contract with swarmkit is read-modify-write, so spec name shouldn't really be changed.

shin-:
Even if you modify the library's code to autofill `name`, it's still an arbitrary breaking change for users of the API in general. I understand rejecting custom names, but I'd strongly recommend letting Swarmkit translate absence of value / null value to `default`.

> I did consider autofill in swarmkit, but I wanted to avoid hidden behavior.

Defaults are filled in everywhere else in the API (e.g. creating a volume without specifying a driver), I wouldn't call that "hidden behavior". Since there's nothing to actually configure here, it's just a frustrating "gotcha" that serves no practical purpose.

anshulpundir:
>Defaults are filled in everywhere else in the API (e.g. creating a volume without specifying a driver), I wouldn't call that "hidden behavior". Since there's nothing to actually configure here, it's just a frustrating "gotcha" that serves no practical purpose.

Setting defaults is OK when doing create(), during update, I prefer more explicit behavior. 

In fact, the need to fill in defaults in this case only arises because of the use of gRPC, which doesn't specify if a field was set in the message. For example, thrift does have mechanism to say which fields were explicitly set.

shin-:
Almost of these tests are for `init_swarm`, though (except [this one](https://github.com/docker/docker-py/pull/1829/files#diff-5045b9ca2e2c5acfebac3a4ce041b23aR118)) - is it reasonable to assume those do not need to be changed then?

I get that the chosen protocol might present additional challenges for doing this right, but that doesn't change the fact that the user experience is atrocious for people who happen to rely on these endpoints.

One of the main reasons for running the SDK tests on moby/moby in the first place was to be able to catch and address breaking changes. I don't feel that updating the tests to match the new, arbitrary constraints really accomplishes that.

tests/integration/api_swarm_test.py
```diff
@@ -45,7 +45,8 @@ def test_init_already_in_cluster(self):
     @requires_api_version('1.24')
     def test_init_swarm_custom_raft_spec(self):
         spec = self.client.create_swarm_spec(
-            snapshot_interval=5000, log_entries_for_slow_followers=1200
+            name='default', snapshot_interval=5000,
```
thaJeztah: Is this change needed? I think moby already sets the default if no name is specified; https://github.com/moby/moby/blob/47b3209f687c240dc2841bcfc147b2ffa341a7fe/daemon/cluster/swarm.go#L460-L464

anshulpundir: You're right. Thanks for pointing this out! @thaJeztah 


The call to init_swarm() below will sanitize it and add the name if its not added.

I didn't recall this and made the change everywhere. I will remove it and only keep it only for test_update_swarm() where init_swarm() is not called.

tests/integration/api_swarm_test.py
```diff
@@ -111,7 +116,8 @@ def test_update_swarm(self):
         assert self.init_swarm()
         swarm_info_1 = self.client.inspect_swarm()
         spec = self.client.create_swarm_spec(
-            snapshot_interval=5000, log_entries_for_slow_followers=1200,
+            name='default', snapshot_interval=5000,
+            log_entries_for_slow_followers=1200,
```
anshulpundir: Should be we doing an init_swarm() here before calling update() ? @thaJeztah

shin-: `init_swarm` is called here: https://github.com/docker/docker-py/pull/1829/files#diff-5045b9ca2e2c5acfebac3a4ce041b23aR116

anshulpundir: Right, I guess I meant doing init_swarm() with a name specified. But I guess it wouldn't matter. Keeping this change and reverting others.

thaJeztah: Can we make the change in moby/moby so that the default is set there if omitted? It would still produce an error if any name _other than_ "default" is given, but at least it won't be a breaking change for users of the API.

TL;DR no changes in the docker-py tests should be needed, and if there are, it means we introduced a breaking change

anshulpundir: I'm OK with setting the default name in create_swarm_spec().

The correct semantics for updating swarmkit objects is read-modify-write. Clients should not be writing cluster spec without reading it first i.e. the spec on the client side should be exactly the same as the spec in swarmkit.

anshulpundir:
>One of the main reasons for running the SDK tests on moby/moby in the first place was to be able to catch and address breaking changes. I don't feel that updating the tests to match the new, arbitrary constraints really accomplishes that.

If clients don't do read-modify-write on swarm objects, it's incorrect usage and should be a breaking change. There's nothing wrong with breaking to enforce correct usage.

tests/integration/api_swarm_test.py
thaJeztah: I'm okay with _adding_ a test to test that updating both _with_ and _without_ a name set works, but want to make sure that the current behaviour doesn't break (if `name` was intended to be a required parameter, we should've enforced that long ago; unfortunately we didn't, so we'll have to deal with that)

anshulpundir: Name is *not* a required param. What is required is that you *don't change it*, basically always read it before writing.

Read-modify-write is not really a new thing and I believe has always been the semantic. If clients always do that, then this should not be breaking. 

The problem in the failing test is that it tries to update the spec without reading it first.

PR #1831, opened 2017-12-07
container: fix docstring for containers()

asottile:


asottile:
ugh right, github editor doesn't sign commits

shin-:
Thanks!

PR #1832, opened 2017-12-10
Make the error message of logs() clearer

qazbnm456:
If we assign 0 to `since`, the error message is not clear enough to users (giving it a value of `int` but being told the value isn't within a valid type).

This PR give users clearer message about what is going on.

### DIFF
--- a/docker/api/container.py
+++ b/docker/api/container.py
@@ -838,7 +838,8 @@ class ContainerApiMixin(object):
                         params['since'] = since
                     else:
                         raise errors.InvalidArgument(
-                            'since value should be datetime or int, not {}'.
+                            'since value should be datetime or positive int\
+                            , not {}'.
                             format(type(since))
                         )
             url = self._url("/containers/{0}/logs", container)


docker/api/container.py
```diff
@@ -834,8 +834,13 @@ def logs(self, container, stdout=True, stderr=True, stream=False,
...
+                    elif isinstance(since, int):
+                        if since > 0:
+                            params['since'] = since
+                        else:
+                            raise errors.InvalidArgument(
+                                'since value should be greater than 0'
+                            )
                     else:
                         raise errors.InvalidArgument(
                             'since value should be datetime or int, not {}'.
```
shin-: Instead of adding more conditional branches, I'd rather update this message to say "datetime or positive int`. What do you think?

qazbnm456: Nice, I'll update it.

qazbnm456:
@shin- Changed it, thanks!

feliperuhland:
Hi, @qazbnm456. The build failed because the line is too long: ``docker/api/container.py:841:80: E501 line too long (85 > 79 characters)``.

feliperuhland:
Hi, @qazbnm456. I think you must sign the commit.

qazbnm456:
@feliperuhland Thank you. 😄

shin-:
Thanks!

PR #1834, opened 2017-12-12
Add support for order property when updating a service

### LINKED ISSUE #1823: Update config does not support order property
Update config for docker services does not support the "update-order" property, which is documented here: https://docs.docker.com/engine/reference/commandline/service_update/ .

mhank:
Resolves #1823 by adding support for order specification when updating a service.

Signed-off-by: Michael Hankin <mjhankin1@gmail.com>

### DIFF
--- a/docker/types/services.py
+++ b/docker/types/services.py
@@ -336,10 +336,9 @@ class UpdateConfig(dict):
           floating point number between 0 and 1. Default: 0
         order (string): Specifies the order of operations when rolling out an
           updated task. Either ``start_first`` or ``stop_first`` are accepted.
-          Default: ``stop_first``
     """
     def __init__(self, parallelism=0, delay=None, failure_action='continue',
-                 monitor=None, max_failure_ratio=None, order='stop-first'):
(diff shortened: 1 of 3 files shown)


docker/types/services.py
```diff
@@ -334,9 +334,12 @@ class UpdateConfig(dict):
         max_failure_ratio (float): The fraction of tasks that may fail during
           an update before the failure action is invoked, specified as a
           floating point number between 0 and 1. Default: 0
+        order (string): Specifies the order of operations when rolling out an
+          updated task. Either ``start_first`` or ``stop_first`` are accepted.
+          Default: ``stop_first``
     """
     def __init__(self, parallelism=0, delay=None, failure_action='continue',
-                 monitor=None, max_failure_ratio=None):
+                 monitor=None, max_failure_ratio=None, order='stop-first'):
```
shin-: This should be `None` by default, otherwise any client using API version < 1.29 will fail unless the user explicitly unsets it.

mhank:
@shin- fixed!

shin-:
Sweet, thank you!

Issue #1813, opened 2017-11-17
Using docker.client.DockerClient.containers.run with auto_remove=True causes a crash


ceridwen:
What happens is straightforward: setting auto_remove=True in the arguments to run() causes the container to be removed [before the Python client attempts to read the logs](https://github.com/docker/docker-py/blob/master/docker/models/containers.py#L734).  This then raises an error like:

```
docker.errors.NotFound: 404 Client Error: Not Found ("No such container: 0e9a9dd8072a1b80ead74bc47466493ffea92eca7acf1c5f3b29ce6faf861161")
```

This is not how executing `docker run -rm` works, where the output still goes to stdout, and it means that calling run() with auto_remove=True isn't useful.

thiagofigueiro:
+1
Minimal steps to reproduce using [example from manual](http://docker-py.readthedocs.io/en/stable/containers.html#docker.models.containers.ContainerCollection.run):

```
In [1]: import docker
In [2]: client = docker.from_env()
In [3]: client.containers.run('alpine:2.7', 'echo hello world')
Out[3]: b'hello world\n'
In [4]: client.containers.run('alpine:2.7', 'echo hello world', auto_remove=True)
---------------------------------------------------------------------------
HTTPError                                 Traceback (most recent call last)
(...)
~/.pyenv/versions/3.6.3/lib/python3.6/site-packages/docker/errors.py in create_api_error_from_http_exception(e)
     29         else:
     30             cls = NotFound
---> 31     raise cls(e, response=response, explanation=explanation)
     32
     33

NotFound: 404 Client Error: Not Found ("No such container: 6151f0219d2dbb90df7288f30f1221e817fd9d3c63afdbadf2ad2a87ba6e7e45")
```

StephenPCG:
This problem is not totally resolved by #1836 . I can still reproduce this issue with current master code.

```
In [1]: import docker
In [2]: client = docker.from_env()
In [3]: client.containers.run('alpine:latest', 'echo hello world', auto_remove=True)
...
~/.venvs/docker/lib/python3.6/site-packages/docker/errors.py in create_api_error_from_http_exception(e)
     29         else:
     30             cls = NotFound
---> 31     raise cls(e, response=response, explanation=explanation)
     32
     33

NotFound: 404 Client Error: Not Found ("No such container: 2e5677d940399606886ad6ce9303d458c27b6c251c33d624f0660640b9489547")
```

[...]

PR #1840, opened 2017-12-19
Don't attempt to retrieve container's stderr if `auto_remove` was set

shin-:
#1813

### DIFF
--- a/docker/models/containers.py
+++ b/docker/models/containers.py
@@ -737,7 +737,9 @@ class ContainerCollection(Collection):
 
         exit_status = container.wait()
         if exit_status != 0:
-            out = container.logs(stdout=False, stderr=True)
+            out = None
+            if not kwargs.get('auto_remove'):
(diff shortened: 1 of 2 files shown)


docker/models/containers.py
```diff
@@ -737,7 +737,10 @@ def run(self, image, command=None, stdout=True, stderr=False,
 
         exit_status = container.wait()
         if exit_status != 0:
-            out = container.logs(stdout=False, stderr=True)
+            if not kwargs.get('auto_remove'):
+                out = container.logs(stdout=False, stderr=True)
+            else:
+                out = None
```
feliperuhland: Hi @shin-, maybe put ``out = None`` before the ``if not kwargs.get('auto_remove')`` looks better. What do you think?

shin-: 👍 Good catch

david-drinn:
I ran into this issue on 3.0.0. Can this be merged into 3.0.x?

Issue #1734, opened 2017-08-30
Add support for service.scale(replicas)


lhenry7:
The ability to scale a service from the API in a simple way would be great. Mimic docker service scale SERVICE=REPLICAS

tobiasfielitz:
second this!

shin-:
@tobiasfielitz It's an `update` call setting `mode` parameter with the appropriate number of replicas, i.e

```python
mode = ServiceMode('replicated', replicas=5)
service.update(mode=mode, fetch_current_spec=True)
```

That said, it'd be reasonable to have a method for that since it's a fairly common operation.

feliperuhland:
Hi @shin-. Can I work on this? Thanks.

tobiasfielitz:
@shin- thanks for pointing me at this, that should do (until @feliperuhland or anyone else has written a function for it).
Thanks for the great work guys!

feliperuhland:
@tobiasfielitz @shin- I am already working on this!

PR #1846, opened 2017-12-22
Added scale method to the Service model.

feliperuhland:
This should satisfy the request #1734

### DIFF
--- a/docker/models/services.py
+++ b/docker/models/services.py
@@ -1,6 +1,6 @@
 import copy
 from docker.errors import create_unexpected_kwargs_error
-from docker.types import TaskTemplate, ContainerSpec
+from docker.types import TaskTemplate, ContainerSpec, ServiceMode
 from .resource import Model, Collection
 
 
@@ -104,6 +104,22 @@ class Service(Model):
             'TTY', False
         )
         return self.client.api.service_logs(self.id, is_tty=is_tty, **kwargs)
+
+    def scale(self, replicas):
+        """
+        Scale service container.
+
+        Args:
+            replicas (int): The number of containers that should be running.
+
+        Returns:
+            ``True``if successful.
+        """
+
+        service_mode = ServiceMode('replicated', replicas)
+        return self.client.api.update_service(self.id, self.version,
+                                              service_mode,
+                                              fetch_current_spec=True)
 
 
 class ServiceCollection(Collection):
--- a/docker/models/services.py
+++ b/docker/models/services.py
@@ -116,7 +116,7 @@ class Service(Model):
             ``True``if successful.
         """
 
-        if not self.attrs['Spec']['Mode'].get('Global'):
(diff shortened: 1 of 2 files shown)


feliperuhland:
@shin- You're right. I just submitted a validation for it. If you think that the exception message isn't good, please tell me. Thanks.

feliperuhland:
> Could not update commit status. Message: Server returned HTTP response code: 201, message: 'Created' for URL: https://api.github.com/repos/docker/docker-py/statuses/0e0a8526801d207bfdf2e9011ab1348525437a79

I think it's a network problem.

tests/integration/models_services_test.py
```diff
@@ -203,6 +205,50 @@ def test_scale_service(self):
         spec = service.attrs['Spec']['TaskTemplate']['ContainerSpec']
         assert spec.get('Command') == ['sleep', '300']
 
+    def test_scale_method_service(self):
```
shin-: This test is failing because the default `mode` for a service is `global`. Make sure to set the initial mode to `replicated`!

feliperuhland: Thanks @shin-. I already changed.

tests/integration/models_services_test.py
```diff
@@ -203,6 +205,50 @@ def test_scale_service(self):
...
+            tasks = service.tasks()
+        assert len(tasks) >= 2
+        # check that the container spec is not overridden with None
+        service.reload()
+        spec = service.attrs['Spec']['TaskTemplate']['ContainerSpec']
+        assert spec.get('Command') == ['sleep', '300']
+
+    def test_scale_method_global_service(self):
+        client = docker.from_env(version=TEST_API_VERSION)
+        mode = ServiceMode('global')
```
feliperuhland: It looks like the default value is ``replicated``.

docker/models/services.py
```diff
@@ -105,6 +105,25 @@ def logs(self, **kwargs):
...
+        Scale service container.
+
+        Args:
+            replicas (int): The number of containers that should be running.
+
+        Returns:
+            ``True``if successful.
+        """
+
+        if not self.attrs['Spec']['Mode'].get('Global'):
```
feliperuhland: The conditional is wrong.

feliperuhland: In fact, the conditional was wrong. It returned an empty dictionary and did not raise the exception.

lhenry7:
Hi, I find this new method does not work consistently. I got varying results, 1st time I called it with 1 task already running, service.scale(2) it worked like 'service update -force' and updated the running task and started a 2nd one.  But after that, service.scale(3) just didn't work at all, didn't start any new tasks or update existing ones.

I find 
    mode = docker.types.ServiceMode('replicated', replicas=running_tasks + 1)
    service.update(mode=mode, fetch_current_spec=True)
works like 'service scale', so I 'll continue doing that for now.

PR #1860, opened 2018-01-14
return current container status

InTheCloudDan:
When using the container status it was returning 'created' only. With this change the property should return whatever the current status of the container is.

Signed-off-by: Daniel OBrien <dobrien.nj@gmail.com>


shin-:
You should use the `reload()` method to obtain the updated status.

PR #1862, opened 2018-01-18
Change default TLS version

dperny:
Detects if python has an up-to-date version of OpenSSL that supports TLSv1.2. If it does, choose that as the default TLS version, instead of TLSv1. The Docker Engine and the majority of other Docker API servers should suppot TLSv1.2, and if they do not, the user can manually set a different (lower) version.

I haven't written Python in many, many moons, so I'm not sure if this is idiomatic or correct. To be honest, I'm not even sure how to run tests. 

Note that in the extremely unlikely chance that the user of this library is using an up-to-date version of OpenSSL pointed at a Docker API server using an out-of-date TLS library that doesn't support TLSv1.2, this is a breaking change, and it will require the user manually set a lower TLS version. However, as far as I can tell, every Docker product already supports TLSv1.2.

### DIFF
--- a/docker/tls.py
+++ b/docker/tls.py
@@ -37,13 +37,40 @@ class TLSConfig(object):
         self.assert_hostname = assert_hostname
         self.assert_fingerprint = assert_fingerprint
 
-        # TLS v1.0 seems to be the safest default; SSLv23 fails in mysterious
-        # ways: https://github.com/docker/docker-py/issues/963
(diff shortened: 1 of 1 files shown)


feliperuhland:
@dperny 5 test failed because ``AttributeError: 'module' object has no attribute 'PROTOCOL_TLSv1_2'``

PR #1868, opened 2018-01-26
"10-9" lost a "^" at some point

mccalluc:
"e" notation should be clear enough?


mccalluc:
Sorry: Let me try again and include the signature.

PR #1876, opened 2018-01-28
Minor tweaks in Container.exec_run

funkyfuture:
i think `None` is semantically more appropriate when `Container.exec_run` returns a generator or a socket. then i had the idea with the namedtuple for expressive code.

### DIFF
--- a/docker/models/containers.py
+++ b/docker/models/containers.py
@@ -152,7 +152,8 @@ class Container(Model):
         Returns:
             (tuple): A tuple of (exit_code, output)
                 exit_code: (int):
-                    Exit code for the executed command
+                    Exit code for the executed command or ``None`` if
+                    either ``stream```or ``socket`` is ``True``.
                 output: (generator or str):
                     If ``stream=True``, a generator yielding response chunks.
                     If ``socket=True``, a socket object for the connection.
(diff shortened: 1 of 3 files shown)


docker/models/containers.py
```diff
@@ -170,10 +171,12 @@ def exec_run(self, cmd, stdout=True, stderr=True, stdin=False, tty=False,
         exec_output = self.client.api.exec_start(
             resp['Id'], detach=detach, tty=tty, stream=stream, socket=socket
         )
-        exit_code = 0
-        if stream is False:
-            exit_code = self.client.api.exec_inspect(resp['Id'])['ExitCode']
-        return (exit_code, exec_output)
+        if socket or stream:
+            return ExecResult(None, exec_output)
+        else:
```
feliperuhland: Hi @funkyfuture. It looks like the ``else`` block is unnecessary because of the ``return``. You can use something like:

```
if socket or stream:
    return ExecResult(None, exec_output)

return ExecResult(
    self.client.api.exec_inspect(resp['Id'])['ExitCode'],
    exec_output)
```

Regards.

funkyfuture: i prefer the explicit over the implicit.

feliperuhland: I think it looks cleaner, but it's a personal matter. Your code, your decision. Greetings.

Issue #1742, opened 2017-09-13
How do I find container ports?


andyjp94:
I'm trying to create an index page for a test server that is autogenerated and just prints the name of the docker container and the port it is running on as a hyperlink. However I don't seem to be able to find a way to retrieve the port the containers are running on from your documentation. Could you give me an example  of how to do this please?

gtseres:
You 'll have to use the low-level API for that:

```python
import docker
client = docker.APIClient(base_url='unix://var/run/docker.sock')
port_data = client.inspect_container(<container_id>)['NetworkSettings']['Ports']
```

larkost-fetch:
With #1882 (so presumably available on the versions that will follow 3.7.2) there is now a second way of doing this:
```python
import docker
client = docker.from_env()
container = client.containers.run("alpine", "sleep 100", detach=True)
container.reload()  # required to get auto-assigned ports, not needed if it was an already running container
print(container.ports)
```

PR #1882, opened 2018-01-30
add ports to containers

larkost:
This adds a `ports` property to the `Container` object to list currently forwarded ports.

This should be the answer to #1742

### DIFF
--- a/tests/integration/models_containers_test.py
+++ b/tests/integration/models_containers_test.py
@@ -346,6 +346,28 @@ class ContainerTest(BaseIntegrationTest):
                     'memory_stats', 'blkio_stats']:
             assert key in stats
 
+    def test_ports(self):
+        client = docker.from_env(version=TEST_API_VERSION)
+        target_ports = {'2222/tcp': None}
+        container = client.containers.run(
+            "alpine", "sleep 100", detach=True,
+            ports=target_ports
+        )
+        self.tmp_containers.append(container.id)
+        container.reload()  # required to get auto-assigned ports
+        actual_ports = container.ports
+        assert sorted(target_ports.keys()) == sorted(actual_ports.keys())
+        for target_client, target_host in target_ports.items():
+            for actual_port in actual_ports[target_client]:
+                actual_keys = sorted(actual_port.keys())
+                assert sorted(['HostIp', 'HostPort']) == actual_keys
+                if target_host is None:
+                    int(actual_port['HostPort'])
+                elif isinstance(target_host, (list, tuple)):
+                    raise NotImplementedError()
+                else:
+                    assert actual_port['HostPort'] == target_host.split('/', 1)
+
     def test_stop(self):
         client = docker.from_env(version=TEST_API_VERSION)
         container = client.containers.run("alpine", "top", detach=True)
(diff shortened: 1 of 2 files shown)


larkost:
Just in case... this was signed 19 days ago.

tests/integration/models_containers_test.py
```diff
@@ -279,6 +279,28 @@ def test_stats(self):
...
+        self.tmp_containers.append(container.id)
+        container.reload()  # required to get auto-assigned ports
+        actual_ports = container.ports
+        assert sorted(target_ports.keys()) == sorted(actual_ports.keys())
+        for target_client, target_host in target_ports.items():
+            for actual_port in actual_ports[target_client]:
+                actual_keys = sorted(actual_port.keys())
+                assert sorted(['HostIp', 'HostPort']) == actual_keys
+                if target_host is None:
+                    int(actual_port['HostPort'])
```
TomasTomecek: shouldn't there be an assert here?

tests/integration/models_containers_test.py
```diff
@@ -279,6 +279,28 @@ def test_stats(self):
...
+                elif isinstance(target_host, (list, tuple)):
+                    raise NotImplementedError()
```
TomasTomecek: `TypeError` or `ValueError` would be better

tests/integration/models_containers_test.py
larkost: No, the `int()` will already throw an error with a reasonably good message if it is not an integer, and that will show up nicely as a failure. And the actual port returned (on the host) is going to be effectively random in this case.

tests/integration/models_containers_test.py
larkost: I don't think so. This is just me being forward-looking. If someone changed the test in the future to have multiple ports, then it would go down this branch. But for the moment it should just report that this has not been implemented (hopefully prompting that person to expand it).

larkost:
@TomasTomecek Could you please respond to my responses? This PR is getting a bit stale.

tests/integration/models_containers_test.py
TomasTomecek: [NotImplementedError](https://docs.python.org/3/library/exceptions.html#NotImplementedError)

    In user defined base classes, abstract methods should raise this exception when they require 
    derived classes to override the method, or while the class is being developed to indicate that 
    the real implementation still needs to be added.

tests/integration/models_containers_test.py
TomasTomecek: makes sense

TomasTomecek:
I'm not a maintainer of this project, my comments are pretty much just my opinions. For a proper code review you need to wait for @shin-

backtrackbaba:
Hey @shin- any update on this PR?

tests/integration/models_containers_test.py
ulyssessouza: Since you are in a test and you have the input and the expected output, IMHO looks more reasonable to assert your expected path like:
```python
assert target_host is None
assert int(actual_port['HostPort']) > 0
```
For the other types, you can can add other tests that does essentially the same but with different structures, like `list` and `tuple`.

yxliang01:
@larkost  @ulyssessouza  Any progress on this PR? It has been a year and this is super sensible and useful. Hope this can be merged.

yxliang01:
Great! Happy to see this is finally merged!

PR #1898, opened 2018-02-02
Add support for generic resource reservations

### LINKED ISSUE #1877: Nvidia support?
I know that a [PR](#1560) was being worked on to provide nvidia support, but apparently it is no longer necessary with [`nvidia-docker2`](https://github.com/NVIDIA/nvidia-docker), but I cannot find anything in the documentation.

Is it currently possible (provided [`nvidia-docker2`](https://github.com/NVIDIA/nvidia-docker) is properly installed/configured) to do the equivalent of:

It looks like this is available via the [`runtime`](https://docker-py.readthedocs.io/en/stable/containers.html#docker.models.containers.ContainerCollection.run) parameter?
```
docker run --runtime=nvidia nvidia/cuda nvidia-smi
```

and as of [`v17.12.0-ce`](https://github.com/NVIDIA/nvidia-docker/issues/141):
```
docker service create -t --generic-resource "gpu=1" nvidia/cuda nvidia-smi
```

It looks like the [`resources`](https://docker-py.readthedocs.io/en/stable/api.html#docker.types.Resources) API does not support `--generic-resource`...

mwilliammyers:
Fix #1877

### DIFF
--- a/docker/types/services.py
+++ b/docker/types/services.py
@@ -306,9 +306,14 @@ class Resources(dict):
         mem_limit (int): Memory limit in Bytes.
         cpu_reservation (int): CPU reservation in units of 10^9 CPU shares.
         mem_reservation (int): Memory reservation in Bytes.
+        generic_resources (dict:`list`): List of node level generic
+          resources, for example a GPU, in the form of
+          ``{ ResourceSpec: { 'Kind': kind, 'Value': value }}``, where
+          ResourceSpec is one of 'DiscreteResourceSpec' or 'NamedResourceSpec'
+          or in the form of ``{ resource_name: resource_value }``.
     """
     def __init__(self, cpu_limit=None, mem_limit=None, cpu_reservation=None,
-                 mem_reservation=None):
+                 mem_reservation=None, generic_resources=None):
         limits = {}
         reservation = {}
         if cpu_limit is not None:
@@ -319,6 +324,33 @@ class Resources(dict):
             reservation['NanoCPUs'] = cpu_reservation
         if mem_reservation is not None:
             reservation['MemoryBytes'] = mem_reservation
+        if generic_resources is not None:
+            # if isinstance(generic_resources, list):
+            #     reservation['GenericResources'] = generic_resources
+            if isinstance(generic_resources, (list, tuple)):
+                reservation['GenericResources'] = list(generic_resources)
+            elif isinstance(generic_resources, dict):
+                resources = []
+                for kind, value in six.iteritems(generic_resources):
+                    resource_type = None
+                    if isinstance(value, int):
+                        resource_type = 'DiscreteResourceSpec'
+                    elif isinstance(value, str):
+                        resource_type = 'NamedResourceSpec'
+                    else:
+                        raise errors.InvalidArgument(
+                            'Unsupported generic resource reservation '
+                            'type: {}'.format({kind: value})
+                        )
+                    resources.append({
+                        resource_type: {'Kind': kind, 'Value': value}
+                    })
+                reservation['GenericResources'] = resources
+            else:
+                raise errors.InvalidArgument(
+                    'Unsupported generic resources '
+                    'type: {}'.format(generic_resources)
+                )
 
         if limits:
             self['Limits'] = limits
--- a/tests/integration/api_service_test.py
+++ b/tests/integration/api_service_test.py
@@ -212,6 +212,57 @@ class ServiceTest(BaseAPIIntegrationTest):
             'Reservations'
         ]
 
+    def _create_service_with_generic_resources(self, generic_resources):
+        container_spec = docker.types.ContainerSpec(BUSYBOX, ['true'])
+
+        resources = docker.types.Resources(
+            generic_resources=generic_resources
+        )
+        task_tmpl = docker.types.TaskTemplate(
+            container_spec, resources=resources
+        )
+        name = self.get_service_name()
+        svc_id = self.client.create_service(task_tmpl, name=name)
+        return resources, self.client.inspect_service(svc_id)
+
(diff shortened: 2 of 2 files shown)


docker/types/services.py
```diff
@@ -306,9 +306,12 @@ class Resources(dict):
         mem_limit (int): Memory limit in Bytes.
         cpu_reservation (int): CPU reservation in units of 10^9 CPU shares.
         mem_reservation (int): Memory reservation in Bytes.
+        generic_reservations (dict): List of node level generic
```
shin-: I understand you were looking to follow the same pattern as with `cpu_` and `mem_`, but I don't see a scenario where "generic limits" becomes a thing (especially because the Docker CLI already uses `--generic-resource` for this), so I think we should rename this to `generic_resources`

docker/types/services.py
```diff
@@ -319,6 +322,26 @@ def __init__(self, cpu_limit=None, mem_limit=None, cpu_reservation=None,
             reservation['NanoCPUs'] = cpu_reservation
         if mem_reservation is not None:
             reservation['MemoryBytes'] = mem_reservation
+        if generic_reservations is not None:
+            generic_resources = []
```
shin-: I'd like to allow users to provide this argument in list form (i.e. already parsed) as well ; we just need to have a `isinstance(generic_reservations, list)` that bypasses parsing.

This is useful for people grabbing the resources data by inspecting another service, for example.

mwilliammyers:
Yeah, should I add the integration tests in: tests/integration/models_resources_test.py and/or `tests/integration/api_service_test.py`?

docker/types/services.py
mwilliammyers: yeah, for sure.

docker/types/services.py
mwilliammyers: So just allow:

```
[  {'DiscreteResourceSpec': {'Kind': 'gpu', 'Value': 1}}, ...]
```

I was wondering if we wanted to expose that API to users...

because I am building a CLI app that mimics docker so they are coming in like:
```
gpu=1
```
but I don't think we should use that at the API level?

shin-: Yeah, CLI-like notation is out of scope for this project ; exposing the original API format when possible is something we do because it allows people to reuse `inspect` results and gives them a workaround when new things make it into the API until we can implement them at the SDK level.

shin-:
Tests for this should be in `tests/integration/api_service_test.py` :+1:

docker/types/services.py
mwilliammyers: ok! Ill change it to match the docker API

mwilliammyers:
@shin- implemented your suggestions. Let me know if you want me to change anything else or if you would like me to squash my commits.

tests/integration/api_service_test.py
```diff
@@ -196,7 +196,11 @@ def test_create_service_with_resources_constraints(self):
         container_spec = docker.types.ContainerSpec(BUSYBOX, ['true'])
         resources = docker.types.Resources(
             cpu_limit=4000000, mem_limit=3 * 1024 * 1024 * 1024,
-            cpu_reservation=3500000, mem_reservation=2 * 1024 * 1024 * 1024
+            cpu_reservation=3500000, mem_reservation=2 * 1024 * 1024 * 1024,
+            generic_resources=[
+                {'DiscreteResourceSpec': {'Kind': 'gpu', 'Value': 1}},
+                {'NamedResourceSpec': {'Kind': 'gpu', 'Value': 'a'}}
```
shin-: We should check that we're able to retrieve this data in `res_template` below (data sent by the server)

shin-: Also it's fine to make a separate test for this.

docker/types/services.py
```diff
@@ -319,6 +323,13 @@ def __init__(self, cpu_limit=None, mem_limit=None, cpu_reservation=None,
             reservation['NanoCPUs'] = cpu_reservation
         if mem_reservation is not None:
             reservation['MemoryBytes'] = mem_reservation
+        if generic_resources is not None:
+            if not isinstance(generic_resources, list):
```
shin-: Oops! I wasn't super clear in what I meant, sorry!
I actually think we can have both notations, the `dict` form that you implemented originally with the list form as a fallback. See for example https://github.com/docker/docker-py/blob/master/docker/types/containers.py#L242-L246

tests/integration/api_service_test.py
mwilliammyers: Is [L215](https://github.com/BYU-PCCL/docker-py/blob/0629f7bd6c3d96460e42071d54ed9cd7ef1f7d69/tests/integration/api_service_test.py#L215) not doing that? As a sanity check I printed out `res_template['Resources']['Reservations']` and it contained the generic resources...

Yeah I wasn't sure if we wanted an separate test function...

mwilliammyers:
@shin- Done. I improved my integration test, but couldn't get the [`@pytest.mark.parametrize`](https://docs.pytest.org/en/latest/parametrize.html) decorator to work from within the `ServiceTest(BaseAPIIntegrationTest)` class so I had to do a workaround to support testing `dict`, `list` and bad inputs.

shin-:
Thank you!

Rebased and merged manually in 820de848fa73f20cb80215ceb4b8cdafc855867e

mwilliammyers:
Thanks! Sorry I didn't break the parsing out into a function and thanks for adding the API tag, I couldn't figure out exactly what version that was added in...

shin-:
No worries!

> I couldn't figure out exactly what version that was added in...

Yep, and I apparently got it wrong too :confounded:  It's not your fault, it's missing from the changelog.

PR #1906, opened 2018-02-13
Add chunk_size parameter to data downloading methods

### LINKED ISSUE #1352: Stream Docker image tarball to disk instead of memory dirst
I'm attempting to create a tarball of an existing Docker image via the https://docker-py.readthedocs.io/en/stable/images.html save example. When attempting to write the data to disk, I get `MemoryError` during the writing process.

The image is 1.516GB so I'm assuming we are running out of memory (this process is within a container).

Can the save be done directly to disk and avoid being pushed into memory?

shin-:
Fixes #1352 

cc @david-drinn @realityone @goldmann Feedback appreciated!


realityone:
Thanks for updating! (´ω｀)

All codes LGTM. 

The only thing i'm concerned about is the comment at line https://github.com/docker/docker-py/pull/1906/files#diff-b5bb91e2b94da86ea05c1fe77a6ff432R354 should be updated now.

We not only use this method to attach to a TTY but also used in more case now.

shin-:
@pvizeli It's not set in stone, but most likely 3.1.0 should come out late February or early March

TomasTomecek:
Would it make sense to cut 3.0.2 soon-ish?

pvizeli:
I think also that need go into a hotfix. Actual it make every image/file handling unusable

shin-:
3.1.0 is out with the fix.

PR #1909, opened 2018-02-14
Clean up created volume from test_run_with_named_volume

yongtang:
This fix adds the volume id to the list so that it could be
cleaned up on test teardown.

The issue was originally from https://github.com/moby/moby/pull/36292
where an additional `somevolume` pre-exists in tests.

Signed-off-by: Yong Tang <yong.tang.github@outlook.com>


shin-:
Thanks!

PR #1914, opened 2018-02-19
Improve .dockerignore compliance

mefyl:
Initially attempting to fix https://github.com/docker/compose/issues/3886, this PR revamps .dockerignore files to comply with the documentation on several aspects:

* AFAICT absolute paths where not handled correctly, although there was an attempt - confirmed manually. The issue is that leading slashes where handled before exceptions (patterns starting with "!").
* Wildcards / globbing where not handled in exceptions.
* The last matching line now takes precedence, instead of the previous "inclusion over exclusion" rule.

The codebase was also made greatly smaller / simpler.

Unit test for "should_check_directory", an internal helper, were dropped since the implementation changed. All other unit test are honored as-is, except one about absolute path whose expectation where, I think, wrong - path are never rooted in the host filesystem at this point.

### DIFF
--- a/docker/utils/build.py
+++ b/docker/utils/build.py
@@ -91,6 +91,10 @@ def walk(root, patterns, default=True):
         matched = default if hit is None else hit
         sub = list(filter(lambda p: p[1], sub))
         if os.path.isdir(cur):
+            # Entirely skip directories if there are no chance any subfile will
+            # be included.
+            if all(not p[0] for p in sub) and not matched:
+                continue
             children = False
             for r in (os.path.join(f, p) for p in walk(cur, sub, matched)):
                 yield r
--- a/docker/utils/build.py
+++ b/docker/utils/build.py
@@ -46,7 +46,7 @@ def exclude_paths(root, patterns, dockerfile=None):
     )
 
 
-def should_include(path, exclude_patterns, include_patterns, root):
+def should_include(path, exclude_patterns, include_patterns):
     """
     Given a path, a list of exclude patterns, and a list of inclusion patterns:
 
@@ -61,15 +61,11 @@ def should_include(path, exclude_patterns, include_patterns, root):
             for pattern in include_patterns:
                 if match_path(path, pattern):
                     return True
-                if os.path.isabs(pattern) and match_path(
-                        os.path.join(root, path), pattern):
-                    return True
             return False
     return True
 
 
-def should_check_directory(directory_path, exclude_patterns, include_patterns,
-                           root):
+def should_check_directory(directory_path, exclude_patterns, include_patterns):
     """
     Given a directory path, a list of exclude patterns, and a list of inclusion
     patterns:
@@ -95,7 +91,7 @@ def should_check_directory(directory_path, exclude_patterns, include_patterns,
         if (pattern + '/').startswith(path_with_slash)
     ]
     directory_included = should_include(
-        directory_path, exclude_patterns, include_patterns, root
+        directory_path, exclude_patterns, include_patterns
     )
     return directory_included or len(possible_child_patterns) > 0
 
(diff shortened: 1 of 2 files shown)


docker/utils/build.py
```diff
@@ -23,127 +27,90 @@ def exclude_paths(root, patterns, dockerfile=None):
...
+        splitted = [pt for pt in re.split(_SEP, p) if pt and pt != '.']
+        # ".." component must be cleared with the potential previous
+        # component, regardless of whether it exists: "A preprocessing
+        # step [...]  eliminates . and .. elements using Go's
+        # filepath.".
+        i = 0
+        while i < len(splitted):
+            if splitted[i] == '..':
+                del splitted[i]
+                if i > 0:
```
shin-: Does this mean that `../xxx` becomes `xxx`? If so, does it match the behavior of the engine?

shin-: More generally, having some tests for `..` components would be nice

docker/utils/build.py
```diff
@@ -23,127 +27,90 @@ def exclude_paths(root, patterns, dockerfile=None):
...
-
-def should_check_directory(directory_path, exclude_patterns, include_patterns,
-                           root):
+    def normalize(p):
+        # Leading and trailing slashes are not relevant. Yes,
+        # "foo.py/" must exclude the "foo.py" regular file. "."
+        # components are not relevant either, even if the whole
+        # pattern is only ".", as the Docker reference states: "For
+        # historical reasons, the pattern . is ignored."
+        splitted = [pt for pt in re.split(_SEP, p) if pt and pt != '.']
```
shin-: nit: `split` (or better yet, `split_path`)

docker/utils/build.py
```diff
@@ -23,127 +27,90 @@ def exclude_paths(root, patterns, dockerfile=None):
...
+            else:
+                i += 1
+        return splitted
+
+    patterns = (
+        (True, normalize(p[1:]))
+        if p.startswith('!') else
+        (False, normalize(p))
+        for p in patterns)
+    patterns = list(reversed(list(chain(
```
shin-: does `patterns` need to be a list? It looks like `walk` only ever uses it as an iterator.

docker/utils/build.py
```diff
@@ -23,127 +27,90 @@ def exclude_paths(root, patterns, dockerfile=None):
...
-    if len(pattern_components) == 1 and IS_WINDOWS_PLATFORM:
-        pattern_components = pattern.split('/')
-
-    if '**' not in pattern:
-        path_components = path.split(os.path.sep)[:len(pattern_components)]
-    else:
-        path_components = path.split(os.path.sep)
-    return fnmatch('/'.join(path_components), '/'.join(pattern_components))
+    def match(p):
+        if p[1][0] == '**':
```
shin-: We're using a modified `fnmatch` implementation that _should_ handle `**` patterns - did you find that not to be the case? If so, maybe we can get rid of the `utils/fnmatch.py` file as well and use Python's `fnmatch` module instead.

docker/utils/build.py
mefyl: The official documentation about this is rather elusive. It does however stipulate that absolute paths are equivalent to relative paths, hence `/../foo` should be equivalent to `../foo`. It also stipulates that paths are run through Go's `filepath.Clean`, which explicitely makes this equivalence: ```Eliminate .. elements that begin a rooted path:
   that is, replace "/.." by "/" at the beginning of a path
```

From this, I would argue that yes, the specification prompts for ../foo to be equivalent to foo.

But as you smartly suggested, I tested with Docker `17.09.1-ce`, and `../foo` does not ignore a `foo` file at the root - I suppose it won't ignore anything ever.

However, brace yourself, if you use `/../foo`, it *does* ignore a `foo` file at the root, which violates the statements that absolute and relative paths are equivalent. Long story short: it's a mess.

It might be more reasonable to mimic the current engine behavior, although not following the spec could seem worse. The spec would probably need some updating if we choose to mimic the engine.

docker/utils/build.py
mefyl: Nah, there's a `for` loop in `walk` reusing the patterns. As for the intermediate `list` inside `reverse`, sadly python2's `reverse` expects a plain collection -_-

docker/utils/build.py
mefyl: The previous implementation did not handle `**` in exceptions, and it didn't look easy with this approach while preserving the optimization to not explore everything recursively when we're sure a directory is entirely excluded. Also, the current implementation looks much shorter and simpler to me, leveraging the stdandard fnmatch - provided I actually use the standard fnmatch, granted ;)

shin-: Cool, I'm fine with the new implementation if it works better than the old one :)
Can we remove utils/fnmatch.py from the tree in this case?

docker/utils/build.py
shin-: Ok - it probably doesn't see a lot of use anyway. I say we keep it as is and we can always revisit it later if it's an issue.

mefyl: I agree that even any `..` in a `.dockerignore` file doesn't make much sense to me. Especially since `anything/..` will be dropped, even if `anything` does not exist as a directory, so I really don't see the point. IMO we should even deprecate and then reject them.

mefyl:
I don't really like the idea of squashing: I don't see what it brings to the table as you can always review the diffs for the whole merge, but I think it removes a lot. As in, readibility, and for instance I added two optional "optimization" commits that can easily be blamed back, and reverted individually while conserving the semantics if those optimisations cause problem.

Now, I don't mean to come in the project on my high horses and impose my ways, if you prefer I'll squash for now, but that's a discussion I'd like to have at some point.

shin-:
Yeah, I think it's fine, these commits are all individually clean and self-contained. 👍

DxCx:
Not intentionally :)
Ill try importing both packages and see if the functions around here returns different values.

Thanks that was very helpful, hope to provide with more info soon :)

Issue #1358, opened 2016-12-16
Ability to stop docker-py.events() generator from another thread?


approxit:
When we want to listen for live docker events with docker-py we have 2 options:

- Fetch events with time limits like `since` and `until`. If `until` is the future date, code is blocking until we meet that date (but not pass over). Minimal time resolution for `since` and `until` is one second.

- Fetch events live without time limits. Generator is blocking until some event pops out.

In projects based on for eg. Python Twisted, we have single event loop thread. We can't run blocking queries, because we don't want to block whole application. That's why we execute blocking code in threads. Each 2 options I mentioned above we have problems.

- Fetching events without time limits can't be stopped. And when whole application is closing, thread is still running, so Twisted waits forever for ~thread.join. We have 0 control over generator, because it uses internal waits for some events. We can't use a variable to tell thread that is should be still running. We are totally stuck.

- We are forced to use fetching with time limits. We want to get events live, so we make smallest possible time chunks. Even smallest chunks gives us up to 1 second delay, so it's not live. I don't even mention about logic to manage that chunked fetching.

Two scenarios are bad.

Is there something that I'm not aware of, or we really can''t do much more?

TomasTomecek:
I'm using `threading.Event` for that: https://github.com/TomasTomecek/sen/blob/24dcbb63ca1dd9af2247f0cd25074dc41b27fc52/sen/tui/views/main.py#L63 https://github.com/TomasTomecek/sen/blob/24dcbb63ca1dd9af2247f0cd25074dc41b27fc52/sen/tui/views/main.py#L44

approxit:
@TomasTomecek Thanks for response.

Yes, threading.Event WOULD do the job, but [browsing your code](https://github.com/TomasTomecek/sen/blob/fc1ac27e8839cced643d7081eba362a06cd7cb75/sen/docker_backend.py#L785) I think you're getting to point I wrote about:
> Fetch events live without time limits. Generator is blocking until some event pops out.

So yes, you are using theading.Event, but it will stop that thread [as soon as new event arrives](https://github.com/TomasTomecek/sen/blob/fc1ac27e8839cced643d7081eba362a06cd7cb75/sen/tui/init.py#L45) and you will [return to your code that can check threading.Event](https://github.com/TomasTomecek/sen/blob/fc1ac27e8839cced643d7081eba362a06cd7cb75/sen/tui/init.py#L55). Not earlier.

[...]

PR #1917, opened 2018-02-20
Allow cancelling the events stream from other threads

rycus86:
Hi,

These changes would allow wrapping the original `client.events()` response into an object, that allows you to close it from another thread. This also means you need to call another method on the original thread to start consuming events from the generator.

Example:
```python
client = docker.from_env()
events = client.events(cancellable=True)

for event in events.stream():  # we're blocked here
  print event

print 'After reading the event stream'  # once closed

# from another thread:
events.close()
```

Not sure if this is the best design choice, so please tell me any feedback. The change introduces a new `CancellableEvents` type for the wrapper, and modifies the `DaemonApiMixin.events(..)` method to accept a new `cancellable` bool flag. When false, the generator is returned as before, if true, the wrapped object.

This PR would be a potential fix for https://github.com/docker/docker-py/issues/1358 and the implementation idea was based on https://stackoverflow.com/a/36511714 (mentioned on the issue).

Thank you!

### DIFF
--- a/docker/api/daemon.py
+++ b/docker/api/daemon.py
@@ -1,7 +1,7 @@
 import os
 from datetime import datetime
 
-from .. import auth, utils
+from .. import auth, types, utils
 
 
 class DaemonApiMixin(object):
@@ -34,8 +34,7 @@ class DaemonApiMixin(object):
                 the fly. False by default.
 
         Returns:
-            (generator): A blocking generator you can iterate over to retrieve
-                events as they happen.
+            A :py:class:`docker.types.daemon.CancellableStream` generator
 
         Raises:
             :py:class:`docker.errors.APIError`
@@ -50,6 +49,14 @@ class DaemonApiMixin(object):
              u'status': u'start',
              u'time': 1423339459}
             ...
+
+            or
+
+            >>> events = client.events()
+            >>> for event in events:
+            ...   print event
+            >>> # and cancel from another thread
+            >>> events.close()
         """
 
         if isinstance(since, datetime):
@@ -68,10 +75,10 @@ class DaemonApiMixin(object):
         }
         url = self._url('/events')
 
-        return self._stream_helper(
-            self._get(url, params=params, stream=True, timeout=None),
-            decode=decode
-        )
+        response = self._get(url, params=params, stream=True, timeout=None)
+        stream = self._stream_helper(response, decode=decode)
+
+        return types.CancellableStream(stream, response)
 
     def info(self):
         """
--- a/docker/types/daemon.py
+++ b/docker/types/daemon.py
@@ -1,0 +1,63 @@
+import socket
+
+try:
+    import requests.packages.urllib3 as urllib3
+except ImportError:
+    import urllib3
+
+
+class CancellableStream(object):
+    """
+    Stream wrapper for real-time events, logs, etc. from the server.
+
+    Example:
+        >>> events = client.events()
+        >>> for event in events:
+        ...   print event
+        >>> # and cancel from another thread
+        >>> events.close()
+    """
+
+    def __init__(self, stream, response):
+        self._stream = stream
+        self._response = response
+
+    def __iter__(self):
+        return self
+
+    def __next__(self):
+        try:
+            return next(self._stream)
+        except urllib3.exceptions.ProtocolError:
+            raise StopIteration
+        except socket.error:
+            raise StopIteration
+
+    next = __next__
+
+    def close(self):
+        """
+        Closes the event streaming.
+        """
+
+        if not self._response.raw.closed:
+            # find the underlying socket object
+            # based on api.client._get_raw_response_socket
+
+            sock_fp = self._response.raw._fp.fp
+
+            if hasattr(sock_fp, 'raw'):
+                sock_raw = sock_fp.raw
+
+                if hasattr(sock_raw, 'sock'):
+                    sock = sock_raw.sock
+
+                elif hasattr(sock_raw, '_sock'):
+                    sock = sock_raw._sock
+
+            else:
+                sock = sock_fp._sock
+
+            sock.shutdown(socket.SHUT_RDWR)
+            sock.makefile().close()
+            sock.close()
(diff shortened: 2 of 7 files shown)


shin-:
I think the idea is good!

Just a short observation before I review this: could `CancellableEvents` implement the iterator interface with `__next__` and `__iter__` methods? Then we wouldn't need the `cancellable=True` parameter since the change would be backward-compatible.

Also, do we want to add this to other methods like `logs` and `attach` as well?

rycus86:
Thanks a lot for the suggestions!
Sure, I can add the iterator methods to the new class, we'll still need to keep a reference of the returned object though, so it would be something like:

```python
events = client.events(cancellable=True)

for event in events:  # we're blocked here
  print event

# then cancel with:
events.close()
```

Also, for `logs` and `attach` and others, it makes complete sense, I could make this a generic cancellable iterator so we could plug it in more places.

rycus86:
I'm think I've got some unrelated test failures on Jenkins. :(

```
22:39:32 [py3.6_17.06.2-ce] tests/integration/api_container_test.py::AttachContainerTest::test_attach_no_stream FAILED
```

docker/types/daemon.py
```diff
@@ -0,0 +1,72 @@
...
+            pass
+
+    def __iter__(self):
+        return self.stream()
+
+    def close(self):
+        """
+        Closes the event streaming.
+        """
+
```
shin-: Can't we simply call `self._response.close()`? http://docs.python-requests.org/en/master/api/#requests.Response.close

docker/api/daemon.py
```diff
@@ -67,11 +84,13 @@ def events(self, since=None, until=None, filters=None, decode=None):
...
         }
         url = self._url('/events')
+        response = self._get(url, params=params, stream=True, timeout=None)
+        stream = self._stream_helper(response, decode=decode)
 
-        return self._stream_helper(
-            self._get(url, params=params, stream=True, timeout=None),
-            decode=decode
-        )
+        if cancellable:
```
shin-: I think we can get rid of this, and simply always return a `CancellableStream`, provided it behaves the same as the regular stream.

tests/integration/client_test.py
```diff
@@ -27,3 +29,25 @@ def test_df(self):
...
+
+        cancel_thread = threading.Thread(target=cancel_events)
+        cancel_thread.start()
+
+        for _ in events.stream():
+            pass
+
+        events.close()
+
+        # pass, otherwise we wouldn't get here
```
shin-: maybe we can have a timeout of some kind, so the test doesn't stall forever if we accidentally break something in the future?

docker/types/daemon.py
```diff
@@ -0,0 +1,72 @@
...
+                happen when the stream is closed.
+        """
+
+        try:
+            for item in self._stream:
+                yield item
+        except urllib3.exceptions.ProtocolError:
+            pass
+
+    def __iter__(self):
```
shin-: Can we add `__next__` as well?

docker/types/daemon.py
rycus86: No, unfortunately, it doesn't interrupt the blocking socket read.
From what I could find out, it looks like we need the `socket.shutdown(SHUT_RDWR)` call to do that fo us.

rycus86:
OK, I think I've got all the changes you requested (with the exception of `close`).
If all looks good, would you want me to add the `CancellableStream` in the other places you mentioned as part of this PR?

shin-:
@rycus86 Yes, can we add this to `attach` and `logs` as well? I think those are where it makes the most sense as potentially "infinite" streams.

On top of that, can you squash these commits into one? Thanks!

docker/types/daemon.py
```diff
@@ -0,0 +1,63 @@
...
+                    sock = sock_raw.sock
+
+                elif hasattr(sock_raw, '_sock'):
+                    sock = sock_raw._sock
+
+            else:
+                sock = sock_fp._sock
+
+            sock.shutdown(socket.SHUT_RDWR)
+            sock.makefile().close()
```
rycus86: This seems to help in Python 2.7 somewhat, but for `container.attach(..)` the test is still a bit flaky...
It fails around once in every 10-20 runs, sometimes it's more often though. :/

tests/integration/api_container_test.py
```diff
@@ -880,6 +881,30 @@ def test_logs_streaming_and_follow(self):
...
+            BUSYBOX, 'sh -c "echo \\"{0}\\" && sleep 3"'.format(snippet)
+        )
+        id = container['Id']
+        self.tmp_containers.append(id)
+        self.client.start(id)
+        logs = six.binary_type()
+
+        generator = self.client.logs(id, stream=True, follow=True)
+
+        exit_timer = threading.Timer(3, os._exit, args=[1])
```
rycus86: To avoid this, we could bring in the `pytest-timeout` module, if that sounds better?

tests/integration/api_container_test.py
```diff
@@ -1226,6 +1251,29 @@ def test_attach_no_stream(self):
         output = self.client.attach(container, stream=False, logs=True)
         assert output == 'hello\n'.encode(encoding='ascii')
 
+    def test_attach_stream_and_cancel(self):
```
rycus86: This test can potentially cause some problems, seems flaky on Python 2.7 (fine on 3.5).

rycus86:
Hi @shin- 

Let me know what you think of adding the `pytest-timeout` module, I think the tests have failed because one of them had to `os._exit(1)` on timeout.
Haven't really worked out how to make shutting the socket down more reliable on 2.7, though seem 10 out of 10 on 3.x

Thanks!

shin-:
I still need to take a closer look at this. Sorry it hasn't happened yet and thank you for being patient :cat:

rycus86:
No worries! :)
Thanks for coming back to it!

shin-:
Merged in #1965 

Thanks again!

PR #1918, opened 2018-02-20
Shorthand method for service.restart()

rycus86:
This PR would add a new shorthand method to restart a service, similarly to the recently added `scale` method.

Usage:
```python
service = client.services.create(..)
service.restart()
```

Thanks!

### DIFF
--- a/docker/models/services.py
+++ b/docker/models/services.py
@@ -69,6 +69,11 @@ class Service(Model):
             spec = self.attrs['Spec']['TaskTemplate']['ContainerSpec']
             kwargs['image'] = spec['Image']
 
+        if kwargs.get('force_update') is True:
+            task_template = self.attrs['Spec']['TaskTemplate']
+            current_value = int(task_template.get('ForceUpdate', 0))
+            kwargs['force_update'] = current_value + 1
+
         create_kwargs = _get_create_service_kwargs('update', kwargs)
 
         return self.client.api.update_service(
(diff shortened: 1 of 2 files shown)


rycus86:
I think I've got some unrelated test failures on Jenkins:

```
22:45:48 [py3.6_17.12.0-ce] tests/integration/api_swarm_test.py::SwarmTest::test_init_swarm_with_ca_config FAILED
```

shin-:
Yeah, sadly that test is hit or miss on 17.06 - don't worry about it.

docker/models/services.py
```diff
@@ -124,6 +124,23 @@ def scale(self, replicas):
                                               service_mode,
                                               fetch_current_spec=True)
 
+    def restart(self):
```
shin-: I think `restart` doesn't accurately represent the operation that's being performed here. In the Docker ecosystem, `restart` usually means "take the existing containers with their configuration and stop them then start them again". The same idea exists in other common software like nginx with the `reload` vs `restart` distinction. 

To my understanding. in this case, the containers for the service will actually be recreated with the same configuration - maybe `force_update` or `reload` is a more accurate name?

rycus86:
Thanks again, @shin- !
I have renamed the method to `force_update` and changed the docstring to be the same (or very similar to) what the engine uses as help strings.

I've also snuck in another change: the `force_update` parameter on the service model's `update` method now accepts a `bool`, and if it is `True` then it does the counter increment logic. I thought it makes more sense to have it there, so you can just say `service.update(force_update=True, **other_kwargs)` and it's more inline with the `docker service update --force` flag being a boolean flag.

It would be great if you could give the change another look because of these extras. Thank you!

shin-:
If you could just squash those commits into one, I can go ahead and merge it.

rycus86:
Sure, it's done now, thanks a lot!

PR #1922, opened 2018-02-23
Cleanup containers during the tests

yongtang:
This fix tries to clean up the containers during the tests
so that no pre-existing volumes left in docker integration tests.

This fix adds `-v` when removing containers, and makes sure
containers launched in non-daemon mode are removed.

This fix is realted to https://github.com/moby/moby/pull/36292

Signed-off-by: Yong Tang <yong.tang.github@outlook.com>


thaJeztah:
ping @shin- 🤗

PR #1925, opened 2018-02-24
Add prune_build

The-Loeki:
Adds https://docs.docker.com/engine/api/v1.36/#operation/BuildPrune

### DIFF
--- a/docker/api/build.py
+++ b/docker/api/build.py
@@ -292,3 +292,19 @@ class BuildApiMixin(object):
             )
         else:
             log.debug('No auth config found')
+
+    @utils.minimum_version('1.31')
+    def prune_build(self):
+        """
+        Delete builder cache
+
+        Returns:
+            (dict): A dict containing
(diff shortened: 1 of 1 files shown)


The-Loeki:
there you go... Although I do have to say I fail to see reasoning behind the hard demand for it; after all the only reason the key is trustworthy is because I said so in GH, so if my account was compromised so would the trust be.

The-Loeki:
@GordonTheTurtle I've added the GPG sig as per your request, but your dco-signed is apparently stuck in 'Waiting for status to be reported' or something?

shin-:
Gordon is a bot, so it won't be able to answer ;
The thing it's requesting is for you to add the signoff line to your commit message (`git commit -s`). Using the commands it suggests should fix everything. GPG signing is a different thing entirely and isn't required.

The-Loeki:
Yeah, but didn't I do just that?

The-Loeki:
Guess I should've read a bit better ;) sorry about that

PR #1937, opened 2018-03-01
Add support for build time volumes

### LINKED ISSUE #1498: build method is missing a volumes argument
Docker can mount read-only volumes at image build time (see docker build --volume) but this is not currently possible to do via the build method.

ymc-dabe:
Resolves #1498

Signed-off-by: Daniel Beyer <ymc-dabe@vcs.ymc.ch>

### DIFF
--- a/docker/api/build.py
+++ b/docker/api/build.py
@@ -1,6 +1,7 @@
 import json
 import logging
 import os
+import six
 
 from .. import auth
 from .. import constants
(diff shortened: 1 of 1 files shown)


shin-:
Given the recent developments and the upcoming inclusion of buildkit into Docker, it's unlikely the official implementation of build-time mounts will look like RedHat's version. See https://github.com/moby/buildkit/pull/442 for details.

I'm closing this as a result, but I'm happy to look at it again if things evolve in a different direction.

socketpair:
so, what is minimal version of Docker where `RUN --mount=....` works ?

PR #1942, opened 2018-03-05
Ignore comment line from the .dockerignore file 

levin-du:
This fixed the bug that test comment line in .dockerignore file as ignore rule
bug.

Add test for "# comment" patterns in .dockerignore.


levin-du:
ping @shin- 🤗  thks

shin-:
Thank you!

PR #1951, opened 2018-03-13
stop(), restart(): Adapt request timeout to user-provided graceful timeout

mnottale:

### DIFF
--- a/docker/api/container.py
+++ b/docker/api/container.py
@@ -1007,7 +1007,10 @@ class ContainerApiMixin(object):
         """
         params = {'t': timeout}
         url = self._url("/containers/{0}/restart", container)
-        res = self._post(url, params=params)
+        conn_timeout = self.timeout
+        if conn_timeout:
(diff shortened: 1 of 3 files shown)


docker/api/container.py
```diff
@@ -1007,7 +1007,10 @@ def restart(self, container, timeout=10):
         """
         params = {'t': timeout}
         url = self._url("/containers/{0}/restart", container)
-        res = self._post(url, params=params)
+        conn_timeout = self.timeout
+        if conn_timeout:
+            conn_timeout = max(conn_timeout, timeout+10)
```
adw1n: How did you decide on the `+10` part? IMO it's not enough, look here: https://github.com/moby/moby/blob/master/daemon/container_operations_unix.go#L341 When SIGTERM fails to kill the container dockerd will send a SIGKILL and sleep for additional 10 seconds.

shin-:
Merged in #2008 :+1:

Issue #1929, opened 2018-02-26
How to close the socket used by Dockerclient(base_url)


ywang412:
I cannot find a Dockerclient.close() API. My code gave me warning about unclosed socket. Although this is not error, is it possible to close the socket elegantly?

shin-:
DockerClient.api.close() should work!

ywang412:
Thanks!

PR #1954, opened 2018-03-14
Add close() method to DockerClient.

mnottale:


mnottale:
Fixes #1929

Issue #1934, opened 2018-02-28
containers.list(all=True) does not respond while any containers in the list are in transition


yugokato:
containers.list(all=True) does not return the result while any containers in the list are stopping/starting/restarting. 
It waits until stopping/starting/restarting of containers is done, which makes the code blocking.

The equivalent docker command 'docker ps -a' shows result in real-time even containers are in transition.

VillanCh:
The similar situation I met, sometimes the `containers.list` blocks all my operations! OMG

reverland:
@shin- 

The similar situation I met, it just blocks like frozen.

shin-:
I'm guessing this is due to `list()` being greedy and calling `inspect` on each container, which might block while the container is in a transitive state. Are you seeing the same issue with `client.api.containers(all=True)`?

We could probably add an option to `list()` to create `Container` objects based solely on the partial information provided by the `/containers` endpoint and let users call `reload()` explicitly to retrieve the complete attribute set.

VillanCh:
Awosome! You are right, thanks for your advice, the low-level api `client.api.containers(all=True)` is [Similar to the docker ps command](https://docker-py.readthedocs.io/en/stable/api.html#docker.api.container.ContainerApiMixin.containers) . It looks like never frozen.

Maybe, is there the same situation as `client.images.list`? I guess.

yugokato:
client.api.containers(all=True) worked fine as you suggested. It would be nice if the option to create Container objects with limited information without blocking is available.

shin-:
Fixed by #1955

PR #1955, opened 2018-03-14
Add sparse argument to DockerClient.containers.list().

mnottale:


mnottale:
Fixes #1934

PR #1961, opened 2018-03-16
avoid race condition in containers.list

### LINKED ISSUE #1485: containers.list() fails if a container is removed while it's being called
I'm getting the following error in an application that uses docker-py with fairly heavy container churn. 

```
  File "executor.py", line 44, in event_loop
    num_executors = len(get_executor_containers())
  File "executor.py", line 186, in get_executor_containers
    running_containers = client.containers.list(all=True)
  File "/home/ben/envs/executor/lib/python3.6/site-packages/docker/models/containers.py", line 767, in list
    return [self.get(r['Id']) for r in resp]
  File "/home/ben/envs/executor/lib/python3.6/site-packages/docker/models/containers.py", line 767, in <listcomp>
    return [self.get(r['Id']) for r in resp]
  File "/home/ben/envs/executor/lib/python3.6/site-packages/docker/models/containers.py", line 720, in get
    resp = self.client.api.inspect_container(container_id)
  File "/home/ben/envs/executor/lib/python3.6/site-packages/docker/utils/decorators.py", line 21, in wrapped
    return f(self, resource_id, *args, **kwargs)
  File "/home/ben/envs/executor/lib/python3.6/site-packages/docker/api/container.py", line 745, in inspect_container
    self._get(self._url("/containers/{0}/json", container)), True
  File "/home/ben/envs/executor/lib/python3.6/site-packages/docker/api/client.py", line 220, in _result
  File "/home/ben/envs/executor/lib/python3.6/site-packages/docker/api/client.py", line 216, in _raise_for_status
    raise create_api_error_from_http_exception(e)
  File "/home/ben/envs/executor/lib/python3.6/site-packages/docker/errors.py", line 30, in create_api_error_from_http_exception
    raise cls(e, response=response, explanation=explanation)
docker.errors.NotFound: 404 Client Error: Not Found for url: http+docker://localunixsocket/v1.24/containers/762ad9c915018c1d38c0ae71ab44ce69b3750b70b0524d0ec6e01dd62bd7fe48/json ("No such container: 762ad9c915018c1d38c0ae71ab44ce69b3750b70b0524d0ec6e01dd62bd7fe48")
```

It seems to me that list() gets all containers from the api, then *while* it's iterating through that list, a container in the list is removed. This causes the call to list() to fail with a NotFound exception.

I think it could be fixed by replacing models/containers.py:767
```python
return [self.get(r['Id']) for r in resp]
```
with
```python
containers = []
for r in resp:
    try:
        containers.append(self.get(r['id']))
    except NotFound: # a container may have been removed while iterating
        pass
return containers
```
but I'm not sure if that would cause unforeseen issues.

I'm seeing this with a few (3 or 4) containers running at a time, but I think the issue would be much worse if you had many containers running at once.

## Misc info
```
→ pip freeze | grep docker && python --version && docker version

docker==2.1.0
docker-pycreds==0.2.1
Python 3.6.0
Client:
 Version:      1.13.1
 API version:  1.26
 Go version:   go1.7.5
 Git commit:   092cba3727
 Built:        Sun Feb 12 02:40:56 2017
 OS/Arch:      linux/amd64

Server:
 Version:      1.13.1
 API version:  1.26 (minimum version 1.12)
 Go version:   go1.7.5
 Git commit:   092cba3727
 Built:        Sun Feb 12 02:40:56 2017
 OS/Arch:      linux/amd64
 Experimental: false
```

```
→ uname -a
Linux jubal 4.9.11-1-ARCH #1 SMP PREEMPT Sun Feb 19 13:45:52 UTC 2017 x86_64 GNU/Linux
```
```
→ cat /etc/issue
Arch Linux \r (\l)
```

BenDoan:
Signed-off-by: Ben Doan <ben@bendoan.me>

If a container is removed when running containers.list()  while it's iterating through the list of container ids, a NotFound exception is thrown which kills the whole call.

If even a couple containers a second are being removed, container.list() doesn't really work.

This change just catches and ignores the NotFound exceptions while iterating through the list of container ids.

fixes #1485

### DIFF
--- a/docker/models/containers.py
+++ b/docker/models/containers.py
@@ -6,7 +6,7 @@ from ..api import APIClient
 from ..constants import DEFAULT_DATA_CHUNK_SIZE
 from ..errors import (
     ContainerError, DockerException, ImageNotFound,
-    create_unexpected_kwargs_error
+    NotFound, create_unexpected_kwargs_error
 )
 from ..types import HostConfig
 from ..utils import version_gte
@@ -896,7 +896,14 @@ class ContainerCollection(Collection):
         if sparse:
             return [self.prepare_model(r) for r in resp]
         else:
-            return [self.get(r['Id']) for r in resp]
+            containers = []
+            for r in resp:
+                try:
+                    containers.append(self.get(r['Id']))
+                # a container may have been removed while iterating
+                except NotFound:
+                    pass
+            return containers
 
     def prune(self, filters=None):
         return self.client.api.prune_containers(filters=filters)


shin-:
Thank you for submitting a PR!

I'm concerned that this could be considered a breaking change. While the behavior is probably desirable in 99% of cases, I'm guessing there's scenarios where ignoring these failures isn't correct.
Maybe we can make it an option for now (`ignore_removed`?), and make it the default for the next major version bump?

BenDoan:
Thanks for the response!

Using sparse mode should avoid this issue as well, so I don't think it makes too much sense to add an `ignore_removed` option.

I'm not sure I agree that this a breaking change. I can't think of any legitimate case where a user would be relying on 404s thrown while list() is iterating, and the bug makes containers.list() useless for some users. 

That being said, I definitely understand wanting to be cautious about [breaking existing users](https://xkcd.com/1172/).

PR #1965, opened 2018-03-19
Allow cancelling streams from other threads 

shin-:
Based on #1917 

Rebased + small reliability changes

cc @rycus86 PTAL I went with your suggestion to use `pytest-timeout`. I believe I also fixed the 2.7 flakiness by removing the useless `select` call in `socket.py`.

### DIFF
--- a/docker/api/container.py
+++ b/docker/api/container.py
@@ -5,7 +5,8 @@ from .. import errors
 from .. import utils
 from ..constants import DEFAULT_DATA_CHUNK_SIZE
 from ..types import (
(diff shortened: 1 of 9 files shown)


rycus86:
Awesome, thanks!
I'll have a quick look at it in the evening.

shin-:
Ah, I broke python 3 apparently :|

rycus86:
I've run it on Python 2.7 and 3.6 a *good* few times and couldn't make it fail.
Thanks a lot again for taking this forward!
LGTM

shin-:
Thank you for testing / reviewing :+1:

PR #1977, opened 2018-03-26
 Support building with Dockerfile outside of context

shin-:
Based on docker/cli#886

Fixes docker/compose#5823

### DIFF
--- a/docker/utils/utils.py
+++ b/docker/utils/utils.py
@@ -1,17 +1,13 @@
 import base64
-import io
 import os
 import os.path
 import json
 import shlex
-import tarfile
-import tempfile
 from distutils.version import StrictVersion
 from datetime import datetime
 
 import six
 
-from .. import constants
 from .. import errors
 from .. import tls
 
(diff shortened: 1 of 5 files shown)


docker/utils/utils.py
```diff
@@ -88,13 +88,17 @@ def build_file_list(root):
...
-def create_archive(root, files=None, fileobj=None, gzip=False):
+def create_archive(root, files=None, fileobj=None, gzip=False,
+                   extra_files=None):
     if not fileobj:
         fileobj = tempfile.NamedTemporaryFile()
     t = tarfile.open(mode='w:gz' if gzip else 'w', fileobj=fileobj)
     if files is None:
         files = build_file_list(root)
     for path in files:
+        if path in [e[0] for e in extra_files]:
```
mefyl: Nitpicking, but `any(e[0] == path for e in extra_files)` won't build the whole list in memory. Maybe even building a `set(e[0] for e in extra_files)` before to avoid iterating too many times.

shin-:
Thank you for reviewing! :+1:

thaJeztah:
Thanks @shin- ! (Not enough experience with python though to review 😅)

Issue #1981, opened 2018-03-30
Building image from different drive is not working


kant2002:
When invoking building image from drive different then location of Dockerfile, the exception appears

```
# Current directory is  D:\GitHub\sc-docker
base_dir = "C:\Users\user\AppData\Roaming/scbw\docker"
# File C:\Users\user\AppData\Roaming/scbw\docker\game.dockerfile is present
docker_client.images.build(path=base_dir, dockerfile="game.dockerfile", tag=local_image)
```

This code produces 

```
  File "D:\github\sc-docker\scbw\docker_utils.py", line 127, in ensure_local_image
    docker_client.images.build(path=base_dir, dockerfile="game.dockerfile", tag=local_image)
  File "d:\github\sc-docker\env\py3.6\lib\site-packages\docker\models\images.py", line 258, in build
    resp = self.client.api.build(**kwargs)
  File "d:\github\sc-docker\env\py3.6\lib\site-packages\docker\api\build.py", line 154, in build
    print(os.path.relpath(dockerfile, path))
  File "d:\github\sc-docker\env\py3.6\lib\ntpath.py", line 585, in relpath
    path_drive, start_drive))
ValueError: path is on mount 'D:', start on mount 'C:'
```

Environment. 
```
docker==3.2.0
Python 3.6.3
Client:
 Version:       18.01.0-ce
 API version:   1.35
 Go version:    go1.9.2
 Git commit:    03596f51b1
 Built: Thu Jan 11 22:29:41 2018
 OS/Arch:       windows/amd64
 Experimental:  false
 Orchestrator:  swarm

Server:
 Engine:
  Version:      18.01.0-ce
  API version:  1.35 (minimum version 1.12)
  Go version:   go1.9.2
  Git commit:   03596f5
  Built:        Wed Jan 10 20:13:12 2018
  OS/Arch:      linux/amd64
  Experimental: false
```

The code is same for current version in the master, so issue should be applicable to current version as well.

Original issue is https://github.com/Games-and-Simulations/sc-docker/issues/59

PR #1982, opened 2018-03-30
Ignore error when Dockerfile and current dir located on different drives

kant2002:
Related to #1981


shin-:
Thanks! We're probably going to go with https://github.com/docker/docker-py/pull/1983 - do you mind checking that it fixes the issue for you as well?

PR #1991, opened 2018-04-13
Set minimum version for configs api to 1.30

ushuz:
According to documentation, configs api was introduced in 1.30:
https://docs.docker.com/engine/reference/commandline/config/
https://docs.docker.com/engine/api/v1.30/

1.25 is the minimum version for secrets api.


shin-:
Thank you!

Issue #2013, opened 2018-04-29
Service scaling does not happen while returning True


gurayyildirim:
When trying to scale up/down a service with service.scale() and it returns True, the task count does not change. So scaling needs to be fixed with these versions:

```
docker==3.3.0
docker-pycreds==0.2.3
Python 2.7.14

Client:
 Version:      18.03.1-ce
 API version:  1.37
 Go version:   go1.9.5
 Git commit:   9ee9f40
 Built:        Thu Apr 26 07:13:02 2018
 OS/Arch:      darwin/amd64
 Experimental: false
 Orchestrator: swarm

Server:
 Engine:
  Version:      18.03.1-ce
  API version:  1.37 (minimum version 1.12)
  Go version:   go1.9.5
  Git commit:   9ee9f40
  Built:        Thu Apr 26 07:22:38 2018
  OS/Arch:      linux/amd64
  Experimental: true
```

PR #2014, opened 2018-04-29
Swarm Mode service scaling parameter mistake is fixed. #2013

gurayyildirim:
Signed-off-by: Güray Yıldırım <guray@gurayyildirim.com.tr>

The problem referenced in #2013 is fixed with adding named parameter.


shin-:
Thanks! CI failure looks unrelated.

PR #2016, opened 2018-05-01
Use six.moves to handle a py2+py3 import

asottile:


asottile:
very very minor, noticed this while debugging another issue

shin-:
Thanks @asottile !

PR #2036, opened 2018-05-24
Add or condition for None

### LINKED ISSUE #1970: service list fail if empty
Calling the client with some filters, if no services matches a exception is thrown:
client.services.list(filters=filters)

  File "************docker/docker/dockercli.py", line 13, in listServices
    services = self.client.services.list(filters=filters)
  File "*********/lib/python3.5/site-packages/docker/models/services.py", line 259, in list
    for s in self.client.api.services(**kwargs)
TypeError: 'NoneType' object is not iterable

srinivasreddy:
Fixes #1970

### DIFF
--- a/docker/models/services.py
+++ b/docker/models/services.py
@@ -256,7 +256,7 @@ class ServiceCollection(Collection):
         """
         return [
             self.prepare_model(s)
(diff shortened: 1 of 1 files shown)


srinivasreddy:
When kwargs are `None`

Issue #2048, opened 2018-06-04
Install plugin with registry authentication


overmike:
Is there a way to install a plugin if authentication is required?

shin-:
It should work ; make sure your credentials are in the Docker config file (or in the credential store you're using), or call `client.login()` first.

overmike:
We are using a private registry with token auth setup. I can install the plugin by docker cli but fail to use docker-py to install.

from the code, seems like only "pull_plugin" add the "X-Registry-Auth" header but not in "plugin_privileges". Does this mean either (1) the token auth service should allow unauth plugin privilege query, or (2) docker-py should include auth header when query plugin privilege

>>> d.plugins.install(remote_name='redacted.host/volume')
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/local/lib/python2.7/site-packages/docker/models/plugins.py", line 182, in install
    privileges = self.client.api.plugin_privileges(remote_name)
  File "/usr/local/lib/python2.7/site-packages/docker/utils/decorators.py", line 34, in wrapper
    return f(self, *args, **kwargs)
  File "/usr/local/lib/python2.7/site-packages/docker/api/plugin.py", line 171, in plugin_privileges
    return self._result(self._get(url, params=params), True)
  File "/usr/local/lib/python2.7/site-packages/docker/api/client.py", line 228, in _result
    self._raise_for_status(response)
  File "/usr/local/lib/python2.7/site-packages/docker/api/client.py", line 224, in _raise_for_status
    raise create_api_error_from_http_exception(e)
  File "/usr/local/lib/python2.7/site-packages/docker/errors.py", line 31, in create_api_error_from_http_exception
    raise cls(e, response=response, explanation=explanation)
docker.errors.APIError: 500 Server Error: Internal Server Error ("Get http://redacted.host/v2/volume/manifests/latest: unauthorized: authentication required")


$ docker plugin install redacted.host/volume
Plugin "redacted.host/volume" is requesting the following privileges:
 - network: [host]
 - mount: [/dev]
 - mount: [/var/lib/docker]
 - allow-all-devices: [true]
 - capabilities: [CAP_SYS_ADMIN]
Do you grant the above permissions? [y/N]
....

shin-:
[...]

PR #2052, opened 2018-06-05
query plugin privilege with registry auth header

overmike:
#2048 Please accept the PR, which is able to install a plugin from private registry with token auth setup


overmike:
May I know when is the tentative 3.4.0 release date?

PR #2063, opened 2018-06-19
Add support for `uts_mode` parameter in `Client.create_host_config`.

mtszb:
This parameter allows to set the UTS namespace of the container, as in the
`--uts=X` attribute of Docker CLI:
<https://docs.docker.com/engine/reference/run/#uts-settings---uts>
The only allowed value, if set, is "host".

Also included are changes to the tests to reflect the new added parameter.

Signed-off-by: Marco Trillo <martri@arantia.com>
Signed-off-by: Diego Alvarez <dyako.developer@gmail.com>

### DIFF
--- a/docker/types/containers.py
+++ b/docker/types/containers.py
@@ -115,11 +115,11 @@ class HostConfig(dict):
                  device_read_iops=None, device_write_iops=None,
                  oom_kill_disable=False, shm_size=None, sysctls=None,
                  tmpfs=None, oom_score_adj=None, dns_opt=None, cpu_shares=None,
-                 cpuset_cpus=None, userns_mode=None, pids_limit=None,
-                 isolation=None, auto_remove=False, storage_opt=None,
-                 init=None, init_path=None, volume_driver=None,
-                 cpu_count=None, cpu_percent=None, nano_cpus=None,
-                 cpuset_mems=None, runtime=None, mounts=None,
+                 cpuset_cpus=None, userns_mode=None, uts_mode=None,
+                 pids_limit=None, isolation=None, auto_remove=False,
+                 storage_opt=None, init=None, init_path=None,
+                 volume_driver=None, cpu_count=None, cpu_percent=None,
+                 nano_cpus=None, cpuset_mems=None, runtime=None, mounts=None,
                  cpu_rt_period=None, cpu_rt_runtime=None,
                  device_cgroup_rules=None):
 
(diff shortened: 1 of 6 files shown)


docker/types/containers.py
```diff
@@ -115,7 +115,7 @@ def __init__(self, version, binds=None, port_bindings=None,
                  device_read_iops=None, device_write_iops=None,
                  oom_kill_disable=False, shm_size=None, sysctls=None,
                  tmpfs=None, oom_score_adj=None, dns_opt=None, cpu_shares=None,
-                 cpuset_cpus=None, userns_mode=None, pids_limit=None,
+                 cpuset_cpus=None, userns_mode=None, uts=None, pids_limit=None,
```
shin-: can we rename the parameter to `uts_mode` across the board? That way it matches the API name, is more explicit, and if more `uts`-related parameters are introduced in the future, they'll be easier to integrate.

docker/types/containers.py
```diff
@@ -392,6 +392,11 @@ def __init__(self, version, binds=None, port_bindings=None,
                 raise host_config_value_error("userns_mode", userns_mode)
             self['UsernsMode'] = userns_mode
 
+        if uts:
+            if uts != "host":
```
shin-: Can we add a version check here as well? I believe this was only introduced recently.

tests/unit/models_containers_test.py
```diff
@@ -95,6 +95,7 @@ def test_create_container_args(self):
             ulimits=[{"Name": "nofile", "Soft": 1024, "Hard": 2048}],
             user='bob',
             userns_mode='host',
+            uts='host',
```
shin-: Can we have an integration test as well? They're a much better indicator of things working as intended for this kind of thing.

docker/types/containers.py
mtszb: Yep, that name it is more sensible. The updated pull request uses the "uts_mode" attribute name as suggested.

tests/unit/models_containers_test.py
mtszb: I have included a simple integration test, `test_create_with_uts_mode`, similar in style to the others container API tests already present.

docker/types/containers.py
mtszb: Hmm... as far as I am aware of, it seems that the `UTSMode` attribute was already present when the `HostConfig` structure itself was moved to within the `Container` API type; that's why I did not include a version check for that attribute.

The actual version when this attribute was introduced seems to be around 1.7, but in that version the `HostConfig` structure was in `runconfig`.

shin-: yup, my bad, looks like it has existed for a long time now.

shin-:
Thank you!

PR #2088, opened 2018-07-18
honor placement preferences via services.create()

knackaron:
this allows creating a service with placement preferences when
calling services.create().  only constraints were being honored
before.

related to https://github.com/docker/docker-py/pull/1615


shin-:
Thank you!

PR #2047, opened 2018-06-02
upgrade Windows dependency

ofek:
`pywin32` moved to GitHub and now ships binary wheels itself. Also, it seems the bug introduced in `222` has been fixed.

https://github.com/mhammond/pywin32
https://pypi.org/project/pywin32/#files
https://github.com/mhammond/pywin32/pull/1153

PR #2089, opened 2018-07-19
Update deps for 3.3 & 3.7 support

shin-:
Note: py33 removed from envlist in tox.ini as it's incompatible with newer version of wheels, meaning the env setup is broken even with older versions of `tox`. Support for Python 3.3 will be dropped in the 4.0 release.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,6 +10,10 @@ matrix:
       env: TOXENV=py35
     - python: 3.6
       env: TOXENV=py36
+    - python: 3.7
+      env: TOXENV=py37
+      dist: xenial
+      sudo: true
     - env: TOXENV=flake8
 
 install:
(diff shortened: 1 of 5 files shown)


ofek:
Can we try to incorporate this? https://github.com/docker/docker-py/pull/2047

shin-:
@ofek Done - we still need to pin on 219 for the older versions though.

ofek:
Nice! Though I meant could we try to make use of their new official packages?

official: https://pypi.org/project/pywin32/#files
deprecated: https://pypi.org/project/pypiwin32/#files

shin-:
Hmm, that official package is missing versions (including 219, which we still want to use for the foreseeable future), and I'm concerned about the potential conflicts that could arise from including the same package from two different sources, even if they're mutually exclusive. On top of that, when I've tried to use it in the past, it was just completely broken.

I'm more comfortable with sticking what works for the time being. If `pypiwin32` starts lagging behind significantly, we can revisit.

PR #2091, opened 2018-07-20
Add 'rollback' command as allowed for failure_action

NikolayMurha:


docker/types/services.py
```diff
@@ -385,7 +385,7 @@ def __init__(self, parallelism=0, delay=None, failure_action='continue',
         self['Parallelism'] = parallelism
         if delay is not None:
             self['Delay'] = delay
-        if failure_action not in ('pause', 'continue'):
+        if failure_action not in ('pause', 'continue', 'rollback'):
             raise errors.InvalidArgument(
                 'failure_action must be either `pause` or `continue`.'
```
withsmilo: Add 'rollback' to error message.

PR #2099, opened 2018-07-30
images/create: add support for specifying custom message

rindeal:
API docs do not mention this parameter, but CLI and server supports it for 3 years now -
https://github.com/moby/moby/commit/936b2c6afe490b562df4ebea854b116c3b40a9f5.

Signed-off-by: Jan Chren <dev.rindeal@gmail.com>


rindeal:
It is now part of the API docs https://docs.docker.com/engine/api/v1.40/#operation/ImageCreate

PR #2103, opened 2018-08-02
Allow user=0 to be passed in create_container

asottile:


asottile:
@shin- does this look ok?

shin-:
LGTM, thank you!

PR #2111, opened 2018-08-12
Document defaults of logs()

bfirsh:
This is not obvious because some are True by default.

(Also hi! :wave:)


shin-:
:wave: @bfirsh 

Changes LGTM - sorry for the delay!

PR #2121, opened 2018-08-17
Fixes to logs --follow

kolyshkin:
Since https://github.com/moby/moby/pull/37656 (because it fixes https://github.com/moby/moby/issues/37630), logs --follow do not send EOF once a container is stopped.

1. `test_logs_streaming_and_follow()` is modified to remove a container,
otherwise it will stuck forever waiting for more logs. As the container
is forcibly killed, the check for exit code is removed.

2. `ContainerCollection.run()` is modified to have `follow=False` for logs.
The `follow=True` that was there is probably from some ancient times
when it meant to read all the saved logs and then attach to container's
live stdout/stderr stream. This is not the case anymore.

Without this change, `test_run_with_json_file_driver()` is stuck forever.

NOTE these changes are backward-compatible, i.e. they do not require the
above mentioned PR to work; any (decent) version of docker engine is
fine.

### DIFF
--- a/docker/models/containers.py
+++ b/docker/models/containers.py
@@ -777,16 +777,15 @@ class ContainerCollection(Collection):
         logging_driver = container.attrs['HostConfig']['LogConfig']['Type']
 
         out = None
-        if logging_driver == 'json-file' or logging_driver == 'journald':
-            out = container.logs(
-                stdout=stdout, stderr=stderr, stream=True, follow=True
-            )
-
         exit_status = container.wait()['StatusCode']
         if exit_status != 0:
-            out = None
             if not kwargs.get('auto_remove'):
                 out = container.logs(stdout=False, stderr=True)
+        else:
+            if logging_driver == 'json-file' or logging_driver == 'journald':
+                out = container.logs(
+                    stdout=stdout, stderr=stderr, stream=True, follow=False
+                )
 
         if remove:
             container.remove()
--- a/tests/integration/api_container_test.py
+++ b/tests/integration/api_container_test.py
@@ -873,12 +873,11 @@ Line2'''
         id = container['Id']
         self.tmp_containers.append(id)
         self.client.start(id)
+        threading.Timer(1, self.client.remove_container,
+                        [id], {'force': True}).start()
         logs = six.binary_type()
         for chunk in self.client.logs(id, stream=True, follow=True):
             logs += chunk
-
-        exitcode = self.client.wait(id)['StatusCode']
-        assert exitcode == 0
 
         assert logs == (snippet + '\n').encode(encoding='ascii')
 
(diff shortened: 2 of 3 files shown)


shin-:
@kolyshkin Thanks! Just a style issue needs to be addressed for the CI to turn green it seems:
```
tests\integration\api_container_test.py:876:80: E501 line too long (87 > 79 characters)
```

docker/models/containers.py
```diff
@@ -779,7 +779,7 @@ def run(self, image, command=None, stdout=True, stderr=False,
         out = None
         if logging_driver == 'json-file' or logging_driver == 'journald':
             out = container.logs(
-                stdout=stdout, stderr=stderr, stream=True, follow=True
+                stdout=stdout, stderr=stderr, stream=True, follow=False
```
shin-: This change concerns me because it means we're losing any logs that happen after the method is called. I don't think that is correct.

We might want to use `attach` instead?

kolyshkin:
> tests\integration\api_container_test.py:876:80: E501 line too long (87 > 79 characters)

fixed

docker/models/containers.py
kolyshkin: I guess it's easier to call container.logs after container is stopped. What about this:

```diff
--- a/docker/models/containers.py
+++ b/docker/models/containers.py
@@ -777,16 +777,15 @@ class ContainerCollection(Collection):
         logging_driver = container.attrs['HostConfig']['LogConfig']['Type']
 
         out = None
-        if logging_driver == 'json-file' or logging_driver == 'journald':
-            out = container.logs(
-                stdout=stdout, stderr=stderr, stream=True, follow=False
-            )
-
         exit_status = container.wait()['StatusCode']
         if exit_status != 0:
-            out = None
             if not kwargs.get('auto_remove'):
                 out = container.logs(stdout=False, stderr=True)
+        else:
+            if logging_driver == 'json-file' or logging_driver == 'journald':
+                out = container.logs(
+                    stdout=stdout, stderr=stderr, stream=True, follow=False
+                )
 
         if remove:
             container.remove()
```

kolyshkin: One thing I am not sure is how it works (and worked before) in case `remove=True`.

shin-: That change looks fine to me.

shin-:
Another pep8 issue. FWIW you can run `make flake8` to catch these quickly
```
tests/integration/api_container_test.py:877:17: E128 continuation line under-indented for visual indent
```

kolyshkin:
patch updated based on review comments

docker/models/containers.py
kolyshkin: Yeah, `container.logs()` after container stop won't work in case of `auto_remove=True`.

We can use `container.logs` with `follow=True` in a separate thread and kill it once the container is stopped. The problem is I am not quite sure how to implement it in Python :(

kolyshkin: Or maybe we just need to do `out.Close()` in proper places...

kolyshkin:
OK the suggestion to gather container logs after container stop won't work in case `auto_remove=True`, so this has to be reverted. Perhaps we just need to do `out.Close()` once the container is stopped?

kolyshkin:
I gave up on the idea of changing the behavior for `ContainerLogs(follow=true)`, as it is not backward-compatible; thus closing this PR.

PR #2130, opened 2018-09-03
Fix pulling images with `stream=True`

### LINKED ISSUE #2116: Can't pull image with stream=True
Hi,

The scenario is as follows:

Mac 10.13.6
docker version v18.06.0-ce
Python 3.6
(python package) docker==3.5.0
private docker registry (docker hub, private repo)
docker login works ✔️
docker pull $image works ✔️
however, pulling via the docker python api fails when using the parameter `stream=True`:
In terminal window I run :
```
python
>>> client = docker.DockerClient(base_url='unix://var/run/docker.sock')
>>> client.login(username='XXXX', password='XXXX', registry='https://index.docker.io/v1/')
>>> client.images.pull('docker.io/vdmtl/portail-datamigration-worker-gcs-lib:latest', stream=True)

  File "/Users/XXXX/apps/open-data-etl/venv/lib/python3.6/site-packages/docker/api/client.py", line 229, in _raise_for_status
    response.raise_for_status()
  File "/Users/XXX/apps/open-data-etl/venv/lib/python3.6/site-packages/requests/models.py", line 937, in raise_for_status
    raise HTTPError(http_error_msg, response=self)
requests.exceptions.HTTPError: 404 Client Error: Not Found for url: http+docker://localhost/v1.35/images/docker.io/vdmtl/portail-datamigration-worker-gcs-lib:latest/json

>>> client.images.pull('docker.io/vdmtl/portail-datamigration-worker-gcs-lib:latest')
<Image: 'vdmtl/portail-datamigration-worker-gcs-lib:latest'>
```
As you can see removing `stream=True` is necessary in order to download my image.

adw1n:
Pulling an image with option `stream=True` like this:
```
client.api.pull('docker.io/user/repo_name', tag='latest', stream=True)
```
without consuming the generator oftentimes results in premature drop of the connection. Docker daemon tries to send progress of pulling the image to the client, but it encounters an error (broken pipe) and therefore cancells the pull action:
```
Thread 1 "dockerd-dev" received signal SIGPIPE, Broken pipe.
ERRO[2018-09-03T05:12:35.746497638+02:00] Not continuing with pull after error: context canceled
```
As described in issue #2116, even though client receives response with status code 200, image is not pulled.

Closes #2116

Signed-off-by: Przemysław Adamek adw1n@users.noreply.github.com

### DIFF
--- a/docker/models/images.py
+++ b/docker/models/images.py
@@ -409,6 +409,7 @@ class ImageCollection(Collection):
         if not tag:
             repository, tag = parse_repository_tag(repository)
 
+        kwargs['stream'] = False
         self.client.api.pull(repository, tag=tag, **kwargs)
         if tag:
             return self.get('{0}{2}{1}'.format(
--- a/docker/api/image.py
+++ b/docker/api/image.py
@@ -334,7 +334,8 @@ class ImageApiMixin(object):
         Args:
             repository (str): The repository to pull
             tag (str): The tag to pull
-            stream (bool): Stream the output as a generator
+            stream (bool): Stream the output as a generator. Make sure to
+                consume the generator, otherwise pull might get cancelled.
             auth_config (dict): Override the credentials that
                 :py:meth:`~docker.api.daemon.DaemonApiMixin.login` has set for
                 this request. ``auth_config`` should contain the ``username``
--- a/tests/unit/models_containers_test.py
+++ b/tests/unit/models_containers_test.py
@@ -232,7 +232,8 @@ class ContainerCollectionTest(unittest.TestCase):
         container = client.containers.run('alpine', 'sleep 300', detach=True)
 
         assert container.id == FAKE_CONTAINER_ID
-        client.api.pull.assert_called_with('alpine', platform=None, tag=None)
+        client.api.pull.assert_called_with('alpine', platform=None, tag=None,
+                                           stream=False)
 
     def test_run_with_error(self):
         client = make_fake_client()
--- a/tests/unit/models_images_test.py
+++ b/tests/unit/models_images_test.py
@@ -43,7 +43,8 @@ class ImageCollectionTest(unittest.TestCase):
     def test_pull(self):
         client = make_fake_client()
         image = client.images.pull('test_image:latest')
-        client.api.pull.assert_called_with('test_image', tag='latest')
+        client.api.pull.assert_called_with('test_image', tag='latest',
+                                           stream=False)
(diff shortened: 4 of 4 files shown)


docker/models/images.py
```diff
@@ -409,6 +409,7 @@ def pull(self, repository, tag=None, **kwargs):
         if not tag:
             repository, tag = parse_repository_tag(repository)
 
+        kwargs['stream'] = False
```
adw1n: I doubt that that the pull log can be big enough to cause memory issues.

shin-:
Hi @adw1n , thanks for the submission.

I'm not sure that's the right approach. Forcing `stream=False` seems hamfisted. If the problem is the one you describe, maybe we simply need a warning in the `DockerClient` documentation that when setting `stream=True`, the generator must be consumed for the operation to complete?

adw1n:
@shin- 
> Forcing stream=False seems hamfisted.

I only suggested forcing it in `docker.models.images.pull`, because I don't see any benefits to using `stream=True` there. The only benefit to using `stream=True` is if you fear that the pull log can be very big (like >1MB) and your app will run out of memory.I doubt that can happen.

I do not force it in docker.api.image.pull


>  maybe we simply need a warning in the DockerClient documentation

I agree about adding the documentation (I've done it in `docker.api.image.pull`), although code in `docker.models.images.pull` definitelly needs to be changed, otherwise `self.client.api.pull(repository, tag=tag, stream=True)` will be called by the library code without consuming the generator.


My suggestions for `docker.models.images.pull`:
1. force `stream=False` (`kwargs.pop('stream')` or `kwargs['stream']=False`) + optionaly log a warning that `stream=False` doesn't make sense here
2. consume the generator if `stream=True` (best choice if the pull log can be big):
```
++        if kwargs.get('stream'):
++            for _ in self.client.api.pull(repository, tag=tag, **kwargs): pass
++        else:
++            self.client.api.pull(repository, tag=tag, **kwargs)
--        self.client.api.pull(repository, tag=tag, **kwargs)
```
or test if result of `pull` is a generator and consume it depending on that
3. 
```
++ list(self.client.api.pull(repository, tag=tag, **kwargs))
-- self.client.api.pull(repository, tag=tag, **kwargs)
```

adw1n:
@shin- ping

shin-:
Sorry this took longer than it should have! I've made a few updates and merged the code in #2186 .

Thank you for your help and your patience!

PR #2150, opened 2018-10-08
implement stream demultiplexing for exec commands

### LINKED ISSUE #1952: Demux stdout and stderr from result of attach() command
When I run `container.attach(stdout=True, stderr=True, stream=True, logs=True)`, the resulting generator contains the output of stdout and stderr multiplexed into one string. I would like to separate to get stdout and stderr in separate generators (or) streams. 

I know the Go SDK contains a [`stdcopy.StdCopy`](https://github.com/moby/moby/blob/master/pkg/stdcopy/stdcopy.go#L94) precisely for the purpose of demuxing the output given a raw socket. I would like to implement something similar for the Docker Python SDK as well. 

I need this for my project, so I am going to implement anyway, but would you be open to the idea of my contributing the patch back to this project? Actually most of the functionality is already implemented in [socket.py](https://github.com/docker/docker-py/blob/master/docker/utils/socket.py). Instead of [discarding the frame type](https://github.com/docker/docker-py/blob/master/docker/utils/socket.py#L64), I will pass it out and have the caller decide how to handle it. The `attach()` method will discard the type, whereas, a new `docker.util.demuxSocketOutput(stdout_stream, stderr_stream)` will write the output to respective the streams.

Would you be open to this idea? I send a PR.

little-dude:
Hi there,

This is an attempt to fix #1952: offering the possibility to demultiplex stdout, stderr and stdin. Currently, it's only implemented for `docker exec` but can be extended to other APIs where it's relevant.

The implementation is not complete: some tests need to be added, and some needs to be adjusted. But since I'm new to this codebase, I wanted to open the PR to get early feedback, before spending time polishing it.

Here is an example of how it works:

```python
# Create a container that runs in the background
>>> container = client.containers.run(
...     'bfirsh/reticulate-splines', detach=True)

# Prepare the command we are going to use. It prints "hello stdout"
# in `stdout`, followed by "hello stderr" in `stderr`:
>>> cmd = '/bin/sh -c "echo hello stdout ; echo hello stderr >&2"'

# With stream=False and demux=False, the output is a string
# that contains both the stdout and the stderr output:
>>> res = container.exec_run(cmd, stream=False, demux=False)
>>> res.output
b'hello stderr\nhello stdout\n'

# With stream=True, and demux=False, the output is a
# generator that yields strings containing the output of both
# `stdout` and `stderr`:
>>> res = container.exec_run(cmd, stream=True, demux=False)
>>> next(res.output)
b'hello stdout\n'
>>> next(res.output)
b'hello stderr\n'
>>> next(res.output)
StopIteration

# With stream=True and demux=True, the generator now
# separates the streams, and yield 3-tuples
# (stdin, stdout, stderr):
>>> res = container.exec_run(cmd, stream=True, demux=True)
>>> next(res.output)
(None, b'hello stdout\n', None)
>>> next(res.output)
(None, None, b'hello stderr\n')
>>> next(res.output)
StopIteration

# Finally, with stream=False and demux=True, the whole output
# is returned, but the streams are still separated:
>>> res = container.exec_run(cmd, stream=True, demux=True)
>>> next(res.output)
(None, b'hello stdout\n', None)
>>> next(res.output)
(None, None, b'hello stderr\n')
>>> next(res.output)
StopIteration

```

fixes https://github.com/docker/docker-py/issues/1952


little-dude:
Just noticed that this breaks the `tty=True` case:

```python
>>> res = container.exec_run(cmd, stream=True, demux=False, tty=True)
>>> next(res.output)                                                 
b'dout!\r\nhello stderr!\r\n'                                        
```

exactly 8 bytes missing.

docker/utils/socket.py
```diff
@@ -84,17 +84,4 @@ def frames_iter(socket):
                 # We have reached EOF
                 return
             n -= data_length
-            yield result
-
-
-def socket_raw_iter(socket):
```
little-dude: doh. I was wondering what was the difference with `frames_iter`. Now that I broken `tty=True` is broken I understand why it's there... Will back out that change

little-dude: Fixed:


```python
In [6]: res = container.exec_run(cmd, stream=True, demux=False, tty=True) 
                                                                          
In [7]: next(res.output)                                                  
Out[7]: b'hello stdout\r\nhello stderr\r\n'                               
                                                                          
In [10]: res = container.exec_run(cmd, stream=True, demux=True, tty=True)         
                                                                                  
In [11]: next(res.output)                                                         
Out[11]: (None, b'hello stdout\r\nhello stderr\r\n', None)                        
```

little-dude:
@shin- do you mind taking a quick look at this? I'll have some time this week to fix the tests (and add some for this feature). If you have any comment I'll handle them as well.

little-dude:
@shin- any chance of getting this reviewed? It's quite demoralizing to not get any feedback...

shin-:
Sorry - it looked like you still had test failures, but they're due to https://github.com/moby/moby/issues/37920

Can you rebase against the current master to pick up a3111d9e00c15f957af19f5bb5e301cc606f9aeb and clear the CI failures?

little-dude:
Thank you! Rebased. I'm not on my dev machine right now though, so I haven't re-run the tests locally. Let's see what CI has to say about this.

little-dude:
@shin- CI passed. What do you think? As I said I can add additional tests, but I'd like to know you like the implementation first to avoid extra work.

shin-:
@little-dude Thanks!

Design-wise, I don't think there's ever a case where the data we receive from the server is in the `STDIN` channel. If we could remove that and go from a 3-tuple to a 2-tuple on `demux`, I think it would make the experience better overall.

docker/api/client.py
```diff
@@ -381,19 +386,58 @@ def _stream_raw_result(self, response, chunk_size=1, decode=True):
...
+    def _read_from_socket(self, response, stream, tty=False, demux=False):
         socket = self._get_raw_response_socket(response)
 
-        gen = None
-        if tty is False:
+        if tty:
+            # When `tty=True`, there's no framing, we just read the raw data
+            # from the socket. It is as if the streams were multiplexed and
+            # spit out on stdout.
+            gen = ((STDOUT, data) for data in socket_raw_iter(socket))
```
shin-: Since this is the only place `socket_raw_iter` is used, maybe we can just make that change inside the function instead?

docker/utils/socket.py
```diff
@@ -53,26 +53,26 @@ def read_exactly(socket, n):
 
 def next_frame_size(socket):
```
shin-: Should we rename this to `next_frame_header` or similar?

docker/api/client.py
```diff
@@ -381,19 +386,58 @@ def _stream_raw_result(self, response, chunk_size=1, decode=True):
...
 
         if stream:
             return gen
         else:
-            return six.binary_type().join(gen)
+            # Wait for all the frames, concatenate them, and return the result
+            return self._consume_socket_output(gen, demux=demux)
+
+    @staticmethod
+    def _consume_socket_output(frames, demux=False):
```
shin-: This method as well as `_demux_adaptor` should probably be in `docker/utils/socket.py`

docker/models/containers.py
```diff
@@ -175,14 +176,79 @@ def exec_run(self, cmd, stdout=True, stderr=True, stdin=False, tty=False,
         Raises:
             :py:class:`docker.errors.APIError`
                 If the server returns an error.
+
+        Example:
```
shin-: Not sure this is the best place for such a long example, especially since it applies to `attach()` as well.

docker/api/client.py
little-dude: done

docker/api/client.py
little-dude: done. I got rid of `socket_raw_iter()`, and added a `tty` argument to `iter_frames()`.

docker/models/containers.py
little-dude: I'm not sure where to put it though. I can just copy paste if for `attach()`, or add a reference to it.

little-dude:
Rebased on top of https://github.com/docker/docker-py/pull/2178 for now to prevent some pytest failures.

shin-:
Looks like Jenkins caught a bit of a cold over the weekend! I kicked the build again, but everything looks solid now.

docker/models/containers.py
shin-: Maybe as a separate docs page. We can figure that out later though, it's not worth blocking the PR.

docker/utils/socket.py
```diff
@@ -84,17 +103,59 @@ def frames_iter(socket):
...
+            demultiplexed, and the result is a 3-tuple where each item is the
+            concatenation of frames belonging to the same stream.
+    """
+    if demux is False:
+        # If the streams are multiplexed, the generator returns strings, that
+        # we just need to concatenate.
+        return six.binary_type().join(frames)
+
+    # If the streams are demultiplexed, the generator returns tuples
+    # (stdin, stdout, stderr)
```
shin-: nit: now just `(stdout, stderr)`

docker/utils/socket.py
```diff
@@ -84,17 +103,59 @@ def frames_iter(socket):
...
+
+
+def consume_socket_output(frames, demux=False):
+    """
+    Iterate through frames read from the socket and return the result.
+
+    Args:
+
+        demux (bool):
+            If False, stdin, stdout and stderr are multiplexed, and the result
```
shin-: -stdin

docker/utils/socket.py
```diff
@@ -84,17 +103,59 @@ def frames_iter(socket):
...
+            is the concatenation of all the frames. If True, the streams are
+            demultiplexed, and the result is a 3-tuple where each item is the
```
shin-: 2-tuple*

little-dude:
Rebased, comments addresses and squashed. I added a bunch of unit tests (and found a bug!). I'll look at the integration tests to see if there's something I can easily add there.

little-dude:
I found a couple places where I could add tests. I'll update the PR in a few hours.

little-dude:
@shin- I think this is ready (provided that CI passes).

shin-:
Thanks! I'll give it another round of review.

tests/integration/api_exec_test.py
```diff
@@ -75,6 +75,75 @@ def test_exec_command_streaming(self):
...
+        exec_log = self.client.exec_start(res, stream=True)
+        assert next(exec_log) == b'hello out\r\n'
+        assert next(exec_log) == b'hello err\r\n'
+        with self.assertRaises(StopIteration):
+            next(exec_log)
+
+        # tty=True, stream=False, demux=True
+        res = self.client.exec_create(id, cmd, tty=True)
+        exec_log = self.client.exec_start(res, demux=True)
+        assert exec_log == (b'hello out\r\nhello err\r\n', b'')
```
little-dude: This is a bit unfortunate. To be consistent, this should probably be:

```python
assert exec_log == (b'hello out\r\nhello err\r\n', None)
```

But on the other hand, `tty=True` and `demux=True` does not really make sense _anyway_ so I doubt it's going to be an issue in practice.

docker/api/exec_api.py
```diff
@@ -130,11 +130,14 @@ def exec_start(self, exec_id, detach=False, tty=False, stream=False,
             stream (bool): Stream response data. Default: False
             socket (bool): Return the connection socket to allow custom
                 read/write operations.
+            demux (bool): Separate return stdin, stdout and stderr separately
```
little-dude: Note to self: remove stdin

docker/api/exec_api.py
```diff
@@ -130,11 +130,14 @@ def exec_start(self, exec_id, detach=False, tty=False, stream=False,
...
 
         Returns:
-            (generator or str): If ``stream=True``, a generator yielding
-            response chunks. If ``socket=True``, a socket object for the
-            connection. A string containing response data otherwise.
+
+            (generator or str or tuple): If ``stream=True``, a generator
+            yielding response chunks. If ``socket=True``, a socket object for
+            the connection. A string containing response data otherwise. If
+            ``demux=True``, stdin, stdout and stderr are separated.
```
little-dude: remove stdin

tests/integration/api_exec_test.py
shin-: Should we raise an exception if someone tries to use `demux` and `tty` together?
Like https://github.com/docker/docker-py/blob/master/docker/api/container.py#L1101-L1104 but in reverse.

little-dude: We could but strictly speaking I don't think it's really _invalid_. I don't see a use case for it, but I'd like not to restrict the API. I propose we fix the consistency issue instead. Here is a possible fix: https://github.com/docker/docker-py/pull/2150/commits/f80eecc0a3dc0be580b1a48259938ca98de1598d

shin-:
Merged - thank you for your help and your patience!

little-dude:
Thanks a lot for your reviews @shin- !

PR #2179, opened 2018-11-22
Log everything onto a single logger

adaszko:
This is to ease configuration of the logger in the applications code
that use the library.  So instead of having to configure all the loggers
corresponding to various `__name__`s, it is enough to configure just one,
named `docker-py`.

### DIFF
--- a/docker/api/build.py
+++ b/docker/api/build.py
@@ -9,7 +9,7 @@ from .. import errors
 from .. import utils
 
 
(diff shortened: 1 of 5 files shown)


ulyssessouza:
This PR is not passing due to an already solved problem.

adaszko:
Pushed but something in Jenkins is still failing.

PR #2201, opened 2018-12-12
Added arguments to create a swarm with a custom address pool and subnet size.

originalgremlin:
Signed-off-by: Barry Shapira <barry@whiterabbit.ai>


originalgremlin:
Adding the arguments to swarm init was relatively straightforward.  I do wonder how the maintainers will feel about adding these arguments at all: all other parts of the docker-py swarm API rely on a docker engine API version of 1.24, whereas these features are from a much more modern docker engine API of 1.39.

If any maintainers have ideas on how best to handle modernization of the swarm part of this repo I would be happy to help round out the codebase and truly "let you do anything the docker command does".

shin-:
Thanks for submitting a PR @originalgremlin !

FWIW, the library already has a fairly established way to handle parameters added in newer versions of the API. see [here](https://github.com/docker/docker-py/blob/master/docker/api/build.py#L222-L244) or [here](https://github.com/docker/docker-py/blob/master/docker/api/secret.py#L36-L42) for example.  Please make sure to adopt the same idiom for this PR so we can move forward!

originalgremlin:
@shin- That makes sense.  Thanks for pointing out the examples for how such things are handled.

Code updated and pushed.  Has been tested a fair amount in my local environment.  Did my best to stick with the project style though you may well have suggestions for things that I missed.

originalgremlin:
@shin- Should be ready to go now.  All tests pass and all protocol followed to the best of my ability.  Let me know if there's any other polishing I should do.

originalgremlin:
@shin- Squeaky wheel ping.  Anything missing or needing improvement in order to get this PR merged?

docker/api/swarm.py
```diff
@@ -82,6 +83,7 @@ def get_unlock_key(self):
 
     @utils.minimum_version('1.24')
     def init_swarm(self, advertise_addr=None, listen_addr='0.0.0.0:2377',
+                   default_addr_pool=None, subnet_size=None,
```
chris-crone: Should add new parameters at the end of the list so that callers of `init_swarm` that use ordered (i.e.: unnamed) parameters continue to work as expected.

e.g.: An existing function could call:
```python
s.init_swarm('192.168.0.1', '0.0.0.0:2377', True)
```

Which would now break.

docker/models/swarm.py
```diff
@@ -34,6 +34,7 @@ def get_unlock_key(self):
     get_unlock_key.__doc__ = APIClient.get_unlock_key.__doc__
 
     def init(self, advertise_addr=None, listen_addr='0.0.0.0:2377',
+             default_addr_pool=None, subnet_size=None,
```
chris-crone: Same comment here for parameters.

docker/api/swarm.py
```diff
@@ -102,6 +104,12 @@ def init_swarm(self, advertise_addr=None, listen_addr='0.0.0.0:2377',
                 or an interface followed by a port number, like ``eth0:4567``.
                 If the port number is omitted, the default swarm listening port
                 is used. Default: '0.0.0.0:2377'
+            default_addr_pool (list of strings): Default Address Pool specifies
+                default subnet pools for global scope networks. Each pool
+                should be specified as a CIDR block, like '10.0.0.0/16'.
```
chris-crone: nit: Might be good to use the default as the example.

Issue #2025, opened 2018-05-09
Couldn't create secret object


ddowding:
I couldn't create secret object, the problem seemed to boil down to the way that a secret was being created from the docker daemon response. 

https://github.com/docker/docker-py/blob/467cacb00d8dce68aa8ff2bdacc85acecd2d1207/docker/models/secrets.py#L31-L33

Docker version 18.03.1-ce and python version 3.5 had the following error:

````
File "docker/models/secrets.py", line 10 in __repr__
return "<%s: %s'>" % (self.__class__.__name__, self.name)
File "docker/models/secrets.py", line 14 in name
return self.attrs['Spec']['Name']
KeyError: 'Spec'

````

When calling: 

````
import docker

client -docker.from_env()
mySecret = client.secrets.create(name='randomName', data='platform_node_requirements.md')

````

Changing the code to the following seemed to fix it. 
````
obj = self.client.api.create_secret(**kwargs)
secret = self.client.secrets.get(obj.get('ID'))
return self.prepare_model(secret)
````

**Comment by douglasmiranda** [comment]
Master branch is working.

**Comment by remorses** [comment]
I get the same error for `client.configs.create`

**Comment by etienne-napoleone** [comment]
docker version: `19.03.5 633a0ea`
docker-py version: `4.1.0`

On my side the creation works but the `__repr__` and `__str__` methods seem to expect a name where the returned object only contains the id attr. Not too bad but it could be handled more gracefully

```python
>>> import docker
>>> d = docker.from_env()
>>> s = d.secrets.create(name="test", data="ok")
>>> s
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/user/Library/Caches/pypoetry/virtualenvs/whatever-py3.7/lib/python3.7/site-packages/docker/models/secrets.py", line 10, in __repr__
    return "<%s: '%s'>" % (self.__class__.__name__, self.name)
  File "/Users/user/Library/Caches/pypoetry/virtualenvs/whatever-py3.7/lib/python3.7/site-packages/docker/models/secrets.py", line 14, in name
    return self.attrs['Spec']['Name']
KeyError: 'Spec'
>>> s.attrs
{'ID': 'wi2jpy7n70wkkrbytnbctv1nj'}
```

feliperuhland:
Hi @etienne-napoleone 

I opened a [PR](https://github.com/docker/docker-py/pull/2470) for that issue. Do you want to review it?
Thanks.

PR #2211, opened 2018-12-24
Update configs.py to create config from id.

johnnydowns:
Same problem then in issue #2025 but with configs


redNixon:
@GordonTheTurtle Is there anything left that is preventing it from being merged?  I have not had any issues with it so far fwiiw.

feliperuhland:
@johnnydowns @redNixon 

That issue was solved by #2793

@ulyssessouza we can close this PR because was fixed by #2793.

Thanks and have a good day :)

PR #2216, opened 2019-01-08
Regression 443 test: relax status-code check

thaJeztah:
This test was testing for a 500 status, but this status is actually a bug in the API (as it's due to an invalid request), and the API should actually return a 400 status.

To make this test handle both situations, relax the test to accept either a 4xx or 5xx status.

relates to https://github.com/moby/moby/pull/38487#issuecomment-452484199

@shin- @LinuxMercedes

### DIFF
--- a/tests/integration/regression_test.py
+++ b/tests/integration/regression_test.py
@@ -14,7 +14,7 @@ class TestRegressions(BaseAPIIntegrationTest):
         with pytest.raises(docker.errors.APIError) as exc:
             for line in self.client.build(fileobj=dfile, tag="a/b/c"):
                 pass
-        assert exc.value.response.status_code == 500
+        assert exc.value.is_error()
         dfile.close()
 
     def test_542_truncate_ids_client_side(self):
--- a/docker/errors.py
+++ b/docker/errors.py
@@ -63,6 +63,9 @@ class APIError(requests.exceptions.HTTPError, DockerException):
         if self.response is not None:
             return self.response.status_code
 
+    def is_error(self):
+        return self.is_client_error() or self.is_server_error()
+
     def is_client_error(self):
         if self.status_code is None:
             return False
--- a/tests/unit/errors_test.py
+++ b/tests/unit/errors_test.py
@@ -79,6 +79,27 @@ class APIErrorTest(unittest.TestCase):
         err = APIError('', response=resp)
         assert err.is_client_error() is True
 
+    def test_is_error_300(self):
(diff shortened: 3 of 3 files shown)


tests/integration/regression_test.py
```diff
@@ -14,7 +14,7 @@ def test_443_handle_nonchunked_response_in_stream(self):
         with pytest.raises(docker.errors.APIError) as exc:
             for line in self.client.build(fileobj=dfile, tag="a/b/c"):
                 pass
-        assert exc.value.response.status_code == 500
+        assert exc.value.response.is_error
```
shin-: should be a func call()

shin-: specifically I think this should be `assert exc.value.is_error()`

thaJeztah: ah, dang; let me update 😅 - I'm really bad at python

thaJeztah: Updated; also happy to remove the new `is_error` if you think we should just inline it here

tests/integration/regression_test.py
```diff
@@ -14,7 +14,7 @@ def test_443_handle_nonchunked_response_in_stream(self):
         with pytest.raises(docker.errors.APIError) as exc:
             for line in self.client.build(fileobj=dfile, tag="a/b/c"):
                 pass
-        assert exc.value.response.status_code == 500
+        assert exc.value.response.is_error()
```
shin-: Removing `.response` should clear up the CI failure

shin-:
Failure on 2.7 is unrelated to this PR and probably needs some dependency wrangling, I'll have to take a closer look.

tests/integration/regression_test.py
thaJeztah: 🤦‍♂️ updated again; thanks for bearing with me 🤗

thaJeztah:
Is this expected? (https://jenkins.dockerproject.org/job/docker/job/docker-py/job/PR-2216/4/execution/node/25/log/)

```
20:23:39 [docker_docker-py_PR-2216-VUUIZLMALJLRQO2LIHC36PEUWVSMCGAOFNW2SQKYNEQKZ7MZMHEA] Running shell script
20:23:40 + docker pull dockerbuildbot/docker-py:py2-219c52141e3cd15db3348c5420220f640323499f
20:23:41 Error response from daemon: manifest for dockerbuildbot/docker-py:py2-219c52141e3cd15db3348c5420220f640323499f not found
```

PR #2228, opened 2019-01-12
Accounting for asynchronous service removals

wk8:
Swarmkit services' removal will soon become asynchronous (see
https://github.com/moby/moby/pull/38525); that is, a service will only actually
get deleted once all of its containers are shut down.

This patch changes service tests' `tearDown` to wait for services to be
actually removed, so that there's no name collision with subsequent tests.

This change is fully backward compatible with the engine's current behaviour.

Couple of other misc changes: consolidated polling in a single
`wait_until_truthy` function, and allowed passing more flags to `py.test` in
the makefile test-integration targets through a new optional `pytest_options`
env variable.

Signed-off-by: Jean Rouge <jean.rouge@docker.com>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -35,11 +35,11 @@ unit-test-py3: build-py3
 
 .PHONY: integration-test
 integration-test: build
-	docker run -t --rm -v /var/run/docker.sock:/var/run/docker.sock docker-sdk-python py.test -v tests/integration/${file}
+	docker run -t --rm -v /var/run/docker.sock:/var/run/docker.sock docker-sdk-python py.test -v tests/integration/${file} ${pytest_options}
 
 .PHONY: integration-test-py3
 integration-test-py3: build-py3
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -35,11 +35,11 @@ unit-test-py3: build-py3
 
 .PHONY: integration-test
 integration-test: build
-	docker run -t --rm -v /var/run/docker.sock:/var/run/docker.sock docker-sdk-python py.test -v tests/integration/${file}
+	docker run -t --rm -v /var/run/docker.sock:/var/run/docker.sock docker-sdk-python py.test -v tests/integration/${file} ${pytest_options}
```
wk8: Comes in handy to just run a test or two using `-k` (https://docs.pytest.org/en/latest/example/markers.html#using-k-expr-to-select-tests-based-on-their-name)

wk8:
Stale, closing

PR #2240, opened 2019-01-23
chore: fix return type docs for `container.logs()`

BobDu:
In container logs , return is bytes or generator iterator of output bytes , not string , the comments is error.


docker/api/container.py
```diff
@@ -804,7 +804,7 @@ def logs(self, container, stdout=True, stderr=True, stream=False,
                 datetime or integer epoch (in seconds)
 
         Returns:
-            (generator or str)
+            (generator iterator of output bytes or bytes)
```
ulyssessouza: I think that you are mixing 2 concerns when using the term `generator iterator`. They are 2 different things. Just using `generator` should be fine.
My proposal would be:

`(generator of bytes or bytes)`

docker/models/containers.py
```diff
@@ -285,7 +285,8 @@ def logs(self, **kwargs):
                 datetime or integer epoch (in seconds)
 
         Returns:
-            (generator or str): Logs from the container.
+            (generatoran iterator of output bytes or bytes):
```
ulyssessouza: Same as a above. And you had a little typo here `generatoran` -> `generator`

Bob-Du: Thanks for your review and comment, I will change it.

milas (approved):

PR #2252, opened 2019-02-09
Improve error message in case of paramiko failure

ssbarnea:
Fix issues where essential error about paramiko dealing code was
lost due to ageneric exception that assumed paramiko library was
missing.

To keep backwards compatibility we keep the custom exception but we
also include information from original exception that caused it.

It was wrong to believe that only missing paramiko can cause it, even
code inside docker-py can thow exceptions, and NamedException is very
broad exception class.

### DIFF
--- a/docker/api/client.py
+++ b/docker/api/client.py
@@ -162,7 +162,7 @@ class APIClient(
                     base_url, timeout, pool_connections=num_pools
                 )
             except NameError as e:
-                # do not hide orrinal exception as it may contain variou
+                # do not hide orriginal exception as it may contain important info
                 raise DockerException(
                     "Paramiko failure while trying ssh:// protocol:\n%s" % e
                 )
--- a/docker/api/client.py
+++ b/docker/api/client.py
@@ -161,9 +161,10 @@ class APIClient(
                 self._custom_adapter = SSHAdapter(
                     base_url, timeout, pool_connections=num_pools
                 )
-            except NameError:
+            except NameError as e:
+                # do not hide orrinal exception as it may contain variou
                 raise DockerException(
-                    'Install paramiko package to enable ssh:// support'
+                    "Paramiko failure while trying ssh:// protocol:\n%s" % e
                 )
             self.mount('http+docker://ssh', self._custom_adapter)
             self._unmount('http://', 'https://')


docker/api/client.py
```diff
@@ -161,9 +161,10 @@ def __init__(self, base_url=None, version=None,
                 self._custom_adapter = SSHAdapter(
                     base_url, timeout, pool_connections=num_pools
                 )
-            except NameError:
+            except NameError as e:
+                # do not hide orrinal exception as it may contain variou
```
TomasTomecek: please update the comment, seems to be cut

docker/api/client.py
```diff
@@ -161,9 +161,10 @@ def __init__(self, base_url=None, version=None,
                 self._custom_adapter = SSHAdapter(
                     base_url, timeout, pool_connections=num_pools
                 )
-            except NameError:
+            except NameError as e:
+                # do not hide orriginal exception as it may contain important info
```
ulyssessouza: typo: orriginal -> original

docker/api/client.py
```diff
@@ -161,9 +161,10 @@ def __init__(self, base_url=None, version=None,
...
                 raise DockerException(
-                    'Install paramiko package to enable ssh:// support'
+                    "Paramiko failure while trying ssh:// protocol:\n%s" % e
```
ulyssessouza: Not sure if changing the message is necessary here.

shin-:
@ssbarnea Do you have an example of a `NameError` being raised on that particular code path that isn't caused by a missing dependency? 

My problem with this PR is that I would still like to give clear information to the user in the case they haven't installed the optional `paramiko` dependency. As such, a better fix may be to catch a more specific error, or to catch it earlier in the callstack to avoid false positives. Making the error message more generic and lower-level is not the right direction, IMO.

ssbarnea:
@shin- Yes I encountered that but I do not remember it now. I would rather throw the original error back instead of creating a new one that is crippled of important information. That code is wrong as it makes some dangereus assumptions.

ulyssessouza:
This PR is not passing due to an already [solved problem](https://github.com/docker/docker-py/pull/2307).

PR #2259, opened 2019-02-21
Fix `network_mode` API documentation wording

p1100i:


jcsirot:
@p1100i thank you for your contribution

Issue #2248, opened 2019-01-31
Cannot use SSH adapter after close


rumpl:
When connecting over SSH it is not possible to do other calls after the connection is closed. However we can still make API calls when connecting on a socket, sample code:

```python
import docker 

client = docker.APIClient(base_url='unix://var/run/docker.sock')
print(client.version()['Version'])
client.close()
print(client.version()['Version'])


ssh_client = docker.APIClient(base_url='ssh://...')
print(ssh_client.version()['Version'])
ssh_client.close()
print(ssh_client.version()['Version'])
```

The output is:
```
18.09.1
18.09.1
18.09.1
Traceback (most recent call last):
  File "main.py", line 16, in <module>
    main()
  File "main.py", line 13, in main
    print(ssh_client.version()['Version'])
  File "/Users/djordjelukic/.virtualenvs/test-url/lib/python2.7/site-packages/docker/api/daemon.py", line 181, in version
    return self._result(self._get(url), json=True)
  File "/Users/djordjelukic/.virtualenvs/test-url/lib/python2.7/site-packages/docker/utils/decorators.py", line 46, in inner
    return f(self, *args, **kwargs)
  File "/Users/djordjelukic/.virtualenvs/test-url/lib/python2.7/site-packages/docker/api/client.py", line 225, in _get
    return self.get(url, **self._set_request_timeout(kwargs))
  File "/Users/djordjelukic/.virtualenvs/test-url/lib/python2.7/site-packages/requests/sessions.py", line 546, in get
    return self.request('GET', url, **kwargs)
  File "/Users/djordjelukic/.virtualenvs/test-url/lib/python2.7/site-packages/requests/sessions.py", line 533, in request
    resp = self.send(prep, **send_kwargs)
  File "/Users/djordjelukic/.virtualenvs/test-url/lib/python2.7/site-packages/requests/sessions.py", line 646, in send
    r = adapter.send(request, **kwargs)
  File "/Users/djordjelukic/.virtualenvs/test-url/lib/python2.7/site-packages/requests/adapters.py", line 449, in send
    timeout=timeout
  File "/Users/djordjelukic/.virtualenvs/test-url/lib/python2.7/site-packages/urllib3/connectionpool.py", line 600, in urlopen
    chunked=chunked)
  File "/Users/djordjelukic/.virtualenvs/test-url/lib/python2.7/site-packages/urllib3/connectionpool.py", line 354, in _make_request
    conn.request(method, url, **httplib_request_kw)
[...]

PR #2262, opened 2019-02-24
On SSHAdapter, close channel after getting the response (#2248)

arcenik:
This is a fix proposal for #2248.

It adds a build_response function on the SSHAdapter object to close the ssh channel oncy the response is received.

Is there a more elegant way to make requests do it ?

The versions are
* python-docker 3.7.0-1
* docker 18.09.2-ce (on Archlinux)
* Python 3.7.2


jcsirot:
@arcenik Thank you for your PR. However we are currently working on a patch globally improving the SSH support, including a fix for #2248

PR #2265, opened 2019-03-01
set buildargs default value if None

lekster:
fix "AttributeError: 'NoneType' object has no attribute 'setdefault'" when buildargs is None

### DIFF
--- a/docker/api/build.py
+++ b/docker/api/build.py
@@ -121,6 +121,7 @@ class BuildApiMixin(object):
         remote = context = None
         headers = {}
         container_limits = container_limits or {}
+        buildargs = buildargs or {}
         if path is None and fileobj is None:
             raise TypeError("Either path or fileobj needs to be provided.")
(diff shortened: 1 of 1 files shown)


docker/api/build.py
```diff
@@ -121,6 +121,7 @@ def build(self, path=None, tag=None, quiet=False, fileobj=None,
         remote = context = None
         headers = {}
         container_limits = container_limits or {}
+        buildargs = buildargs or {}
```
ulyssessouza: Any special reason not to use the default parameter value in line 20?
Looks like a `buildargs={}` in line 20 would be cleaner.

shin-: @ulyssessouza https://docs.quantifiedcode.com/python-anti-patterns/correctness/mutable_default_value_as_argument.html

ulyssessouza: Ahhh, OK!

PR #2272, opened 2019-03-13
Return API response on service update

hannseman:
The response contains a `Warnings` key that should be returned so it can be handled by the caller. 

See the 200 Response documentation here: https://docs.docker.com/engine/api/v1.39/#operation/ServiceUpdate

Currently the only return value from this method is `True`. It can never return `False` (it will raise instead). 

With this change checks like this on the return value will still work:
```python
response = client.update_service(...)
if response:
  ...
```

But it would break backwards compatability if used like this:
```python
response = client.update_service(...)
if response is True:
  ...
```

As previously stated both these checks does not make sense as the only possible return value is `True`.

### DIFF
--- a/docker/api/service.py
+++ b/docker/api/service.py
@@ -387,7 +387,7 @@ class ServiceApiMixin(object):
                 current specification of the service. Default: ``False``
 
         Returns:
-            ``True`` if successful.
+            A dictionary containing a ``Warnings`` key.
 
         Raises:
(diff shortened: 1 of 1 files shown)


hannseman:
> We can ship the change with 4.0, and the API change won't be an issue in that case.

Sounds good!

@shin- and @ulyssessouza thanks for the review and merge.

PR #2273, opened 2019-03-13
Rename `container_id` to `container_name_or_id`.

heckad:
Removes the misunderstanding caused by the difference between the variable name and its meaning.

### DIFF
--- a/docker/models/containers.py
+++ b/docker/models/containers.py
@@ -298,7 +298,7 @@ class Container(Model):
         Pauses all processes within this container.
 
         Raises:
-            :py:class:`docker.errors.APIError`
+            :py:class:`docker.errors.APIError`def get(self, container_id):
                 If the server returns an error.
         """
         return self.client.api.pause(self.id)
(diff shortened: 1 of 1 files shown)


shin-:
Thank you for your contribution! That said, I fail to see the upside of that breaking change, especially if applied to a single method, when many of the method signatures in our codebase use the same shorthand.

heckad:
A quick review showed that in other places it will be exactly `container_id`, but here `container_id` can also be a name. Perhaps need to rename docstring on 301 line in `containers.py`

heckad:
So, merge or close? If we merge, it will be better to read.

ulyssessouza:
Hello @heckad ! Thank you for the PR.

I agree with you that `container_name_or_id` would make it more explicit and if that wasn't a breaking change, I would approve but since it's not the case and the upside is minor, I have to decline.

And we still have the documentation to make it explicit.

heckad:
What does it break?

shin-:
It breaks the code of anyone using named parameters when calling this method.

heckad:
Just out of curiosity, why do so?

shin-:
It generally makes the code more explicit and better at self-documenting.

PR #2281, opened 2019-03-18
Support for SCTP

### LINKED ISSUE #2249: Allow exposing SCTP ports
Exposing SCTP ports has been supported in docker since 18.03 but docker-py currently does not support it but instead raises a ValueError.
Because of this compose doesn't support SCTP either, see the original issue on docker/compose (https://github.com/docker/compose/issues/6294) for more info.

I think the solution could simply be to add 'sctp' to PORT_SPEC here :
https://github.com/docker/docker-py/blob/61bab6379890841f698ed820e464bc00bf2d379d/docker/utils/ports.py#L10

Sadly I can't deliver a proper PR for this.

hannseman:
Fixes #2249 by adding `sctp` as an allowed protocol when exposing ports.

Also extends some port related tests to cover protocols.

### DIFF
--- a/docker/api/container.py
+++ b/docker/api/container.py
@@ -915,9 +915,10 @@ class ContainerApiMixin(object):
         if '/' in private_port:
             return port_settings.get(private_port)
 
-        h_ports = port_settings.get(private_port + '/tcp')
-        if h_ports is None:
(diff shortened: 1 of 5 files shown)


hannseman:
@ulyssessouza @shin- thank you for the review and merge!

PR #2282, opened 2019-03-19
Avoid race condition on short execution

ulyssessouza:
- Add a sleep of 5 seconds to be sure the logs
can be requested before the daemon removes the
container when run with auto_remove=True

This avoids to fail in line https://github.com/docker/docker-py/blob/master/docker/models/containers.py#L811

Signed-off-by: Ulysses Souza <ulysses.souza@docker.com>


ijc:
You could probably also construct some similar interlock using the network using `nc -l -p 8080`, running the container with `-p ::8080` and then connecting to the port...

Issue #2251, opened 2019-02-06
using ssh protocol gives paramiko.ssh_exception.PasswordRequiredException: Private key file is encrypted


ssbarnea:
When trying to use ssh transport to talk with docker, docker-py chokes with belower error complaining that the private key is encrypted when in reality it should connect succesfully as the key is provided by the ssh agent.

Docker cli on the other hand works as expected.

```
export DOCKER_HOST=ssh://localhost
#  ^ can as well be another host, it does not matter.

Traceback (most recent call last):
File "./py-docker", line 5, in <module>
c = docker.from_env()
File "/Users/ssbarnea/.pyenv/versions/2.7.15/lib/python2.7/site-packages/docker/client.py", line 85, in from_env
timeout=timeout, version=version, **kwargs_from_env(**kwargs)
File "/Users/ssbarnea/.pyenv/versions/2.7.15/lib/python2.7/site-packages/docker/client.py", line 40, in __init__
self.api = APIClient(*args, **kwargs)
File "/Users/ssbarnea/.pyenv/versions/2.7.15/lib/python2.7/site-packages/docker/api/client.py", line 162, in __init__
base_url, timeout, pool_connections=num_pools
File "/Users/ssbarnea/.pyenv/versions/2.7.15/lib/python2.7/site-packages/docker/transport/sshconn.py", line 84, in __init__
parsed.hostname, parsed.port, parsed.username,
File "/Users/ssbarnea/.pyenv/versions/2.7.15/lib/python2.7/site-packages/paramiko/client.py", line 437, in connect
passphrase,
File "/Users/ssbarnea/.pyenv/versions/2.7.15/lib/python2.7/site-packages/paramiko/client.py", line 749, in _auth
raise saved_exception
paramiko.ssh_exception.PasswordRequiredException: Private key file is encrypted
```

This is caused by the fact that current implementation assumes that if `~/.ssh/id_rsa` exists it should be an unencrypted key (which would be something really stupid from securitu point of view).

I can confirm that the ssh agent is working correctly on the machine and that ssh and docker-cli do have not problem accessing the ssh keys from the agent.
```
ssh-add -l
```

ssbarnea:
It proved to be caused by paramiko https://github.com/paramiko/paramiko/issues/1383 -- in case someone else encounters it we could maybe add a runtime warning if the ssh:// url does not contain a user. Missing user does not mean that `~/.ssh/config` one will be used, just mean current user.

[...]

PR #2290, opened 2019-03-24
obey .ssh/config and set logging in paramiko to warn

riedel:
As #2251 mentions currently the .ssh/config is ignored. This adds basic support. Particularly also for proxy_commands (no jump hosts yet) but also for setting ssh username, hostname and port in the config.

I tested this with my docker-compose under cygwin and it seems to work fine. I am actually not sure about Windows related to the ssh config file. However, I guess the code should still cause no problems beyond the ones already existing.

Furthermore I turned off the annoying default logging settings...

### DIFF
--- a/docker/transport/sshconn.py
+++ b/docker/transport/sshconn.py
@@ -2,6 +2,7 @@ import paramiko
 import requests.adapters
 import six
 import logging
+import os
 
 from docker.transport.basehttpadapter import BaseHTTPAdapter
 from .. import constants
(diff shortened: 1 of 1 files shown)


riedel:
forgot to mention in the commit that I set a default host key policy as well

docker/transport/sshconn.py
```diff
@@ -72,13 +74,25 @@ def _get_conn(self, timeout):
...
+        if os.path.exists(ssh_config_file):
+            conf = paramiko.SSHConfig()
+            with open(ssh_config_file) as f:
+                conf.parse(f)
+            parsed = six.moves.urllib_parse.urlparse(base_url)
+            host_config = conf.lookup(parsed.hostname)
+            self.ssh_conf = host_config
+
         self.ssh_client.load_system_host_keys()
+        self.ssh_client.set_missing_host_key_policy(paramiko.AutoAddPolicy())
```
ijc: If we aren't going to obey the various config options which effect this (`StrictHostKeyChecking` et al) then I think `WarningPolicy()` would be a better one to hard code.

docker/transport/sshconn.py
```diff
@@ -90,8 +104,22 @@ def __init__(self, base_url, timeout=60,
 
     def _connect(self):
         parsed = six.moves.urllib_parse.urlparse(self.base_url)
```
ijc: With the addition above we are not now parsing the URL twice? We should parse it exactly once. That suggests either the code here should be in `__init__` or the code from `__init__` should be here.

riedel: I tried to do rather minimal refactoring. I changed this now consistently.

After looking at the code I however wonder, why a connection is created at all in the init code and why base_url is passed to the constructor...

docker/transport/sshconn.py
riedel: I changed this in 0e2fe7f

riedel:
I tried to address everything except for testing. 
For this I would suggest:
1. passing an optional constructor argument for specifying a .ssh/config file (equal to ssh -F)
1. removing _connect from init and rather connect on first connection

Then one could at least test if some overwrites work as expected.

maiermic:
I just stumbled on this issue. What is the status? What actions are required? @riedel made suggestions in his last comment. Do you expect someone to approve this? What are we waiting for?

riedel:
it was actually merged in #2448  it seems 👍

PR #2292, opened 2019-03-26
Fix base_url to keep TCP protocol

ulyssessouza:
This fix lets the responsability of changing the
protocol to `parse_host` afterwards, letting
`base_url` with the original value.

This issue causes https://github.com/docker/compose/issues/6551


tests/integration/api_container_test.py
```diff
@@ -1080,7 +1080,6 @@ def test_kill_with_signal_integer(self):
 
 class PortTest(BaseAPIIntegrationTest):
     def test_port(self):
-
         port_bindings = {
```
ulyssessouza: That was an autoformat from my IDE. I can remove if you would like

PR #2295, opened 2019-03-28
Add support for rotate_manager_unlock_key

hannseman:
This PR adds support for rotating the manager unlock key on swarm update.

See https://docs.docker.com/engine/api/v1.39/#operation/SwarmUpdate

### DIFF
--- a/docker/api/swarm.py
+++ b/docker/api/swarm.py
@@ -406,8 +406,10 @@ class SwarmApiMixin(object):
         return True
 
     @utils.minimum_version('1.24')
-    def update_swarm(self, version, swarm_spec=None, rotate_worker_token=False,
-                     rotate_manager_token=False):
+    def update_swarm(self, version, swarm_spec=None,
+                     rotate_worker_token=False,
+                     rotate_manager_token=False,
+                     rotate_manager_unlock_key=False):
         """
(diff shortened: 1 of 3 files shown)


docker/api/swarm.py
```diff
@@ -407,11 +411,18 @@ def update_swarm(self, version, swarm_spec=None, rotate_worker_token=False,
...
+                raise errors.InvalidVersion(
+                    'Rotate manager unlock key '
+                    'is only available for API version >= 1.25'
+                )
 
         url = self._url('/swarm/update')
         response = self._post_json(url, data=swarm_spec, params={
             'rotateWorkerToken': rotate_worker_token,
             'rotateManagerToken': rotate_manager_token,
+            'rotateManagerUnlockKey': rotate_manager_unlock_key,
```
ijc: This will pass `rotateManagerUnlockkey=false` even to services with api < 1.25. Did you check that this is ok or do we need to completely exclude the key in that case?

hannseman: You're right. Totally missed that. Better to be safe and only include it when set to `True` and >= 1.25.

hannseman:
@ijc @shin- thanks for the review!

PR #2296, opened 2019-03-28
Return node id on swarm init

hannseman:
This PR changes so that the return value of `init_swarm` is not a boolean (which will always be `True` as we raise on HTTP errors). It will now return the Node ID of the created node.

See https://docs.docker.com/engine/api/v1.39/#operation/SwarmInit

It does somewhat break backwards compatibility if used like this:
```python
result = client.init_swarm()
if result is True:
   ...
```

So I guess it makes sense to release this in the next major if merged.

### DIFF
--- a/docker/api/swarm.py
+++ b/docker/api/swarm.py
@@ -117,7 +117,7 @@ class SwarmApiMixin(object):
                 networks created from the default subnet pool. Default: None
 
         Returns:
-            ``True`` if successful.
+            (str): The ID of the created node.
 
         Raises:
             :py:class:`docker.errors.APIError`
@@ -155,8 +155,7 @@ class SwarmApiMixin(object):
             'Spec': swarm_spec,
         }
         response = self._post_json(url, data=data)
-        self._raise_for_status(response)
-        return True
+        return self._result(response, json=True)
 
     @utils.minimum_version('1.24')
     def inspect_swarm(self):
--- a/docker/models/swarm.py
+++ b/docker/models/swarm.py
@@ -96,7 +96,7 @@ class Swarm(Model):
                 created in the orchestrator.
 
         Returns:
-            ``True`` if the request went through.
+            (str): The ID of the created node.
 
         Raises:
             :py:class:`docker.errors.APIError`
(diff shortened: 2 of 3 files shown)


ijc:
Seems ok to me, apart from the lack of test updates.

However, is there an API compat concern here?

Duck typing, I think, means that anyone who is treating this result as a boolean gets the same behaviour even after this change -- since any non-empty string evaluates as True if you treat it as a boolean and we never return the empty string.

I think I need to defer to @shin- on this one.

hannseman:
> Seems ok to me, apart from the lack of test updates.

I can add a test to verify that the returned string is the running node.

> However, is there an API compat concern here?

I don't think so. `/swarm/init` have returned the node id since 1.24:  https://docs.docker.com/engine/api/v1.24/#38-swarm

ijc:
WRT API compat I meant for users of the `docker-py` library API rather than the daemon API.

hannseman:
I've added a test to check that the returned value is the initialised node.

ijc:
Thanks, I'll defer to @shin- on the API compat thing and the right time to merge/release this.

CI seems to have some sort of global issue, unrelated to your change (here or in your other PRs). I'll take a look.

hannseman:
> WRT API compat I meant for users of the docker-py library API rather than the daemon API.

Ah yes. In a previous PR (https://github.com/docker/docker-py/pull/2272) that made a similar change it was decided to be shipped in 4.0.0 to allow breaking backwards compatibility.

ijc:
> CI seems to have some sort of global issue, unrelated to your change (here or in your other PRs). I'll take a look.

This is now fixed in master (see #2298), if you rebase your various PRs then they should now be able to pass CI.

hannseman:
> This is now fixed in master (see #2298), if you rebase your various PRs then they should now be able to pass CI.

Sweet! Done. Looks like Travis CI is having a bad day but hopefully the builds will complete soon enough.

shin-:
Sorry, forgot I was pinged on this. It makes sense to do this, and since the upcoming 4.0 allows us those small compatibility breaks, I think this is all good to go.

hannseman:
@ijc @shin- thanks for the review!

PR #2297, opened 2019-03-28
Add support for setting init on services

### LINKED ISSUE #2293: client.services.create does not support init option
`docker service create` supports an `--init` option since API 1.37.

https://docs.docker.com/engine/reference/commandline/service_create/

This option is not yet available in client.services.create.

hannseman:
This PR adds support for setting the boolean `Init` in `ContainerSpec` when creating and updating services.

I'm not really sure about what API version this feature was actually introduced.

https://docs.docker.com/engine/reference/commandline/service_create/  states that `--init` is supported from 1.37. But when comparing the API spec it looks like it was supported in 1.38.

Check `TaskTemplate.ContainerSpec.Init`: 
https://docs.docker.com/engine/api/v1.38/#operation/ServiceCreate
vs
https://docs.docker.com/engine/api/v1.37/#operation/ServiceCreate

I've set the minimum version to 1.38.

Fixes #2293

### DIFF
--- a/docker/api/service.py
+++ b/docker/api/service.py
@@ -88,6 +88,10 @@ def _check_api_features(version, task_template, update_config, endpoint_spec,
                 if container_spec.get('Isolation') is not None:
                     raise_version_error('ContainerSpec.isolation', '1.35')
 
+            if utils.version_lt(version, '1.38'):
+                if container_spec.get('Init') is not None:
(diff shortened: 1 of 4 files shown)


shin-:
Thanks!

hannseman:
@ijc @shin- thanks for the review! 😄

PR #2300, opened 2019-03-29
implementing the revert suggested by @shin- ages ago

abrugh:
Fixes the issue described, tested and reproduced in https://github.com/docker/compose/issues/6018


ulyssessouza:
This PR is not passing due to an already solved problem.

abrugh:
Rebased, still failing AttachContainerTest::test_attach_stream_and_cancel but only 2 out of 4 times....

shin-:
Looks like it's failing pretty consistently on Python 2.7 (if you look at the test matrix [here](https://jenkins.dockerproject.org/job/docker/job/docker-py/job/PR-2300/)). To me that suggests that that condition does something helpful, and the fix is less trivial than it appears.

PR #2304, opened 2019-03-31
Document correct listen_addr default on join swarm

hannseman:
Currently the default the parameter `listen_addr` for swarm `join` is `None` while it actually is `0.0.0.0:2377`. This PR corrects this.


hannseman:
@ulyssessouza @chris-crone thanks for the review and merge! 😃

PR #2309, opened 2019-04-10
Remove init_path from create

thaJeztah:
This option was never functional, and was not intended to be added to the "container create" API, so let's remove it, because it has been removed in Docker 17.05, and was broken in versions before that; see

- https://github.com/moby/moby/issues/32355 --init-path does not seem to work
- https://github.com/moby/moby/pull/32470 remove --init-path from client

### DIFF
--- a/docker/api/container.py
+++ b/docker/api/container.py
@@ -487,7 +487,6 @@ class ContainerApiMixin(object):
                 IDs that the container process will run as.
             init (bool): Run an init inside the container that forwards
                 signals and reaps processes
-            init_path (str): Path to the docker-init binary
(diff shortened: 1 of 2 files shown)


thaJeztah:
Noticed this was cluttering our CI on every run;

```
00:21:00 XFAIL ../../../../../docker-py/tests/integration/api_container_test.py::CreateContainerTest::test_create_with_init_path
00:21:00   init-path removed in 17.05.0
```

@chris-crone @ulyssessouza

thaJeztah:
I think Jenkins is failing because there's no 19.03.0 image yet (should be `19.03.0-beta1`, `19.03-rc`, `rc`, or `test`)

PR #2312, opened 2019-04-15
adding unless-stopped to the restart_policy option documentation

naorlivne:


naorlivne:
Is everything alright with the Jenkins builds? 

it fails on SwarmTest.test_init_swarm_with_ca_config with an error of ` docker.errors.APIError: 503 Server Error: Service Unavailable ("Swarm is encrypted and needs to be unlocked before it can be used. Please use "docker swarm unlock" to unlock it.")` which has noting to do with the changes of this pull requests (which are all documentation changes)

PR #2314, opened 2019-04-16
Add service rollback method.

tjovicic:
Rollback method was added in version 1.31 and permits rollback to the
previous service spec.

Signed-off-by: Tihomir Jovicic <tihomir.jovicic.develop@gmail.com>


docker/api/service.py
```diff
@@ -476,3 +476,123 @@ def update_service(self, service, version, task_template=None, name=None,
             url, data=data, params={'version': version}, headers=headers
         )
         return self._result(resp, json=True)
+
+    @utils.minimum_version('1.31')
+    @utils.check_resource('service')
+    def rollback_service(self, service, version, task_template=None, name=None,
```
shin-: I don't think it makes sense for this to be a different method, since it will just hit the `update_service` endpoint with specific parameters. I would much rather see us add a well-documented `rollback` parameter to `update_service` instead.

On the other hand, I think it makes sense as a separate method in the high-level API.

tjovicic: Well, i didn't want to add one more argument to `update_service` method. 

I agree with you i need to refactor rollback method as it uses the same code as `update` method. I just wanted to check in with you. :) Do you agree i leave `rollback_service` method and refactor both `update_service` and `rollback_service` method?

shin-: For the `APIClient`, the mapping should be 1 API route = 1 method to keep consistent with the way it was designed. I don't think it's a good idea to stray from this.
In the same vein, the `update_service` method should have as many arguments as the `/service/update` endpoint has parameters.

Issue #1293, opened 2016-11-11
ResourceWarning: unclosed <socket.socket>


kelvich:
Hello.

It seems that in certain situations http socket isn't properly closed.

```
/usr/local/Cellar/python3/3.5.1/Frameworks/Python.framework/Versions/3.5/lib/python3.5/http/client.py:395: ResourceWarning: unclosed <socket.socket fd=7, family=AddressFamily.AF_UNIX, type=SocketKind.SOCK_STREAM, proto=0, raddr=/Users/stas/Library/Containers/com.docker.docker/Data/s60>
```

Python 3.5.1 on OSX.

Can be reproduced with following script:

```python
import unittest
import docker

# Warning depends on code layout, so add here some ballast classes,
# probably that puts GC in specific position.

class A(object):
    def __init__(self, node):
        print('A')

class B(object):
    def __init__(self, node):
        print('B')

class C(object):
    def __init__(self, node):
        print('C')

class D(object):
    def __init__(self, node):
        print('D')

class E(object):
    def __init__(self, node):
        print('E')

class SingleNodePartition(object):

    def __init__(self, node):
        self.node = node
        self.docker_api = docker.Client()

    def container_exec(self, node, command):
        exec_id = self.docker_api.exec_create(node, command, user='root')
        output = self.docker_api.exec_start(exec_id)

    def start(self):
        self.container_exec(self.node, "ls")
        self.container_exec(self.node, "ls")
        self.container_exec(self.node, "ls")
        self.container_exec(self.node, "ls")

class RecoveryTest(unittest.TestCase):

    def test_node_partition(self):
        failure = SingleNodePartition('node3')
        failure.start()

if __name__ == '__main__':
    unittest.main()
```

hensing:
Still a problem in version 2.4.2 (observed on python:3.6.1-alpine)

Lekensteyn:
Have you tried calling `docker_api.close()`? See https://docker-py.readthedocs.io/en/stable/client.html#docker.client.DockerClient.close

caaespin:
Has this been fixed yet?  I get this with docker py v4.2.0

Lekensteyn:
@caaespin presumably not, https://github.com/docker/docker-py/blob/master/docker/api/client.py has not been modified since March 15, 2019 while PR #2320 is still open.

asottile-sentry:
as a workaround:

```python
def docker_from_env() -> ContextManager[docker.DockerClient]:
[...]

PR #2320, opened 2019-04-26
exec: fix file handle leak with container.exec_* APIs

Lekensteyn:
Requests with stream=True MUST be closed or else the connection will
never be returned to the connection pool. Both ContainerApiMixin.attach
and ExecApiMixin.exec_start were leaking in the stream=False case.
exec_start was modified to follow attach for the stream=True case as
that allows the caller to close the stream when done (untested).

Tested with:

    # Test exec_run (stream=False) - observe one less leak
    make integration-test-py3 file=models_containers_test.py' -k test_exec_run_success -vs -W error::ResourceWarning'
    # Test exec_start (stream=True, fully reads from CancellableStream)
    make integration-test-py3 file=api_exec_test.py' -k test_execute_command -vs -W error::ResourceWarning'

After this change, one resource leak is removed, the remaining resource
leaks occur because none of the tests call client.close().

Fixes https://github.com/docker/docker-py/issues/1293
(Regression from https://github.com/docker/docker-py/pull/1130)

Signed-off-by: Peter Wu <pwu@cloudflare.com>
___
Relevant documentation:
https://2.python-requests.org//en/master/user/advanced/#body-content-workflow

Note: it is not easy to write a test to catch this issue. For one, lot of other tests also suffer from the resource leak issue (`client.close()` not being called). The other issue is that this error is only written to stderr.

While pytest supports the [capsys](https://docs.pytest.org/en/latest/reference.html#capsys) fixture to capture stderr, this is not easily possible due to the use of unittest which does not allow fixtures as function parameters: https://docs.pytest.org/en/latest/unittest.html
(Potential workarounds include using [request.getfixturevalue](https://docs.pytest.org/en/latest/reference.html#_pytest.fixtures.FixtureRequest.getfixturevalue), https://github.com/pytest-dev/pytest/issues/4143 or manually changing sys.stderr.)


jgfoster:
This fixes the problem I observed and I hope it gets incorporated soon!

inkychris:
Are there any technical reasons why this hasn't been merged, or has it just slipped under the radar? Tentatively tagging @shin- in case they may be able to help!

Nitrooo:
@milas can you please review and possibly merge this fix? Thank you.

milas (approved):
Thanks for the PR & fix!

PR #2325, opened 2019-05-01
Fix versions.py to include release stage

shin-:
This was causing CI to fail when trying to use the yet-to-be-released 19.03.0 stable instead of the beta.


shin-:
Going to ignore the flakes here as this does what's intended.

PR #2326, opened 2019-05-01
Avoid demux test flakiness

shin-:
Log order from the daemon seems to be inconsistent even with the delay added. This changes the test so that we don't assume a specific order, and instead just check that both lines are present in the return.


thaJeztah:
Thank you!

PR #2333, opened 2019-05-03
Add NetworkAttachmentConfig for service create/update

hannseman:
The current documentation makes it appear like the only way to attach networks to a service is to provide them as a list of strings of ids/names. 

Attaching networks also support setting aliases and driver options. I've added the new type `docker.types.NetworkAttachmentConfig` which is used to document these features.

I also found that a documentation link to `ConfigReference` did not work. It should be referenced as `docker.types.ConfigReference`, this is now fixed.


hannseman:
Let me know if there's anything I can help with regarding this PR. Thanks!

hannseman:
@shin- No problem! Thanks for the review.

PR #2335, opened 2019-05-09
reference swarm page correctly

isidentical:
Current version raises an HTTP404 when [trying](https://docker-py.readthedocs.io/en/stable/swarm.rst) to access swarm page.


shin-:
Thanks!

PR #2351, opened 2019-05-24
Fix documentation for inspect_secret referring to removal.

mildlyincompetent:
The documentation for `inspect_secret` takes a secret ID, but documents this as the "full ID of the secret to remove". Since `inspect_secret` inspects a secret as opposed to removing it, it would be more appropriate to document this as "full ID of the secret to inspect".


shin-:
Thank you! :)

Issue #2359, opened 2019-06-15
 docker.from_env blocks with paramiko on python2.7


ssbarnea:
The below block always reproduces with python2.7 on MacOS when ``DOCKER_HOST=ssh://somehost``. The same code works fine with on other python versions and I also tested with latest ``paramiko==2.5.0``. This seems to be related to an ancient paramiko ticket at https://github.com/paramiko/paramiko/issues/109 which could be avoided based on few comments I read there.

```
exitstatus: 2
    client = docker.from_env(timeout=5, version="auto")
../../.pyenv/versions/2.7.16/lib/python2.7/site-packages/docker/client.py:85: in from_env
terminalreporter: <_pytest.terminal.TerminalReporter object at 0x109c54ed0>
    timeout=timeout, version=version, **kwargs_from_env(**kwargs)
../../.pyenv/versions/2.7.16/lib/python2.7/site-packages/docker/client.py:40: in __init__
    self.api = APIClient(*args, **kwargs)
../../.pyenv/versions/2.7.16/lib/python2.7/site-packages/docker/api/client.py:166: in __init__
    base_url, timeout, pool_connections=num_pools
../../.pyenv/versions/2.7.16/lib/python2.7/site-packages/docker/transport/sshconn.py:84: in __init__
    self._connect()
../../.pyenv/versions/2.7.16/lib/python2.7/site-packages/docker/transport/sshconn.py:94: in _connect
    parsed.hostname, parsed.port, parsed.username,
../../.pyenv/versions/2.7.16/lib/python2.7/site-packages/paramiko/client.py:397: in connect
    t.start_client(timeout=timeout)
../../.pyenv/versions/2.7.16/lib/python2.7/site-packages/paramiko/transport.py:597: in start_client
    event.wait(0.1)
../../.pyenv/versions/2.7.16/lib/python2.7/threading.py:614: in wait
    self.__cond.wait(timeout)
../../.pyenv/versions/2.7.16/lib/python2.7/threading.py:359: in wait
    _sleep(delay)
E   KeyboardInterrupt
```

ssbarnea:
I managed to narrow down the issue a little bit more, it seems that this only happens when the code is run with pytest: ``python2.7 -m pytest bug.py`:

```
import docker

c = docker.from_env(timeout=5, version="auto")
if not c.ping():
    raise Exception("Failed to ping docker server.")


def test_foo():
    print("done!")
```

[...]

PR #2360, opened 2019-06-15
Avoid blocking caused by ssh connect on paramiko

ssbarnea:
Fix bug where the ssh connection was missing the timeout parameter
and its default value being None (wait forever).

Partial-Fix: #2359


ssbarnea:
@shin- @trickeydan Please review.

ssbarnea:
Apparently Docker Inc lacks ability to review even simple contributions from community. The only involvement was from a bot.

PR #2380, opened 2019-07-12
xfail test_init_swarm_data_path_addr

thaJeztah:
Relates to https://github.com/moby/moby/pull/39068#issuecomment-510943803

This test can fail if `eth0` has multiple IP addresses;

    E   docker.errors.APIError: 400 Client Error: Bad Request ("interface eth0 has more than one IPv6 address (2001:db8:1::242:ac11:2 and fe80::42:acff:fe11:2)")

Which is not a failiure, but depends on the environment that
the test is run in.


thaJeztah:
@shin- ptal 🤗

shin-:
Hm, I wish there was a "safe" value we could use, but I guess this does the trick in the meantime 🤷‍♂

PR #2393, opened 2019-07-30
[WIP] Fix #2389

### LINKED ISSUE #2389: Loading multiple docker images saved with image IDs returns empty list
Environment:

- Docker server version: 18.09.7-ce
- docker-py version: 2.0.2

To reproduce the problem, run the following code:

```python
import subprocess

import docker


def main():
    docker_client = docker.from_env(version='auto')
    image_names = ['debian', 'ubuntu']  # Replace this with any of your Docker images names.
    image_ids = [docker_client.images.get(image_name).id for image_name in image_names]

    with subprocess.Popen(['docker', 'save', *image_ids], stdout=subprocess.PIPE) as docker_save_process:
        print(docker_client.images.load(docker_save_process.stdout))


if __name__ == '__main__':
    main()
```

I did a little digging, it seems that in the following code:

https://github.com/docker/docker-py/blob/17c86429e454e65f74660393ffb427adb8db154a/docker/models/images.py#L383-L393

`chunk` contains multiple lines of loaded image info that makes the regular expression match fail.

EFanZh:
Test cases are missing, I will add test cases later.


feliperuhland:
The code looks good.

PR #2399, opened 2019-08-07
Correct INDEX_URL logic in build.py _set_auth_headers

fluetm:
A typo in conditional logic can prevent `auth.INDEX_URL` from being added to `auth_data` in cases where `auth.INDEX_URL` is missing but an `auth.INDEX_NAME` is available.

Signed-off-by: Matt Fluet <matt.fluet@appian.com>


fluetm:
@shin- ready for a review

PR #2401, opened 2019-08-07
Fix for empty auth keys in config.json

fluetm:
`docker-compose` builds that pull from private registries are broken on macs that use the OS keychain (ie. osxkeychain credsStore) due to a change in the way the auth information is processed in `auth.py`.

The osxkeychain maintains an auths section with host keys but no values in `.docker/config.json`, so this needs to be accounted for:
```
  "auths": {
    "some.registry.host": {},
    },
```

This change will resolve any keys from the auths that have no values defined as a first step.  Confirmed that this fixed the build for me locally.

Affects docker-py 3.7.0+ and docker-compose 1.24.0+


fluetm:
@shin- FYI: a fix for the problem with macs pulling images from private registries during docker-compoose builds.  Can you take a look at this?

randy5235:
We are seeing this behavior as well. It is impacting us and wasted a fair amount of time.

shin-:
@fluetm Thanks for the submission! I think we can minimize the amount of roundtrips we're making here by calling `convert_to_hostname()` on the results of the `store.list()` call instead of adding a second loop. So, instead of:
```
            for k in store.list().keys():
                auth_data[k] = self._resolve_authconfig_credstore(
                    k, self.creds_store
                )
```

we should do

```
            for k in store.list().keys():
                auth_data[convert_to_hostname(k)] = self._resolve_authconfig_credstore(
                    k, self.creds_store
                )
```

otherwise we're inefficiently fetching data twice.

Having some unit tests to verify the new behavior would also be great! We already have a handful of them [here](https://github.com/docker/docker-py/blob/master/tests/unit/auth_test.py#L460), but could obviously use more!

fluetm:
@shin- I can definitely adjust the fix here and add some test coverage.  Had a followup question on your suggestion though...will changing the format of the key have any side effects?  Any auths coming from `config.json` starting with `https://` will now be empty.  Maybe that's not going to affect anything, but perhaps a safer fix would be to add the credentials under both key formats?  Something like this?
```
for k in store.list().keys():
                auth_data[k] = self._resolve_authconfig_credstore(
                    k, self.creds_store
                )
                auth_data[convert_to_hostname(k)] = auth_data[k]
```

shin-:
I'm good with that, sure!

fluetm:
@shin- have the amended fix ready with some new tests.  Everything's passing...can you take another look?

scott-joe:
@shin- This is breaking our dev environment. Would appreciate a 2nd look. Thanks!

adamkelm:
@shin- Lots of hours spent trying to figure out this issue. Would love to see a fix merged as soon as possible. Thanks!

forstermatth:
@fluetm Thanks a tonne for your work on this.

fluetm:
@shin- do you know when the 4.1.0 release will be cut?

ilyaGotfryd:
@shin- I am presently running docker 19.03.4. Any idea how I can get a hold of this update or when it is going to become available?

PR #2419, opened 2019-08-29
Gpus update

hnine999:
Unfortunately wasn't able to build docs -- doesn't look like they need to be updated for this change(?)

Mod was based on "runtime" argument -- searched for this argument in the source files, and wherever it appeared, added "gpus" argument, with dependency on api version 1.40 (the most recent version as of Docker version 19.03.1, build 74b1e89).


DesktopMan:
Is this still being worked on? As far as I understand "Gpus" is not part of HostConfig which is being set here. When --gpus is used with Docker run the GPU configuration ends up in the DeviceRequests section under HostConfig, as documented here:
https://docs.docker.com/engine/api/v1.40/#operation/ContainerCreate

hnine999:
Still being worked on.

sigurdkb:
I did a quick hack based on the patch by @hnine999  that grabs all available gpus (https://github.com/sigurdkb/docker-py/tree/gpus_parameter). It is not pretty, but it works :-)

As mentioned by @DesktopMan, it converts any gpus parameter value to a hard coded DeviceRequests section under HostConfig that grabs all GPUs:
https://github.com/sigurdkb/docker-py/blob/f5e11cdc6e3bd179312aceededf323cbb7cdc448/docker/types/containers.py#L525

sigurdkb:
Considering the Docker API implements the DeviceRequests parameter and not the Gpus parameter, which parameter should this library use? device_requests would match the Docker API, while gpus match the Docker client. Are there any guidelines for this?

AndrewJDR:
Who is responsible for answering some of the questions that @sigurdkb posed so that this can be addressed?

I'm open to submitting a PR myself, but the fact that @sigurdkb has already proposed a PR several weeks ago, asked some architectural questions on how to proceed, and still hasn't received any response to those questions is not encouraging.

Thanks.

Kamagawa:
Any Progress on this PR?

La0:
There is a cleaner and better implementation here https://github.com/docker/docker-py/pull/2471

EpicWink:
Is this superseded by #2471?

PR #2440, opened 2019-09-30
Add network_alias argument to client.containers.run.

dhatch:
Adds the ability to pass `network_alias` as an option to `client.containers.run`.  This option is supported in the docker CLI as [`--network-alias`](https://docs.docker.com/v17.09/engine/reference/run/#network-settings).

This is available in the low-level CLI, but would be great if it is available in the higher level API.


Manu10744:
It would be great if this is going to be merged.

PR #2448, opened 2019-10-14
Support .ssh/config

ndeloof:
A refreshed version of https://github.com/docker/docker-py/pull/2290

Would allow to fix https://github.com/docker/compose/issues/6655

image

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -32,7 +32,7 @@ def buildImages = { ->
 
 def getDockerVersions = { ->
   def dockerVersions = ["17.06.2-ce"]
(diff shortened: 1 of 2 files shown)


rdavaillaud:
Hi.

When can we expect a new release with this fix, and then release a new docker-compose version?

ssbarnea:
I suspect that this solves the problem of docker-py failing to use API when you do not mention username in DOCKER_HOST URI but works perfectly from CLI (user is configured in ssh config)

When should we be able to use this? I see no release made since the code was merged.

PR #2456, opened 2019-11-05
Add empheral dockerfile handling

### LINKED ISSUE #2105: Can not pass `fileobj` together with `path` context to `build`
It seems to be impossible to pass ephemeral `Dockerfile` to build command together with build context. Here is command line use case:

    docker build -t . -f-<<EOF
    Lorem ipsum docker file
    EOF

According to the documentation https://docker-py.readthedocs.io/en/stable/images.htmlit should be possible to pass `Dockerfile` as file object, but in current implementation `path` param with context is ignored.

Here is path that fix part of this issue (it don't work when `path.startswith(('http://', 'https://','git://', 'github.com/', 'git@'))`)

    diff --git a/docker/api/build.py b/docker/api/build.py
    index 0486dce..711def2 100644
    --- a/docker/api/build.py
    +++ b/docker/api/build.py
    @@ -133,7 +133,7 @@ class BuildApiMixin(object):
                if not fileobj:
                    raise TypeError("You must specify fileobj with custom_context")
                context = fileobj
    -        elif fileobj is not None:
    +        elif path is None and fileobj is not None:
                context = utils.mkbuildcontext(fileobj)
            elif path.startswith(('http://', 'https://',
                                'git://', 'github.com/', 'git@')):
    @@ -149,7 +149,7 @@ class BuildApiMixin(object):
                            lambda x: x != '' and x[0] != '#',
                            [l.strip() for l in f.read().splitlines()]
                        ))
    -            dockerfile = process_dockerfile(dockerfile, path)
    +            dockerfile = process_dockerfile(dockerfile, path, fileobj)
                context = utils.tar(
                    path, exclude=exclude, dockerfile=dockerfile, gzip=gzip
                )
    @@ -332,7 +332,13 @@ class BuildApiMixin(object):
                log.debug('No auth config found')
    
    
    -def process_dockerfile(dockerfile, path):
    +def process_dockerfile(dockerfile, path, fileobj):
    +    if fileobj is not None:
    +        return (
    +            '.dockerfile.{0:x}'.format(random.getrandbits(160)),
    +            fileobj.read()
    +        )
    +
        if not dockerfile:
            return (None, None)

majkrzak:
close #2105 according to https://github.com/docker/docker-py/issues/2105#issuecomment-411154477

### DIFF
--- a/docker/api/build.py
+++ b/docker/api/build.py
@@ -330,6 +330,12 @@ def process_dockerfile(dockerfile, path):
     if not dockerfile:
         return (None, None)
 
+    if '\n' in dockerfile:
+        return (
+            '.dockerfile.{0:x}'.format(random.getrandbits(160)),
+            dockerfile
+        )
+
     abs_dockerfile = dockerfile
     if not os.path.isabs(dockerfile):
         abs_dockerfile = os.path.join(path, dockerfile)


majkrzak:
:man_facepalming:

majkrzak:
:wave: are you fine down there? :wave:

majkrzak:
:wave:

majkrzak:
I wonder if this PR will get merged faster than https://github.com/gkoberger/stacksort/pull/4

ulyssessouza:
Hello @majkrzak. I think that trying to detect if it's the content or not is not an accurate way.
For example, this is a valid `Dockerfile` content AND a filename
```
FROM ubuntu
```

IMO another parameter with the content would be better

PR #2458, opened 2019-11-08
Save multiple images into single tar file from ImageCollection

gravypod:
The image collection in the Docker API allows you to save multiple
images into a single bundle. This makes it easy to export and transfer
multiple containers together.

This feature is documented within the swagger for the apiserver:
https://github.com/moby/moby/blob/6f8c671d702197a189d162d86a3f4cccfa5a3db2/api/swagger.yaml#L7617-L7645

This feature mirrors the docker cli's `docker image save` command:
https://docs.docker.com/engine/reference/commandline/image_save/


gravypod:
If there's anything that needs to be updated with this just let me know and I'll get it done ASAP. Thanks for your time!

PR #2469, opened 2019-11-28
Switch from pypiwin32 to pywin32

mrginglymus:
Now that the supported versions of python are 2.7 and 3.5+, we can
switch to pywin32 (which as of 223 supports 2.7, 3.5, 3.6 and 3.7).

Arguably the lower bound is not required; it's merely included such that
for python 3+ the requirement remains the same as pypiwin32==223 was a
proxy for pywin32>=223.

Signed-off-by: Bill Collins <bill@bill.works>

### DIFF
--- a/docker/api/client.py
+++ b/docker/api/client.py
@@ -156,7 +156,7 @@ class APIClient(
                 )
             except NameError:
                 raise DockerException(
-                    'Install pypiwin32 package to enable npipe:// support'
+                    'Install pywin32 package to enable npipe:// support'
                 )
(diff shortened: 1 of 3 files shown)


arossert:
Can this be merged to master? the `pypiwin32` is deprecated and causing some issues if installed together with `pywin32`.

charlesdarkwind:
Can this be merged? The issue might be preventing me from installing the moto package on my system https://github.com/spulec/moto/issues/2686

mrginglymus:
Fixed in https://github.com/docker/docker-py/pull/2593

PR #2470, opened 2019-11-28
Fix secret repr

feliperuhland:
This PR fixes issue #2025 

- Create a unit test to reproduce the repr issue - 52d030d
- Fix secrets repr issue - 514bfd0

### DIFF
--- a/docker/models/secrets.py
+++ b/docker/models/secrets.py
@@ -4,14 +4,22 @@ from .resource import Model, Collection
 
 class Secret(Model):
     """A secret."""
-    id_attribute = 'ID'
+    id_attribute = 'Id'
 
     def __repr__(self):
         return "<%s: '%s'>" % (self.__class__.__name__, self.name)
 
     @property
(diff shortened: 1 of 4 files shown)


etienne-napoleone:
Maybe it would be better to change the `__repr__` method than the name getter?
I'm pretty sure that someone using `my_secret.name` is not expecting to get an id instead of a name in certain cases.

feliperuhland:
> Maybe it would be better to change the **repr** method than the name getter?
> I'm pretty sure that someone using `my_secret.name` is not expecting to get an id instead of a name in certain cases.

Thank you, @etienne-napoleone.

I changed the repr in 7750c91

feliperuhland:
> docker.errors.APIError: 503 Server Error: Service Unavailable ("Swarm is encrypted and needs to be unlocked before it can be used. Please use "docker swarm unlock" to unlock it.")

Seems like a CI problem. Is there a way to run it again?

PR #2478, opened 2019-12-10
put_archive: note the data may also be a stream

akx:
The docs imply the `data` parameter to `put_archive()` must be `bytes` (a bytestring), which is not the case; a stream is fine too [since it's passed directly to Requests](https://2.python-requests.org/en/master/user/advanced/#streaming-uploads).

In fact, the tests for the `put_archive()` feature always pass in a stream:

https://github.com/docker/docker-py/blob/a0b9c3d0b38abd4af1880ca3dde2845556dd2f70/tests/integration/api_container_test.py#L678-L679

### DIFF
--- a/docker/api/container.py
+++ b/docker/api/container.py
@@ -966,7 +966,7 @@ class ContainerApiMixin:
             container (str): The container where the file(s) will be extracted
             path (str): Path inside the container where the file(s) will be
                 extracted. Must exist.
-            data (bytes): tar data to be extracted
+            data (bytes or stream): tar data to be extracted
 
         Returns:
             (bool): True if the call succeeds.
--- a/docker/models/containers.py
+++ b/docker/models/containers.py
@@ -324,7 +324,7 @@ class Container(Model):
         Args:
             path (str): Path inside the container where the file(s) will be
                 extracted. Must exist.
-            data (bytes): tar data to be extracted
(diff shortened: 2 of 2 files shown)


felixfontein:
This is also something I noticed recently. I've taken a look at the Requests adapters provided by docker-py, it doesn't look like they do anything funky which will prevent this. But I'm not really familiar with these aspects of Requests (and the adapters) to be able to say this with full confidence :)

akx:
@felixfontein Well, it works for me in production as we speak 😁

felixfontein:
@akx out of curiosity, which transports are you using it with / did you test it with? There's `unix`, `tcp`, `tcp+tls`, `ssh+unix`, and `npipe` (Windows only).

akx:
@felixfontein UNIX in this instance.

damwiw:
Hi,
After a code review of request.py and Go docker client, I can confirm:
- request accepts streams (documentation: file-like objects) as data field. __iter__ and read methods are expected.
- docker cp command with stdin as input uses such a mechanism.
By the way, tar archive may be a tar gzipped archive.
Sucessfully tested with 4 Go of compressed tar archive.

akx:
Rebased. Could a maintainer take a look?

milas (approved):
Thanks for rebasing, let's get this merged 🙂

Issue #2477, opened 2019-12-05
Maximum number of connections


lorenzo93:
Hi,

I am currently using your library to run a lot of docker containers simultaneously.
I am working on a server with 40+ cores and using a thread-pool to scale up.

Unfortunately your code is limited to a thread-pool of 10 threads due to your initialization of urllib3 [here](https://github.com/docker/docker-py/blob/4d7d4084138fa6161ef5f31d410b0d326d41f777/docker/transport/unixconn.py#L59) so we cannot run too many threads.

Is it possible to change this behavior and allow the user to set the max connection pool of urllib3?

Thanks a lot

feliperuhland:
Hi @lorenzo93 

I took a quick look at the function, and I'm pretty sure that it can be done adding the parameter to get_connection and subsequent functions to change that behavior.

Do you want to propose that change and send a pull request?

lorenzo93:
Hi @feliperuhland ,

I and @tcaiazzi studied the code today and couldn't find where the get_connection method is called.

Wouldn't be better to change the default theads to Cpu count directly into the urllib3 inizialization?

What do you think?

feliperuhland:
Hi @lorenzo93 

As you point it out, the ` maxsize` attribute is defined in https://github.com/docker/docker-py/blob/a0b9c3d0b38abd4af1880ca3dde2845556dd2f70/docker/transport/unixconn.py#L59

`UnixHTTPConnectionPool` is instantiated in https://github.com/docker/docker-py/blob/a0b9c3d0b38abd4af1880ca3dde2845556dd2f70/docker/transport/unixconn.py#L97

`UnixHTTPAdapter` is instanciated in https://github.com/docker/docker-py/blob/a0b9c3d0b38abd4af1880ca3dde2845556dd2f70/docker/api/client.py#L140 inside `APIClient`:

https://github.com/docker/docker-py/blob/a0b9c3d0b38abd4af1880ca3dde2845556dd2f70/docker/api/client.py#L102

Now you can suggest the `max_pool_size` - or a better name - field to APIClient until `UnixHTTPConnection` initialization. It can be done with `**kwargs` as well.

Any change like that one must happen with good test coverage and well wrote documentation.

PR #2480, opened 2019-12-18
Add max_pool_size parameter to api/client.py 

tcaiazzi:
As requested in #2477, it is now possible to choose the number of connections for each urllib3 pool.

### DIFF
--- a/docker/client.py
+++ b/docker/client.py
@@ -1,5 +1,5 @@
 from .api.client import APIClient
-from .constants import DEFAULT_TIMEOUT_SECONDS
+from .constants import (DEFAULT_TIMEOUT_SECONDS, DEFAULT_MAX_POOL_SIZE)
 from .models.configs import ConfigCollection
 from .models.containers import ContainerCollection
 from .models.images import ImageCollection
@@ -28,6 +28,8 @@ class DockerClient(object):
         version (str): The version of the API to use. Set to ``auto`` to
             automatically detect the server's version. Default: ``1.35``
         timeout (int): Default timeout for API calls, in seconds.
+        max_pool_size (int): The maximum number of connections
+            to save in the pool.
         tls (bool or :py:class:`~docker.tls.TLSConfig`): Enable TLS. Pass
             ``True`` to enable it with default options, or pass a
             :py:class:`~docker.tls.TLSConfig` object to use custom
@@ -64,6 +66,8 @@ class DockerClient(object):
             version (str): The version of the API to use. Set to ``auto`` to
                 automatically detect the server's version. Default: ``1.35``
             timeout (int): Default timeout for API calls, in seconds.
+            max_pool_size (int): The maximum number of connections
+                to save in the pool.
             ssl_version (int): A valid `SSL version`_.
             assert_hostname (bool): Verify the hostname of the server.
             environment (dict): The environment to read environment variables
@@ -82,9 +86,12 @@ class DockerClient(object):
         """
         timeout = kwargs.pop('timeout', DEFAULT_TIMEOUT_SECONDS)
         version = kwargs.pop('version', None)
-        max_pool_size = kwargs.pop('max_pool_size', 10)
+        max_pool_size = kwargs.pop('max_pool_size', DEFAULT_MAX_POOL_SIZE)
         return cls(
-            timeout=timeout, version=version, max_pool_size=max_pool_size, **kwargs_from_env(**kwargs)
+            timeout=timeout,
+            version=version,
+            max_pool_size=max_pool_size,
+            **kwargs_from_env(**kwargs)
         )
 
     # Resources
--- a/docker/client.py
+++ b/docker/client.py
@@ -74,7 +74,6 @@ class DockerClient(object):
                 from. Default: the value of ``os.environ``
             credstore_env (dict): Override environment variables when calling
                 the credential store process.
-            max_pool_size (int): number of connections for pool
 
         Example:
 
--- a/docker/client.py
+++ b/docker/client.py
@@ -70,6 +70,7 @@ class DockerClient(object):
                 from. Default: the value of ``os.environ``
             credstore_env (dict): Override environment variables when calling
                 the credential store process.
+            max_pool_size (int): number of connections for pool
 
         Example:
 
@@ -81,8 +82,9 @@ class DockerClient(object):
         """
         timeout = kwargs.pop('timeout', DEFAULT_TIMEOUT_SECONDS)
         version = kwargs.pop('version', None)
+        max_pool_size = kwargs.pop('max_pool_size', 10)
         return cls(
-            timeout=timeout, version=version, **kwargs_from_env(**kwargs)
+            timeout=timeout, version=version, max_pool_size=max_pool_size, **kwargs_from_env(**kwargs)
         )
 
     # Resources
--- a/docker/transport/npipeconn.py
+++ b/docker/transport/npipeconn.py
@@ -34,9 +34,9 @@ class NpipeHTTPConnection(httplib.HTTPConnection, object):
 
 
 class NpipeHTTPConnectionPool(urllib3.connectionpool.HTTPConnectionPool):
-    def __init__(self, npipe_path, timeout=60, maxsize=10):
+    def __init__(self, npipe_path, timeout=60, max_pool_size=10):
         super(NpipeHTTPConnectionPool, self).__init__(
-            'localhost', timeout=timeout, maxsize=maxsize
+            'localhost', timeout=timeout, maxsize=max_pool_size
         )
         self.npipe_path = npipe_path
         self.timeout = timeout
@@ -73,15 +73,18 @@ class NpipeHTTPAdapter(BaseHTTPAdapter):
 
     __attrs__ = requests.adapters.HTTPAdapter.__attrs__ + ['npipe_path',
                                                            'pools',
-                                                           'timeout']
+                                                           'timeout',
+                                                           'max_pool_size']
 
     def __init__(self, base_url, timeout=60,
-                 pool_connections=constants.DEFAULT_NUM_POOLS):
+                 pool_connections=constants.DEFAULT_NUM_POOLS,
+                 max_pool_size=10):
         self.npipe_path = base_url.replace('npipe://', '')
         self.timeout = timeout
         self.pools = RecentlyUsedContainer(
             pool_connections, dispose_func=lambda p: p.close()
         )
+        self.max_pool_size = max_pool_size
         super(NpipeHTTPAdapter, self).__init__()
 
     def get_connection(self, url, proxies=None):
@@ -91,7 +94,8 @@ class NpipeHTTPAdapter(BaseHTTPAdapter):
                 return pool
 
             pool = NpipeHTTPConnectionPool(
-                self.npipe_path, self.timeout
+                self.npipe_path, self.timeout,
+                max_pool_size=self.max_pool_size
             )
             self.pools[url] = pool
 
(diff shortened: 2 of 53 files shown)


docker/client.py
```diff
@@ -64,12 +66,15 @@ def from_env(cls, **kwargs):
...
             timeout (int): Default timeout for API calls, in seconds.
+            max_pool_size (int): The maximum number of connections
+                to save in the pool.
             ssl_version (int): A valid `SSL version`_.
             assert_hostname (bool): Verify the hostname of the server.
             environment (dict): The environment to read environment variables
                 from. Default: the value of ``os.environ``
             credstore_env (dict): Override environment variables when calling
                 the credential store process.
+            max_pool_size (int): number of connections for pool
```
feliperuhland: This argument is duplicated.

docker/transport/npipeconn.py
```diff
@@ -34,9 +34,9 @@ def connect(self):
 
 
 class NpipeHTTPConnectionPool(urllib3.connectionpool.HTTPConnectionPool):
-    def __init__(self, npipe_path, timeout=60, maxsize=10):
+    def __init__(self, npipe_path, timeout=60, max_pool_size=10):
         super(NpipeHTTPConnectionPool, self).__init__(
-            'localhost', timeout=timeout, maxsize=maxsize
+            'localhost', timeout=timeout, maxsize=max_pool_size
```
feliperuhland: That change will break backward compatibility.

docker/transport/sshconn.py
```diff
@@ -38,9 +38,9 @@ def connect(self):
 class SSHConnectionPool(urllib3.connectionpool.HTTPConnectionPool):
     scheme = 'ssh'
 
-    def __init__(self, ssh_client, timeout=60, maxsize=10):
+    def __init__(self, ssh_client, timeout=60, max_pool_size=10):
         super(SSHConnectionPool, self).__init__(
-            'localhost', timeout=timeout, maxsize=maxsize
+            'localhost', timeout=timeout, maxsize=max_pool_size
```
feliperuhland: That change also breaks backward compatibility.

docker/transport/unixconn.py
```diff
@@ -56,9 +56,9 @@ def response_class(self, sock, *args, **kwargs):
 
 
 class UnixHTTPConnectionPool(urllib3.connectionpool.HTTPConnectionPool):
-    def __init__(self, base_url, socket_path, timeout=60, maxsize=10):
+    def __init__(self, base_url, socket_path, timeout=60, max_pool_size=10):
```
feliperuhland: That breaks backward compatibility also.

Skazzino:
Hi @feliperuhland, I've fixed what you reported in the previous comment. My colleague @tcaiazzi erroneously renamed a parameter in *ConnectionPool classes. However, we don't know how to automatically test if the pool size changes accordingly (we tested it manually on our computers). 

If you have any ideas, report them here and we'll be happy to build the tests.

feliperuhland:
Hi @Skazzino.

Thank you for the changes.

About the tests, `urllib3.connectionpool.HTTPConnectionPool` can be mocked and the test can check if `APIClient` argument changed `HTTPConnectionPool` the pool size. That will be useful to guarantee that any future change will break this behavior.

Also, the commit must be signed https://github.com/docker/docker-py/pull/2480#issuecomment-566976649

Skazzino:
Hi @feliperuhland,
sorry if I bother you again but for me it's the first time that I write a test for an API like this... So I'm quite newbie on it. I don't know if the test that you require should be an unit test or an integration test.

I think that [this one](https://github.com/docker/docker-py/blob/master/tests/integration/api_client_test.py) is really similar to what we want to test (since I can create an APIClient instance with a desired pool size). In other case, I don't have any idea of how mock the urllib3 and inject it into the APIClient.

Thanks for your patience.

Skazzino:
Hi @feliperuhland,
still waiting for a response! Thank you! :)

feliperuhland:
Hi @Skazzino 

IMHO, those tests should be a unit test, like [this one](https://github.com/docker/docker-py/blob/master/tests/unit/client_test.py).

If you need help, let me know.

Skazzino:
Hi @feliperuhland,
I've managed to create 2 unit tests for the max pool size. However, I wasn't able to mock the `urllib3.connectionpool.HTTPConnectionPool` because each Adapter internally creates those objects when required.

I was able to read the actual value from each `urllib3.connectionpool.HTTPConnectionPool` through the Adapter and I asserted if it is equal to the desired pool size.

I hope that this test is enough for this feature. 

Thanks.

feliperuhland:
Hi @Skazzino 

My mock idea was not to need a docker daemon. Something like:

```python
@mock.patch("docker.transport.unixconn.UnixHTTPConnectionPool")              
def test_default_pool_size(self, mock_obj):                                                                 
    client = docker.DockerClient()                                                                          
    client.ping()                                                                                           
    base_url = "{base_url}/v{version}/_ping".format(base_url=client.api.base_url, version=client.api._version)
    timeout = 60                                                                                            
    docker_sock = "/var/run/docker.sock"                                                                    
    mock_obj.assert_called_once_with(base_url, docker_sock, timeout, maxsize=10)
```

Skazzino:
Hi @feliperuhland,
thanks for the help! I was heading in the same direction yesterday (after pushing the last commit), but I found those problems:
1. Without a Docker Daemon running, this test still fails with `requests.exceptions.ConnectionError: ('Connection aborted.', ConnectionRefusedError(111, 'Connection refused'))` because requests catches the Exception before opening a pool.
2. From the stack trace of AppVeyor, it uses NPipe socket (so probably Windows): `docker\transport\npipesocket.py:59: in connect`. So if I patch the `UnixHTTPConnectionPool` this test won't work.
3. Even resolving the point 1 and 2, another issue remains. The "mocked response" returned from the UnixHTTPConnectionPool raises this Exception 
```
    if 400 <= self.status_code < 500:
TypeError: '<=' not supported between instances of 'int' and 'MagicMock'
```
Since the mocked response cannot be compared.

Hope we can find a way to make it work! :)

Skazzino:
Hi @feliperuhland,
any news? Sorry if I bother you but this PR is really important to me and my project. 

Thanks.

feliperuhland:
Hi @Skazzino, I'll take a look today. Sorry about the delay.

feliperuhland:
Hi @Skazzino 

1. If the transport class is mocked, the docker daemon will not start. That way, the test will run fast and can check if the class was called correctly.

2. To create tests for Linux and windows, it should be used `pytest.mark.skipif`, like [here](https://github.com/docker/docker-py/blob/cb19bf117d5088904f15fd7f0cf21213c7141555/tests/unit/utils_build_test.py#L232), then you can specify the transport class.

3. The idea to mock the transport class is to be sure that the variable used in DockerClient changes the transport class behavior. In other words, the primary purpose is to be sure that the transport class is called with the correct values.

Sorry again for the delay.

I hope that helps and if you need anything, let me know.

Skazzino:
Hi @feliperuhland,

>     1. If the transport class is mocked, the docker daemon will not start. That way, the test will run fast and can check if the class was called correctly.

Ok.

>     2. To create tests for Linux and windows, it should be used `pytest.mark.skipif`, like [here](https://github.com/docker/docker-py/blob/cb19bf117d5088904f15fd7f0cf21213c7141555/tests/unit/utils_build_test.py#L232), then you can specify the transport class.

Okay, I'll add them to the tests (accordingly to what I'm mocking)

>     3. The idea to mock the transport class is to be sure that the variable used in DockerClient changes the transport class behavior. In other words, the primary purpose is to be sure that the transport class is called with the correct values.

Yes, I agree with you. But if I run the unit tests (like the example you wrote before), I get the following stack trace, since the mocked response isn't an integer and I cannot access to the response from the Client:

```
  File "/usr/lib/python3.7/unittest/mock.py", line 1204, in patched
    return func(*args, **keywargs)
  File "/home/skazza/Scrivania/git/docker-py/tests/unit/client_test.py", line 85, in test_default_pool_size
    client.info()
  File "/usr/local/lib/python3.7/dist-packages/docker-4.1.0.dev0-py3.7.egg/docker/client.py", line 187, in info
    return self.api.info(*args, **kwargs)
  File "/usr/local/lib/python3.7/dist-packages/docker-4.1.0.dev0-py3.7.egg/docker/api/daemon.py", line 95, in info
    return self._result(self._get(self._url("/info")), True)
  File "/usr/local/lib/python3.7/dist-packages/docker-4.1.0.dev0-py3.7.egg/docker/api/client.py", line 271, in _result
  File "/usr/local/lib/python3.7/dist-packages/docker-4.1.0.dev0-py3.7.egg/docker/api/client.py", line 265, in _raise_for_status
    response.raise_for_status()
  File "/home/skazza/.local/lib/python3.7/site-packages/requests/models.py", line 933, in raise_for_status
    if 400 <= self.status_code < 500:
TypeError: '<=' not supported between instances of 'int' and 'MagicMock'
```

feliperuhland:
Hi @Skazzino 

Is [this](https://github.com/tcaiazzi/docker-py) branch up to date?

Skazzino:
Hi @feliperuhland,
and thanks for the quick reply! I've pushed the new tests right now.

feliperuhland:
Hi @Skazzino 

The first thing I noticed was the need to mock `response.status` when the ping method is called. It's necessary to add the return value behavior:

```python
mock_obj.return_value.urlopen.return_value.status = 200
```

Another thing that I saw was the `assert_called_once_with` call that needs to be the same as the original one, so it should be without `docker_sock` and `timeout` parameters name.

```python
 mock_obj.assert_called_once_with(base_url, "/var/run/docker.sock", 60, maxsize=DEFAULT_MAX_POOL_SIZE)
```

I hope that could help.

Skazzino:
Hi @feliperuhland,
new unit tests passed but AppVeyor failed on insalling a library (I don't think this is a problem related to the edits that we made). Any ideas? 

Thanks.

feliperuhland:
Hi @Skazzino 

It seems related to dependencies installation:

`ERROR: Package 'setuptools' requires a different Python: 2.7.17 not in '>=3.5'`

feliperuhland:
Also, it's essential to rebase from master to be sure that everything is up to date.

:)

Skazzino:
Hi @feliperuhland,
I've merged the fork with the master. However the failure persists... I haven't touched the requirements.txt (nor the test-requirements.txt).

It seems to be a Python2.7 related issue. How we should proceed? 

Thanks.

Skazzino:
Hi @feliperuhland,
still waiting a response. Thanks.

Skazzino:
Hi @feliperuhland,
after 17 days I'm still waiting a response... Thanks.

tcaiazzi:
Hi @feliperuhland,
I'm still waiting a response!
Thanks!

feliperuhland:
Hi @Skazzino 

Cloud you update de branch again, please?

I push into another branch, and all the tests passed.

Thanks

Skazzino:
Hi @feliperuhland,
all tests passed, finally! 

Hope everything is okay now, tell me if you need something else. Thanks for all the help in these months.

feliperuhland:
Hi @Skazzino 
You have to sign the commits. https://github.com/docker/docker-py/pull/2480#issuecomment-566976649

Skazzino:
Hi @feliperuhland,
is this command still valid? `git rebase -i HEAD~842354325856` or should i change the HEAD stuff?

feliperuhland:
Hi @Skazzino 

You can use `git rebase -i HEAD~30`, (the number represents how many commits do you want to rebase) and edit commits one by one.

If you are not confortable with git command line, you can create another branch to execute the commands.

Skazzino:
Hi @feliperuhland,
I've signed the commit.

feliperuhland:
The bot should approve the signature check

Skazzino:
> 
> 
> The bot should approve the signature check

Hi @feliperuhland,
nothing seems to happen. What should i do?

feliperuhland:
Hi @Skazzino 

My suggestion is to clean the past commits and sign those that are going to stay.
If you need help with that, let me know.

Skazzino:
Hi,
some months passed since last reply on this PR. Is it ok or should we do something else?

PR #2485, opened 2020-01-03
Fix ImageCollectionTest.test_pull_multiple flakiness

thaJeztah:
relates to https://github.com/moby/moby/pull/40340 (https://github.com/moby/moby/pull/40340#issuecomment-570578753)
relates to https://github.com/moby/moby/pull/40343
introduced in https://github.com/docker/docker-py/pull/1883

The ImageCollectionTest.test_pull_multiple test performs a `docker pull` without
a `:tag` specified) to pull all tags of the given repository (image).

After pulling the image, the image(s) pulled are checked to verify if the list
of images contains the `:latest` tag.

However, the test assumes that all tags of the image are tags for the same
version of the image (same digest), and thus a *single* image is returned, which
is not always the case.

Currently, the `hello-world:latest` and `hello-world:linux` tags point to a
different digest, therefore the `client.images.pull()` returns multiple images:
one image for digest, making the test fail:

```
    ____________________ ImageCollectionTest.test_pull_multiple ____________________
    tests/integration/models_images_test.py:90: in test_pull_multiple
        assert len(images) == 1
    E   AssertionError: assert 2 == 1
    E    +  where 2 = len([<Image: 'hello-world:linux'>, <Image: 'hello-world:latest'>])
```

This patch updates the test to not assume a single image is returned, and instead
loop through the list of images and check if any of the images contains the
`:latest` tag.

### DIFF
--- a/tests/integration/models_images_test.py
+++ b/tests/integration/models_images_test.py
@@ -87,8 +87,10 @@ class ImageCollectionTest(BaseIntegrationTest):
     def test_pull_multiple(self):
         client = docker.from_env(version=TEST_API_VERSION)
         images = client.images.pull('hello-world')
-        assert len(images) == 1
-        assert 'hello-world:latest' in images[0].attrs['RepoTags']
+        assert len(images) >= 1
+        assert any([
+            'hello-world:latest' in img.attrs['RepoTags'] for img in images
+        ])
 
     def test_load_error(self):
         client = docker.from_env(version=TEST_API_VERSION)


thaJeztah:
ping @ulyssessouza @shin- ptal

thaJeztah:
Ah booh. It's failing. Not sure what I did wrong 🤔 

```

____________________ ImageCollectionTest.test_pull_multiple ____________________
tests/integration/models_images_test.py:97: in test_pull_multiple
    assert found
E   AssertionError: assert False
```

tests/integration/models_images_test.py
```diff
@@ -87,8 +87,14 @@ def test_pull_with_sha(self):
     def test_pull_multiple(self):
         client = docker.from_env(version=TEST_API_VERSION)
         images = client.images.pull('hello-world')
-        assert len(images) == 1
-        assert 'hello-world:latest' in images[0].attrs['RepoTags']
+        assert len(images) >= 1
+        found = False
+        for img in images:
+            if 'hello-world:latest' == img.attrs['RepoTags']:
```
feliperuhland: Hi @thaJeztah 

Could you try use unicode for image name, like `u'hello-world:latest'`

thaJeztah: Sure, I can try that (would that be a bit odd though, given that image references only support a-z0-9 ? 🤔 or does python return it as unicode? (I'm not really a Python expert 😅)

feliperuhland: Python 2 doesn't use Unicode as default.

Also, I forgot that `img.attrs['RepoTags']` is a list, so you can change the comparator `==` to `in` and the test will pass.

thaJeztah: Doh! You're right, completely forgot about that; yes, that should be the problem

feliperuhland: Yeah, Unicode wasn't that bad guy this time. hahaha

thaJeztah:
all green now! thanks for your help (and review) @feliperuhland 🤗

tests/integration/models_images_test.py
```diff
@@ -87,8 +87,14 @@ def test_pull_with_sha(self):
     def test_pull_multiple(self):
         client = docker.from_env(version=TEST_API_VERSION)
         images = client.images.pull('hello-world')
-        assert len(images) == 1
-        assert 'hello-world:latest' in images[0].attrs['RepoTags']
+        assert len(images) >= 1
+        found = False
+        for img in images:
```
zappy-shu: Could you use `any()` instead of doing a for loop. E.g. something like `assert any('hello-world:latest' in img[0].attrs['RepoTags'] for img in images)`?

thaJeztah: Ah, yes, sure. Didn't know of that approach 🤗

feliperuhland: It's just `img.attrs`, not `img[0].attrs`.

thaJeztah:
@zappy-shu @ulyssessouza @feliperuhland updated; ptal

feliperuhland:
@thaJeztah looks great
:tada:

PR #2486, opened 2020-01-06
tar entry size must be computed from utf-8 encoded content

ndeloof:
See https://github.com/docker/compose/issues/7112


milas:
This was fixed by #2553 and is in v4.4.0+

PR #2471, opened 2019-11-29
Add device requests

Lucidiot:
Another take on #2395: while the Docker CLI adds a `--gpus` parameter, it is in fact syntatic sugar for the `DeviceRequests` argument in a container's host config. This PR adds support for any kind of device request, as it seems the API was designed to handle more in the future than just GPUs.

It would still be possible to support another `gpus=...` argument in a way similar to #2465, that could handle the previous `nvidia` runtime and set the environment variables itself or use the new device requests.

For example, with NVIDIA's test image, the equivalent of `--gpus all` would be:

``` python
client.containers.run(
    'nvidia/cuda:9.0-base',
    'nvidia-smi',
    device_requests=[
        docker.types.DeviceRequest(count=-1, capabilities=[['gpu']])
    ]
)
```

The equivalent of `--gpus 2` would be `DeviceRequest(count=2, capabilities=[['gpu']])`.

And for something like `--gpus device=GPU-c6e298ba-aa33-4083-8ae3-db4e27037475,driver=nvidia,capabilities="utility,display"`:

``` python
client.containers.run(
    'nvidia/cuda:9.0-base',
    'nvidia-smi',
    device_requests=[
        docker.types.DeviceRequest(
            driver='nvidia',
            device_ids=['GPU-c6e298ba-aa33-4083-8ae3-db4e27037475'],
            capabilities=[
                ['gpu', 'utility', 'display']
            ]
        )
    ]
)
```

PR #2487, opened 2020-01-06
support HostConfig.DeviceRequests introduced by API 1.40

ndeloof:
required so we can investigate https://github.com/docker/compose/issues/6691


zajozorPho:
https://github.com/docker/docker-py/pull/2471/files seems to do the same and also includes tests?

ndeloof:
indeed, should have searched for existing PR on this topic before

PR #2511, opened 2020-02-11
raise an error for binding specific ports in 'host' mode of network

daeseokyoun:
The binding ports are ignored where the network mode is 'host'.
It could be a problem in case of using these options together on Mac or Windows OS. Because the [limitation](https://docs.docker.com/docker-for-mac/networking/)  that could not use the 'host' in network_mode on Mac and Windows. 
When 'host' mode is set on network_mode, the specific ports in 'ports' are ignored so the network is not able to be accessed through defined ports by developer.

### DIFF
--- a/docker/api/container.py
+++ b/docker/api/container.py
@@ -523,6 +523,10 @@ class ContainerApiMixin(object):
                 - ``container:<name|id>`` Reuse another container's network
                   stack.
                 - ``host`` Use the host network stack.
+                  This mode is incompatible with ``port_bindings``.
+                  If ``host`` is used as network_mode, all of listed up to
+                  ``port_bindings``` are ignored in running container.
+
             oom_kill_disable (bool): Whether to disable OOM killer.
             oom_score_adj (int): An integer value containing the score given
                 to the container in order to tune OOM killer preferences.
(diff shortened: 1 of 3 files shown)


daeseokyoun:
@ulyssessouza @shin- 
Do I need to create an issue related to this PR?
And I am not able to solve failure in jenkins(pr-merge) job.

aiordache (approved):
SGTM. Thank you!

docker/api/container.py
```diff
@@ -520,6 +520,10 @@ def create_host_config(self, *args, **kwargs):
                 - ``container:<name|id>`` Reuse another container's network
                   stack.
                 - ``host`` Use the host network stack.
+                  This mode is incompatible with ``port_bindings``.
+                  If ``host`` is used as network_mode, all of listed up to
```
gtardif, line 524: port bindings are not ignored in this case, the 2 options are not compatible (raising an error)

gtardif: Same in following doc updates

gtardif (commented):
Just rephrase doc to mention incompatible options rather than values possibly ignored when it will actually raise errors. Otherwise LGTM

docker/api/container.py
daeseokyoun: Yes, right. the ```host``` mode is not abled to use with ```port_bindings```. I will update docs.

daeseokyoun:
@aiordache @gtardif this PR is just rebased on recent master branch and corrected doc followed by gtardif's review comment.

Please have a look at the updated code. Thanks.

gtardif (approved):

PR #2512, opened 2020-02-11
Update default API version to v1.39

thaJeztah:
When running the docker-py integration tests in the Moby repository, some
tests were skipped because the API version used was too low:

    SKIPPED [1] tests/integration/api_service_test.py:882: API version is too low (< 1.38)
    SKIPPED [1] tests/integration/api_swarm_test.py:59: API version is too low (< 1.39)
    SKIPPED [1] tests/integration/api_swarm_test.py:38: API version is too low (< 1.39)
    SKIPPED [1] tests/integration/api_swarm_test.py:45: API version is too low (< 1.39)
    SKIPPED [1] tests/integration/api_swarm_test.py:52: API version is too low (< 1.39)

While it's possible to override the API version to use for testing
using the `DOCKER_TEST_API_VERSION` environment variable, we may want
to set the default to a version that supports all features that were
added.

This patch updates the default API version to v1.39, which is the minimum
version required for those features, and corresponds with Docker 18.09.

Note that the API version of the current (19.03) Docker release is v1.40,
but using that version as default would exclude users that did not update
their Docker version yet (and would not be needed yet for the features provided).


thaJeztah:
@ulyssessouza @chris-crone @shin- PTAL

Also wondering what the policy is on updating this default; is this something to include in a patch-release ?

chris-crone:
As a straw man, I would say that we should match the default API version with the latest stable release. We can bump the minor version of the Python SDK when we bump this version.

thaJeztah:
The problem likely is that (afaik) `docker-py` doesn't do API version negotiation with the daemon, so if the daemon is older then 18.09, it will fail. Not exactly sure how the package is used though, so perhaps consumers if this package also set the API version (and thus would it set to the version they need?)

thaJeztah:
@ulyssessouza rebased this one

PR #2517, opened 2020-02-19
Fix secret repr

feliperuhland:
This PR fixes issue #2025

### DIFF
--- a/docker/models/secrets.py
+++ b/docker/models/secrets.py
@@ -4,14 +4,22 @@ from .resource import Model, Collection
 
 class Secret(Model):
     """A secret."""
-    id_attribute = 'ID'
+    id_attribute = 'Id'
 
     def __repr__(self):
         return "<%s: '%s'>" % (self.__class__.__name__, self.name)
 
     @property
     def name(self):
-        return self.attrs['Spec']['Name']
+        """
+        Create secret only returns Id field, so if the name field is not
+        present, it will show id value instead.
+        Issue #2025
+        """
+        if self.attrs.get('Spec', {}).get('Name'):
+            return self.attrs['Spec']['Name']
+
+        return self.id
 
     def remove(self):
         """
--- a/docker/models/secrets.py
+++ b/docker/models/secrets.py
@@ -7,19 +7,11 @@ class Secret(Model):
     id_attribute = 'Id'
 
     def __repr__(self):
-        return "<%s: '%s'>" % (self.__class__.__name__, self.name)
+        return "<%s: '%s'>" % (self.__class__.__name__, self.id)
 
     @property
     def name(self):
-        """
-        Create secret only returns Id field, so if the name field is not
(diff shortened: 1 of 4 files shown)


feliperuhland:
@shin- @ulyssessouza 

Could you take a look?

Thanks.

docker/models/secrets.py
```diff
@@ -4,14 +4,14 @@
 
 class Secret(Model):
     """A secret."""
-    id_attribute = 'ID'
+    id_attribute = 'Id'
```
ulyssessouza: I'm not sure to understand this change. Can you explain why?

docker/models/secrets.py
```diff
@@ -4,14 +4,14 @@
...
 
     def __repr__(self):
-        return "<%s: '%s'>" % (self.__class__.__name__, self.name)
+        return "<%s: '%s'>" % (self.__class__.__name__, self.id)
```
ulyssessouza: Why changing the attribute here?

docker/models/secrets.py
feliperuhland: I don't remember why I did it, but it's fixed now.
Thanks.

docker/models/secrets.py
feliperuhland: It's because [List Secrets](https://docs.docker.com/engine/api/v1.30/#operation/SecretList) doesn't return the name attribute.

Issue #2025

ulyssessouza: Well... 
The API says it does. That's in the very link you mentioned. If not, it's a bug

thaJeztah: Name is nested in the `Spec` in the response, so not at the same level as `id`;

```json
[
  {
    "ID": "ktnbjxoalbkvbvedmg1urrz8h",
    "Version": {
      "Index": 11
    },
    "CreatedAt": "2016-11-05T01:20:17.327670065Z",
    "UpdatedAt": "2016-11-05T01:20:17.327670065Z",
    "Spec": {
      "Name": "app-dev.crt"
    }
  }
]
```

Screenshot 2020-02-28 at 17 01 01

feliperuhland: Hi @ulyssessouza @thaJeztah the [secret creation endpoint](https://docs.docker.com/engine/api/v1.41/#operation/SecretCreate)  just return the `ID` value, so the method `__repr__` raise an exception:

```python
     12     @property
     13     def name(self):
---> 14         return self.attrs['Spec']['Name']
     15 
     16     def remove(self):

KeyError: 'Spec'
```

image

feliperuhland:
Hi @ulyssessouza @thaJeztah @shin- 

Could you please take a look?
Thanks!

feliperuhland:
Hi @ulyssessouza @thaJeztah @shin- @aiordache 
Could you please take a look?

Thanks!

PR #2519, opened 2020-02-22
Open .dockerignore as utf-8

carlbordum:
This should solve https://github.com/docker/compose/issues/7236

### DIFF
--- a/docker/api/build.py
+++ b/docker/api/build.py
@@ -1,3 +1,4 @@
+import io
 import json
 import logging
 import os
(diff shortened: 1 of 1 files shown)


milas:
We now only support Python3 and the file will be opened using the default system encoding:

> The default encoding is platform dependent (whatever [locale.getpreferredencoding()](https://docs.python.org/3/library/locale.html#locale.getpreferredencoding) returns)

https://docs.python.org/3/library/functions.html#open

That said, I think that `.dockerignore` files are likely implicitly UTF-8 due to the Docker tooling being primarily written in Go, in which strings are kinda sorta UTF-8 if you don't think too hard. So, I'd be willing to accept an updated PR for Python3 for this if it's still causing someone trouble.

carlbordum:
I don't think this is a Python 2/3 issue, but an issue because Windows default encoding is not UTF-8.

PR #2520, opened 2020-02-26
Disable compression by default when using container:get_archive method

Nicceboy:
Some pre-discussion can be found from here: https://github.com/moby/moby/issues/40577

This PR attempts to disable compression by default and parametrize the encoding of data stream, when using container's get_archive method. This should bring similar performance than command `docker cp` has when copying files from container as tar file into host system.

Current implementation is not giving any headers for this GET request, and this causes the use of default headers of `requests` library: [`Accept-Encoding: gzip, deflate`](https://github.com/psf/requests/blob/ca6f9af5dba09591007b15a7368bc0f006b7cc50/requests/utils.py#L807). 

This makes Docker engine to encode response stream as gzip-compressed format, and this slows download speed a lot.

I'm not sure if giving possibility for parameter is that useful, but maybe there are some use cases where network bandwidth is bottleneck, and compression could be helpful.

### DIFF
--- a/docker/api/container.py
+++ b/docker/api/container.py
@@ -694,7 +694,8 @@ class ContainerApiMixin(object):
         return self._stream_raw_result(res, chunk_size, False)
 
     @utils.check_resource('container')
-    def get_archive(self, container, path, chunk_size=DEFAULT_DATA_CHUNK_SIZE):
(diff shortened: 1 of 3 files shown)


thaJeztah:
Thanks! @ulyssessouza @shin- @ndeloof PTAL

Nicceboy:
Any updates about this?

thaJeztah:
ping @ulyssessouza

PR #2534, opened 2020-03-20
Support OpenSSH Identityfile option

### LINKED ISSUE #2533: Support OpenSSH's IdentityFile config option
# Description
As a developer I want `docker-py` to load the identity file according to my setting specified in `~/.ssh/config`, not only the "default" `id_rsa` file.

# Example
I have a `~/.ssh/config` file with contents:
```ssh
Host myHost
    User myUser
    HostName myIp
    IdentityFile C:/Users/me/.ssh/id_rsa_custom_file
    IdentitiesOnly yes
```

Now I would like `docker-py` to pick up `C:/Users/me/.ssh/id_rsa_custom_file` for trying to connect to the remote host (For example when I want to run `docker-compose -H "ssh://myHost" ps`. However right now it does not.

When I rename `C:/Users/me/.ssh/id_rsa_custom_file` to `C:/Users/me/.ssh/id_rsa` it works just fine. However this is not an option for me since I have multiple identity files and hosts that I want to use.

# Proposal
Add something like this to the [`sshcon.py`](https://github.com/docker/docker-py/blob/c285bee1bc59f6b2d65cee952b5522c88047a3bc/docker/transport/sshconn.py#L105):
```python
if 'identityfile' in host_config:
    self.ssh_params['key_filename '] = host_config['identityfile']
```

HaaLeo:
Fixes #2533

### DIFF
--- a/docker/transport/sshconn.py
+++ b/docker/transport/sshconn.py
@@ -100,6 +100,8 @@ class SSHHTTPAdapter(BaseHTTPAdapter):
                 )
             if 'hostname' in host_config:
                 self.ssh_params['hostname'] = host_config['hostname']
+            if 'identityfile' in host_config:
+                self.ssh_params['key_filename'] = host_config['identityfile']
             if base_url.port is None and 'port' in host_config:
                 self.ssh_params['port'] = self.ssh_conf['port']
             if base_url.username is None and 'user' in host_config:


ulyssessouza:
@HaaLeo Thank you for the PR.
Could you please add tests?

HaaLeo:
@ulyssessouza I could not find any existing tests for the `sshconn.py` file. Am I missing something? Could you please advise where a starting point to add unit tests is?

ulyssessouza:
@HaaLeo Looks quite complicated to test that (as integration test)... I think that's good enough
LGTM

PR #2549, opened 2020-04-17
Fix tests to support both log plugin feedbacks

wpjunior:
I am improving the feedback when docker uses a log driver, could you see the change at: https://github.com/moby/moby/pull/40807.

After above change I broke one test that is inside this project, then, I am sending the fix.

Thanks.

### DIFF
--- a/tests/integration/api_container_test.py
+++ b/tests/integration/api_container_test.py
@@ -273,11 +273,14 @@ class CreateContainerTest(BaseAPIIntegrationTest):
 
     def test_invalid_log_driver_raises_exception(self):
         log_config = docker.types.LogConfig(
-            type='asdf-nope',
+            type='asdf',
             config={}
         )
 
-        expected_msg = "logger: no log driver named 'asdf-nope' is registered"
+        expected_msgs = [
+            "logger: no log driver named 'asdf' is registered",
+            "looking up logging plugin asdf: plugin \"asdf\" not found",
+        ]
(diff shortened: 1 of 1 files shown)


thaJeztah:
Looks like tests are failing on something else (unrelated); @ulyssessouza could you have a look?

ulyssessouza:
Looks like it was just a flaky test... The only thing missing is to have the `dco-signed`

wpjunior:
Hi @ulyssessouza how can I complete dco-signed check ?, I signed my commits with related signature and it is not work.

thaJeztah:
I think only the first one is supported (the `Docker-DCO-1.1-Signed-off-by:` was something from the past, and the `"DCO"` bot won't understand it.

Can you remove the last (`Docker-DCO...`) line? I think after that it should be correct

thaJeztah:
Oh! Besides that, @ulyssessouza I see you enabled the `required` option for the DCO check; unfortunately that doesn't work; there's a bug in GitHub's handling of those checks; if you add it to a repository that has open PR's and you set it to "required", it never completes (gets stuck in "pending").

If you disable the "required" check, things should pass

thaJeztah:
Unfortunately, I don't have write/admin access on this repository otherwise I'd have updated that setting, but @wpjunior if you can update the commit message, then things should be OK from your side 🤗

ulyssessouza:
Got it merged!

@wpjunior Obrigado pela PR!

PR #2553, opened 2020-04-28
fix docker build error when dockerfile contains unicode character.

fengbaolong:
if dockerfile contains unicode character,len(contents) will return character length,this length will less than len(contents_encoded) length,so contants data will be truncated.


aiordache (approved):
Thank you!

gtardif (approved):

PR #2567, opened 2020-05-18
copy deprecated splitnport from stdlib Fixes #2566

### LINKED ISSUE #2566: DeprecationWarning: urllib.parse.splitnport() is deprecated as of 3.8, use urllib.parse.urlparse() instead
https://github.com/docker/docker-py/blob/9a24df5cdd03c679dc929735e4766e19ff1c2bdb/docker/utils/utils.py#L23

graingert:


docker/utils/utils.py
```diff
@@ -207,6 +202,27 @@ def parse_repository_tag(repo_name):
...
 
 
+def _splitnport(host, defport=-1):
+    """
+    Split host and port, returning numeric port.
+    Return given default port if no ':' found; defaults to -1.
+    Return numerical port if a valid number are found after ':'.
+    Return None if ':' but not a valid number.
+    
+    From https://github.com/python/cpython/blob/3.8/Lib/urllib/parse.py
```
graingert, line 212: is this code long enough to need a licence declaration?

docker/utils/utils.py
```diff
@@ -207,6 +202,27 @@ def parse_repository_tag(repo_name):
     return repo_name, None
 
 
+def _splitnport(host, defport=-1):
+    """
+    Split host and port, returning numeric port.
+    Return given default port if no ':' found; defaults to -1.
+    Return numerical port if a valid number are found after ':'.
+    Return None if ':' but not a valid number.
+
```
graingert, line 211:
```suggestion

```

docker/utils/utils.py
```diff
@@ -207,6 +202,26 @@ def parse_repository_tag(repo_name):
     return repo_name, None
 
 
+def _splitnport(host, defport=-1):
+    """
+    Split host and port, returning numeric port.
+    Return given default port if no ':' found; defaults to -1.
+    Return numerical port if a valid number are found after ':'.
+    Return None if ':' but not a valid number.
```
graingert, line 210:
```suggestion
    Return None if ':' but not a valid number.

```

docker/utils/utils.py
thaJeztah: No really familiar with the Python license, but reading https://github.com/python/cpython/blob/v3.8.3/LICENSE#L74-L84

> provided, however, that PSF's License Agreement and PSF's notice of copyright,
i.e., "Copyright (c) 2001, 2002, 2003, 2004, 2005, 2006, 2007, 2008, 2009, 2010,
2011, 2012, 2013, 2014, 2015, 2016, 2017, 2018, 2019, 2020 Python Software Foundation;
All Rights Reserved" are retained in Python alone or in any derivative version
prepared by Licensee.

Perhaps it should be mentioned, just to be on the safe side (I _guess_ using a range for the years to keep it a bit shorter would be ok though) 🤔 

It's also better to link to a _tag_ here instead of a release-branch (release branches could potentially be removed, which is less likely for tags). Perhaps link to the actual lines while doing so;


```python
    Based on https://github.com/python/cpython/blob/v3.8.3/Lib/urllib/parse.py#L1094-L1108
    Copyright (c) 2001-2020 Python Software Foundation; All Rights Reserved
```

thaJeztah:
@ulyssessouza PTAL

docker/utils/utils.py
graingert: @thaJeztah can you make this as a GitHub `` ```suggestion ``?

thaJeztah: Oh, I avoided doing that, because overall we prefer changes to be grouped in logical commits (so no "touch-up" commits if not needed) 😅

milas:
Hi! Sorry we never got this PR over the finish line before 😓 

I'm going to close this one because unfortunately there's been a lot of repo churn since this was opened, so it's no longer valid with CI, and additionally, there were actually other bugs as a result of our usage of `splitnport`. A fix should be forthcoming in #3004 

Thanks for your contribution and time!

PR #2584, opened 2020-06-03
Update test engine version to 19.03.12

thaJeztah:

PR #2585, opened 2020-06-03
Jenkinsfile: update node selection labels

thaJeztah:
Make sure we use the LTS nodes, to prevent using machines that
we prepared with cgroups v2 (which is not yet supported by docker v19.03)

relates to https://github.com/docker/docker-py/pull/2584#discussion_r434402870


thaJeztah:
ping @ulyssessouza @StefanScherer ptal

ulyssessouza:
Thanks @thaJeztah !

PR #2586, opened 2020-06-03
Uses a slimmer version of python

edumco:
### Update python image to the slim variant.

The slim variant is much smaller and has a smaller attack surface.


Dockerfile
```diff
@@ -1,4 +1,4 @@
-ARG PYTHON_VERSION=2.7
+ARG PYTHON_VERSION=2.7-slim
 
 FROM python:${PYTHON_VERSION}
```
thaJeztah, line 3: I think including it in the `ARG` could break some uses (e.g., if CI specifies the python version);

Could you instead add it here?

```suggestion
FROM python:${PYTHON_VERSION}-slim
```

Looks like the python-3 version of the Dockerfile also needs the same change; https://github.com/docker/docker-py/blob/master/Dockerfile-py3#L3

(also wondering if we should remove the Python 2 variant, as it reached EOL

edumco:
> left some suggestions, but looks like your commit is missing a DCO sign-off in the commit message; https://github.com/moby/moby/blob/master/CONTRIBUTING.md#sign-your-work

I will close this one and open a new with the signe

Issue #2601, opened 2020-06-25
Fix context load for non-docker endpoints


aiordache:
Docker-py gives an error when trying to load an aci context as it requires fields as Host and SkipTLSVerify to be set.

Update context load to skip parsing non-docker context endpoints.

PR #2602, opened 2020-06-25
Fix context load for non-docker endpoints

aiordache:
Skip non-docker endpoint parsing and add fallbacks for docker endpoint fields.

closes https://github.com/docker/docker-py/issues/2601

### DIFF
--- a/docker/context/context.py
+++ b/docker/context/context.py
@@ -16,30 +16,42 @@ class Context:
         if not name:
             raise Exception("Name not provided")
         self.name = name
+        self.context_type = None
         self.orchestrator = orchestrator
+        self.endpoints = {}
+        self.tls_cfg = {}
+        self.meta_path = "IN MEMORY"
+        self.tls_path = "IN MEMORY"
+
         if not endpoints:
+            # set default docker endpoint if no endpoint is set
             default_endpoint = "docker" if (
                 not orchestrator or orchestrator == "swarm"
                 ) else orchestrator
+
             self.endpoints = {
                 default_endpoint: {
                     "Host": get_context_host(host, tls),
                     "SkipTLSVerify": not tls
                 }
             }
-        else:
-            for k, v in endpoints.items():
-                ekeys = v.keys()
-                for param in ["Host", "SkipTLSVerify"]:
-                    if param not in ekeys:
-                        raise ContextException(
-                            "Missing parameter {} from endpoint {}".format(
-                                param, k))
-            self.endpoints = endpoints
+            return
 
-        self.tls_cfg = {}
-        self.meta_path = "IN MEMORY"
-        self.tls_path = "IN MEMORY"
+        # check docker endpoints
+        for k, v in endpoints.items():
+            if not isinstance(v, dict):
+                # unknown format
+                raise ContextException("""Unknown endpoint format for
+                    context {}: {}""".format(name, v))
+
+            self.endpoints[k] = v
+            if k != "docker":
+                continue
+
+            self.endpoints[k]["Host"] = v.get("Host", get_context_host(
+                host, tls))
+            self.endpoints[k]["SkipTLSVerify"] = bool(v.get(
+                "SkipTLSVerify", not tls))
 
     def set_endpoint(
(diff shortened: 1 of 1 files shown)


.vscode/launch.json
```diff
@@ -0,0 +1,15 @@
+{
+    // Use IntelliSense to learn about possible attributes.
+    // Hover to view descriptions of existing attributes.
+    // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
```
thaJeztah, line 4: Don't think these should have been included?

docker/api/build.py
```diff
@@ -153,7 +153,7 @@ def build(self, path=None, tag=None, quiet=False, fileobj=None,
                 with open(dockerignore, 'r') as f:
                     exclude = list(filter(
                         lambda x: x != '' and x[0] != '#',
-                        [l.strip() for l in f.read().splitlines()]
+                        [line.strip() for line in f.read().splitlines()]
```
thaJeztah, line 156: unrelated change?

docker/context/context.py
```diff
@@ -16,30 +16,42 @@ def __init__(self, name, orchestrator=None, host=None, endpoints=None,
...
+        # check docker endpoints
+        for k, v in endpoints.items():
+            if not isinstance(v, dict):
+                # unknown format
+                raise ContextException("""Unknown endpoint format for
+                    context {}: {}""".format(name, v))
+
+            self.endpoints[k] = v
+            if k != "docker":
+                continue
```
thaJeztah, line 49: I wonder if `docker-py` should ignore these, or just load them. I'm not too familiar with the code, buta context has _some_ (semi-)strict fields, and other metadata that could be "anything". I think it should be ok to _load_ these, as long as they're not used to try to _connect_ with them.

(basically; if this library is just to _parse_ / _load_ contexts, it should not be too opinionated)

 WDYT?

docker/context/context.py
```diff
@@ -16,30 +16,42 @@ def __init__(self, name, orchestrator=None, host=None, endpoints=None,
...
+
+            self.endpoints[k]["Host"] = v.get("Host", get_context_host(
+                host, tls))
+            self.endpoints[k]["SkipTLSVerify"] = bool(v.get(
+                "SkipTLSVerify", not tls))
```
thaJeztah, line 54: I think this was the part where the problem occurred; would it be possible to ignore missing fields? (and use the default for those)? So if `SkipTLSVerify` is not present, then keep the default (`false`)

scripts/versions.py
```diff
@@ -72,5 +72,6 @@ def main():
         results.add(str(latest))
     print(' '.join(results))
 
+
```
thaJeztah, line 75: unrelated change

.vscode/launch.json
aiordache: Yes, my mistake. Made the commit through vscode and didn't notice these files were also staged. Thanks!

docker/context/context.py
aiordache: They are loaded,  for docker endpoints we only verify that we have a Host and SkipTLSVerify fields set (otherwise we set some defaults).  For cloud endpoints we don't check anything.

thaJeztah:
(oh, and pls squash the commits at the end 😅)

docker/context/context.py
aiordache: We do this with the **v.get()**. The 2nd parameter is the fallback value if the key is no found. The SkipTLSVerify fallback value depends on the specified tls certs(it should be false if we set certs otherwise True). For the host we need to detect the platform and set the path to the docker engine according to it.

thaJeztah: > We do this with the v.get(). The 2nd parameter is the fallback value if the key is no found. 

Oh! See yes, the old code used `bool(v["SkipTLSVerify"])` and that caused the error

docker/context/context.py
thaJeztah: 🤦 I see it now, yes. I read the `continue` as "skip this endpoint", but it's loaded the line above, and only the special handling is skipped

PR #2621, opened 2020-07-10
Correct example in Image save methods docstr

jbn:
The example incorrectly tries to iterate over the
image object. This commit fixes it so that it calls
the save method and iterates over the generator.


milas:
Fixed in #2570

Issue #2628, opened 2020-07-17
Username shouldn't be required for login when using dockercfg_path for docker client login.


anorth2:
The auth included in the docker config includes both the username and password for the registry. As a result the username should not be required.

Additionally, these should not work:

```
docker_client.login(username="my-fake-username-does-not-exist", dockercfg_path="/Users/myusername/.docker/config.json")

docker_client.login(username="", dockercfg_path="/Users/myusername/.docker/config.json")
```

yet they do:
`{'IdentityToken': '', 'Status': 'Login Succeeded'}`

If `username` is required then it should actually be _required_

PR #2629, opened 2020-07-17
Update to allow using only docker config to login

anorth2:
No longer require a username be provided when the docker config will contain the username already.


anorth2:
partially resolves https://github.com/docker/docker-py/issues/2628

Issue #2635, opened 2020-08-05
Question - the container.create or run API does not pull ?


sumukhs:
Based on the documentation, I expected that container create API would pull the image. However, I get a failure exception with ImageNotFound.

Is there a way to get the CLI behavior to pull if the image doesn't exist?

Thank you!

dan-hipschman-od:
Are you looking at the docker CLI documentation for `docker create` and assuming the Python SDK `containers.create` does the same thing? The Python SDK does document the `ImageNotFound` exception: https://github.com/docker/docker-py/blob/b37a25d70b7d483cc0fbc78fb6863636ca84cbb3/docker/models/containers.py#L846-L858

I agree it would be nice if they did the same thing. That change might break existing usage that relies on the current documentation, but perhaps the maintainers would be open to it.

`containers.run` does pull the image if it's missing.

PR #2636, opened 2020-08-08
Fix pulling with container create api

navid9675:
Fix the pulling issue stated in #2635. Basically, similar to `docker
create` it should pull the image if it does not exist locally.


dan-hipschman-od:
I would like to see the Python SDK and Docker CLI behavior in sync, so support this PR, but should point out that the current difference is documented: https://github.com/docker/docker-py/blob/b37a25d70b7d483cc0fbc78fb6863636ca84cbb3/docker/models/containers.py#L846-L858

This PR should at least update the documentation and `ContainerCollection.run` (which can call `create` twice). Then the maintainers can decide whether they want to change the documented behavior.

BeardedDonut:
@dan-hipschman-od Thanks for your comment, I just didn't get this part would you please clarify a bit.

> This PR should at least update the documentation and ContainerCollection.run (which can call create twice). Then the maintainers can decide whether they want to change the documented behavior.

Thanks.

dan-hipschman-od:
Yea, I was just saying two things:
* Could you update the documentation of `ContainerCollection.create` so it says it will try to pull the image if it doesn't exist locally?
* And `ContainerCollection.run` currently calls `create` and assumes it will not pull the image, so `run` itself catches `ImageNotFound` and tries to pull the image (see below). Since you're changing `create` to pull the image, then `run` can be simplified:
https://github.com/docker/docker-py/blob/b37a25d70b7d483cc0fbc78fb6863636ca84cbb3/docker/models/containers.py#L808-L814

milas (changes requested):
Thanks for the PR!

I'm going to echo the great feedback already provided by @dan-hipschman-od:

> Yea, I was just saying two things:
> 
> * Could you update the documentation of `ContainerCollection.create` so it says it will try to pull the image if it doesn't exist locally?
> * And `ContainerCollection.run` currently calls `create` and assumes it will not pull the image, so `run` itself catches `ImageNotFound` and tries to pull the image (see below). Since you're changing `create` to pull the image, then `run` can be simplified:
>   https://github.com/docker/docker-py/blob/b37a25d70b7d483cc0fbc78fb6863636ca84cbb3/docker/models/containers.py#L808-L814

PR #2640, opened 2020-08-11
fixing docstring to match behaviour

nadavwe:
i think this might have been missed in #2512.

(on a side note,  I would love to see the default being "auto", as for basic usage, I don't really care about the version of docker).

Thanks!


thaJeztah:
Thanks! Yes, looks like I missed these 😅 

> on a side note, I would love to see the default being "auto", as for basic usage, I don't really care about the version of docker


Agreed! I'm not really a python coder myself, so haven't looked into how it would be implemented in Python (but contributions would be welcome).

it could use the same logic (could still be  configurable to enable/disable it), as https://github.com/moby/moby/blob/ecdb0b22393bb669325099320d26d18687425e5f/client/client.go#L200-L246

(and https://github.com/moby/moby/pull/39032, which handles the API version negotiation the moment a request will be made)

thaJeztah:
Oh! I see you probably made the changes through GitHub's web UI, and didn't add a DCO-sign-off in the commit message (see https://github.com/moby/moby/blob/master/CONTRIBUTING.md#sign-your-work).

It's possible to amend your commit-message to fix that, but you'll have to do that from the command-line (unfortunately can't be done through GitHub's web UI 😞)

Let me know if you need help doing so, then I can post some instructions if needed.

nadavwe:
fixed the dco.
looks like the `auto` thing already has some support, no need to do anything...

thaJeztah:
Ah, you're right; looks like there's something in place in https://github.com/docker/docker-py/blob/4d7d4084138fa6161ef5f31d410b0d326d41f777/docker/api/client.py#L188-L216


Looks like it's currently calling the `/version` endpoint. For recent versions of the daemon, using the `/_ping` endpoint may be slightly more lightweight for that (https://docs.docker.com/engine/api/v1.40/#operation/SystemPingHead), but not critical I guess.

PR #2647, opened 2020-08-19
auth_config build function support

Chuseuiti:
The goal of this PR is to allow the build function to support images relying on private registries on as their base docker image layers, as in:

FROM private_image


Chuseuiti:
Current error on CI seems to be unrelated to the PR as it is due to flake8.

Chuseuiti:
On a side note, I ran autopep8 to deal with all the blank spaces but I wonder if this should be done automatically by the automation. Attached the command: 
`    autopep8 --select=W293 --in-place your_file.py`

Chuseuiti:
On my comment on automated styling there are some Github tools that may help
 https://github.com/marketplace/restyled-io

docker/models/images.py
```diff
@@ -327,7 +335,7 @@ def get_registry_data(self, name, auth_config=None):
 
         Returns:
             (:py:class:`RegistryData`): The data object.
-
+_url
```
ulyssessouza, line 338: typo?

tests/unit/api_image_test.py
```diff
@@ -14,6 +14,14 @@
     import mock
 
 
+class MockAnyArg(object):
+    def __eq__(a, b):
+        if 'X-Registry-Auth' in b.keys():
+            return True
+        else:
+            return False
```
ulyssessouza, line 22:
```suggestion
        return 'X-Registry-Auth' in b.keys():
```

tests/unit/models_images_test.py
```diff
@@ -13,14 +13,27 @@ def test_build(self):
...
         assert image.id == FAKE_IMAGE_ID
 
+    # Due to mocking on fake_api_client the result
+    # is not behaiving as expected,
+    # depending of feedback it will be refactored
+    # def test_build_with_auth_config(self):
+    #     client = make_fake_client()
+    #     image = client.images.build(auth_config={"auth":1})
+    #     client.api.build.assert_called_with()
+    #     client.api.inspect_image.assert_called_with(
```
ulyssessouza, line 27: Not sure why the call to `inspect_image` here.

docker/api/image.py
```diff
@@ -225,8 +225,19 @@ def import_image_from_image(self, image, repository=None, tag=None,
...
+            if header:
+                headers['X-Registry-Auth'] = header
+        else:
+            log.debug('Sending supplied auth config')
+            headers['X-Registry-Auth'] = auth.encode_header(auth_config)
+        return headers
+
     @utils.check_resource('image')
-    def inspect_image(self, image):
+    def inspect_image(self, image, auth_config=None):
```
ulyssessouza, line 240: Instead of passing `auth_config` on all the client methods, you could just create a setter for it in the client. 
That would also help you with the commented unit test when checking the `inspect`.

docker/models/images.py
```diff
@@ -262,14 +262,17 @@ def build(self, **kwargs):
...
+                ``username`` and ``password`` keys to be valid.
 
         Returns:
             (tuple): The first item is the :py:class:`Image` object for the
                 image that was build. The second item is a generator of the
                 build logs as JSON-decoded objects.
 
         Raises:
-            :py:class:`docker.errors.BuildError`
+            :py:class:`docker.errors.Bu_urlildError`
```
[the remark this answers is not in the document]
Chuseuiti: Fixed!

docker/models/images.py
Chuseuiti: Fixed!

tests/unit/api_image_test.py
Chuseuiti: also deleted extra colon on return

Issue #2639, opened 2020-08-10
Docker API support issue on version 4.3.0


darkwizard242:
# Issue:

It appears that docker-py (pip's package **docker**) version 1.4.0 is erroring out with the following:

```shell
Traceback (most recent call last):
  File "/home/travis/virtualenv/python3.8.1/lib/python3.8/site-packages/docker/api/client.py", line 261, in _raise_for_status
    response.raise_for_status()
  File "/home/travis/virtualenv/python3.8.1/lib/python3.8/site-packages/requests/models.py", line 941, in raise_for_status
    raise HTTPError(http_error_msg, response=self)
requests.exceptions.HTTPError: 400 Client Error: Bad Request for url: http+docker://localhost/v1.39/_ping
During handling of the above exception, another exception occurred:
Traceback (most recent call last):
  File "/home/travis/virtualenv/python3.8.1/bin/molecule", line 8, in <module>
    sys.exit(main())
  File "/home/travis/virtualenv/python3.8.1/lib/python3.8/site-packages/click/core.py", line 829, in __call__
    return self.main(*args, **kwargs)
  File "/home/travis/virtualenv/python3.8.1/lib/python3.8/site-packages/click/core.py", line 782, in main
    rv = self.invoke(ctx)
  File "/home/travis/virtualenv/python3.8.1/lib/python3.8/site-packages/click/core.py", line 1259, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/home/travis/virtualenv/python3.8.1/lib/python3.8/site-packages/click/core.py", line 1066, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/home/travis/virtualenv/python3.8.1/lib/python3.8/site-packages/click/core.py", line 610, in invoke
    return callback(*args, **kwargs)
  File "/home/travis/virtualenv/python3.8.1/lib/python3.8/site-packages/click/decorators.py", line 21, in new_func
    return f(get_current_context(), *args, **kwargs)
  File "/home/travis/virtualenv/python3.8.1/lib/python3.8/site-packages/molecule/command/test.py", line 147, in test
    base.execute_cmdline_scenarios(scenario_name, args, command_args)
  File "/home/travis/virtualenv/python3.8.1/lib/python3.8/site-packages/molecule/command/base.py", line 105, in execute_cmdline_scenarios
    execute_scenario(scenario)
[...]

PR #2650, opened 2020-08-19
Set default version to 'auto'

aiordache:
Set version to 'auto'  to avoid breaking on old engines not implementing v1.39.

Relates to https://github.com/docker/docker-py/issues/2639

### DIFF
--- a/docker/api/client.py
+++ b/docker/api/client.py
@@ -7,6 +7,19 @@ import requests.exceptions
 import six
 import websocket
 
+from .. import auth
+from ..constants import (DEFAULT_DOCKER_API_VERSION, DEFAULT_NUM_POOLS,
+                         DEFAULT_NUM_POOLS_SSH, DEFAULT_TIMEOUT_SECONDS,
+                         DEFAULT_USER_AGENT, IS_WINDOWS_PLATFORM,
+                         MINIMUM_DOCKER_API_VERSION, STREAM_HEADER_SIZE_BYTES)
+from ..errors import (DockerException, InvalidVersion, TLSParameterError,
+                      create_api_error_from_http_exception)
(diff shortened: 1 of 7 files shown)


docker/api/client.py
```diff
@@ -7,6 +7,19 @@
 import six
 import websocket
 
+from .. import auth
```
thaJeztah, line 10: where do these changed order of these imports come from? Was that intentional? If not needed, better to keep it separate and focus the PR on just the "auto" change

aiordache: My IDE is sorting imports on save. Could you ignore the reordering please, it is not bad having them sorted anyway.

ulyssessouza (approved):
LGTM

docker/api/client.py
thaJeztah: I see it's been merged, but it's really better to keep these changes separate (at _least_ a separate commit). In this commit, more than half of the line changes is now actually unrelated to the change itself.

I also see that the second commit was to fix the tests to adjust for the changes of the first commit. Order of the commits should have been reversed to not break `git bisect`

doi155:
Note for users: this PR may break your program.

On an environment without access to a docker daemon,
Previously:
```
client = docker.from_env()
client.version()  <--- will raise
```

After this PR,
```
client = docker.from_env()  <-- will raise
client.version()
```

If your program caches the client object at, for example, a global context without caring accessibility to the docker daemon, your program will no longer start. (My program does it so.)

Possible fixes:
* Stop caching the client object. Instantiate it just before use.
* Specify docker version.  `client  = docker.from_env(version="1.39")`.

The second one is quicker, but the first one may be better in light of the purpose of this PR.

Issue #2545, opened 2020-04-08
Bug or Misleading Docu: All docker images will be downloaded when the image tag is not specified


bonacci-johannes:
BUG
Setup: Python3.8: docker 4.2.0

When starting a clean postgres container using python docker sdk the docker sdk pulls all postgres images.

It would be helpful to update the docu mentioning that the version tag should be specified.
Or even more handy, introduce a default to the latest version?

Here is the code:
```python
import docker
from docker.models.containers import Container
client = docker.from_env()
container: Container = client.containers.run(
    image='postgres', remove=True, auto_remove=True, detach=True,
    ports={'5432': '5432'},
    environment=
    [
        'POSTGRES_PASSWORD=pwd',
        'POSTGRES_USER=test',
        'POSTGRES_DB=test'
    ])
```
Bash output of 'docker images' before running the code:
```shell
REPOSITORY          TAG                 IMAGE ID            CREATED             SIZE
```

Bash output of 'docker images' after running the code:
```shell
REPOSITORY          TAG                 IMAGE ID            CREATED             SIZE
postgres            10.12               1fb929c54703        7 days ago          200MB
postgres            10-alpine           1bd67b33a51c        2 weeks ago         70.3MB
postgres            10.12-alpine        1bd67b33a51c        2 weeks ago         70.3MB
postgres            10.11               5fea18bb7e20        2 months ago        250MB
postgres            10.11-alpine        fb495265b81f        2 months ago        72.4MB
postgres            10.10-alpine        a071f71cc9bb        5 months ago        71.8MB
postgres            10.10               9a05a2b9e69f        5 months ago        211MB
postgres            10.2-alpine         400ef39e0dbf        2 years ago         39.5MB
...
...
...
```

thaJeztah:
defaulting to `:latest` would make sense, and would be in line with what the `docker` CLI does (https://github.com/docker/cli/blob/6deb4f1f634003188e8319b8266ccdc588651fb1/cli/command/image/pull.go#L56-L63), which requires the `-a` / `--all-tags` option to be set to pull all tags (note that for `push` the `docker cli` currently has the same behavior as is reported here, but this is changed in the upcoming release: https://github.com/moby/moby/pull/40302

```console
Usage:	docker pull [OPTIONS] NAME[:TAG|@DIGEST]

Pull an image or a repository from a registry

Options:
  -a, --all-tags                Download all tagged images in the repository
      --disable-content-trust   Skip image verification (default true)
[...]

PR #2656, opened 2020-08-29
ContainerCollection.run sets tag to "latest" if not set.

dan-hipschman-od:
This fixes issue https://github.com/docker/docker-py/issues/2545.

The issue is when you call `client.containers.run("some-image")` and `some-image:latest` doesn't exist locally. `run` will call `ImageCollection.pull` without passing a tag:
https://github.com/docker/docker-py/blob/2b1c7eb72406bcdb4b6eb3567333a27e0eb6e1ed/docker/models/containers.py#L808-L814

Since the tag doesn't exist, `pull` will pull all tags from the repository:
https://github.com/docker/docker-py/blob/2b1c7eb72406bcdb4b6eb3567333a27e0eb6e1ed/docker/models/images.py#L398-L403

This behavior of `run` seems unintuitive since it will only run the `latest` image and the rest are not needed.

This patch only modifies `run`. I felt like it might be preferable for `pull` to mimic `docker pull`, and only pull all tags if the user explicitly asks for that with a keyword arg like `all=True`. However, the current behavior of `pull` is documented, so that would be a major breaking change.

I also considered whether `ContainerCollection.create` should append the `latest` tag if there was none, but its current behavior is also documented. Hence, only updating `run` seems like the safest approach.

### DIFF
--- a/docker/models/containers.py
+++ b/docker/models/containers.py
@@ -9,7 +9,7 @@ from ..errors import (
     NotFound, create_unexpected_kwargs_error
 )
 from ..types import HostConfig
-from ..utils import version_gte
+from ..utils import parse_repository_tag, version_gte
 from .images import Image
 from .resource import Collection, Model
 
@@ -541,7 +541,8 @@ class ContainerCollection(Collection):
             'Reticulating spline 1...\\nReticulating spline 2...\\n'
 
         Args:
-            image (str): The image to run.
+            image (str): The image to run. If the tag is not specified, it will
+                be set to "latest".
             command (str or list): The command to run in the container.
             auto_remove (bool): enable auto-removal of the container on daemon
                 side when the container's process exits.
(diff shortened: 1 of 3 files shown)


aiordache:
Thank you for this PR  but I think a better way to fix this is in https://github.com/docker/docker-py/pull/2671 as this issue is a side effect of missing a default for tag in the pull method.
If you don't mind I will close this in favor of what I proposed.

dan-hipschman-od:
> Thank you for this PR but I think a better way to fix this is in #2671 as this issue is a side effect of missing a default for tag in the pull method.
> If you don't mind I will close this in favor of what I proposed.

I prefer your approach of making the default of `pull()` to pull only the latest tag. The only reason I didn't do that is that the "pull all" behavior is documented, and so I thought it might break some existing code.

PR #2671, opened 2020-09-15
Set image default tag on pull

aiordache:
Set default tag to `latest` on pull.
Resolves https://github.com/docker/docker-py/issues/2545

### DIFF
--- a/docker/api/image.py
+++ b/docker/api/image.py
@@ -349,7 +349,8 @@ class ImageApiMixin(object):
 
         Args:
             repository (str): The repository to pull
-            tag (str): The tag to pull
+            tag (str): The tag to pull. If ``tag`` is ``None`` or empty, it
+                is set to ``latest``.
             stream (bool): Stream the output as a generator. Make sure to
                 consume the generator, otherwise pull might get cancelled.
             auth_config (dict): Override the credentials that are found in the
@@ -358,7 +359,8 @@ class ImageApiMixin(object):
             decode (bool): Decode the JSON data from the server into dicts.
                 Only applies with ``stream=True``
             platform (str): Platform in the format ``os[/arch[/variant]]``
-            all_tags (bool): Pull all image tags.
+            all_tags (bool): Pull all image tags, the ``tag`` parameter is
+                ignored.
 
         Returns:
             (generator or str): The output
--- a/docker/api/image.py
+++ b/docker/api/image.py
@@ -343,7 +343,7 @@ class ImageApiMixin(object):
         return self._result(self._post(url, params=params), True)
 
     def pull(self, repository, tag=None, stream=False, auth_config=None,
-             decode=False, platform=None):
+             decode=False, platform=None, all_tags=False):
         """
         Pulls an image. Similar to the ``docker pull`` command.
 
@@ -358,6 +358,7 @@ class ImageApiMixin(object):
             decode (bool): Decode the JSON data from the server into dicts.
                 Only applies with ``stream=True``
             platform (str): Platform in the format ``os[/arch[/variant]]``
+            all_tags (bool): Pull all image tags.
 
         Returns:
             (generator or str): The output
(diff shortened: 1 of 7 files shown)


thaJeztah:
Looks like a duplicate of https://github.com/docker/docker-py/pull/2656

thaJeztah:
With this change, will it still be possible to achieve the "pull all tags" behavior? (i.e. to implement "docker pull --all-tags ..." ?

aiordache:
> With this change, will it still be possible to achieve the "pull all tags" behavior? (i.e. to implement "docker pull --all-tags ..." ?

Oops, don't think so. Let me try something else.

docker/models/images.py
```diff
@@ -395,7 +395,7 @@ def load(self, data):
 
         return [self.get(i) for i in images]
 
-    def pull(self, repository, tag=None, **kwargs):
+    def pull(self, repository, tag='latest', **kwargs):
         """
         Pull an image of the given name and return it. Similar to the
         ``docker pull`` command.
```
dan-hipschman-od, line 401: The wording of the docstring should be updated to make it more clear. E.g.:
```
If no tag is specified, all tags from that repository will be pulled.
```
should be
```
If tag is None, all tags from that repository will be pulled.
```

docker/api/image.py
```diff
@@ -342,14 +342,14 @@ def prune_images(self, filters=None):
             params['filters'] = utils.convert_filters(filters)
         return self._result(self._post(url, params=params), True)
 
-    def pull(self, repository, tag=None, stream=False, auth_config=None,
+    def pull(self, repository, tag='latest', stream=False, auth_config=None,
```
thaJeztah, line 345: Thinking out loud: would it work to add an `all_tags` parameter, and based on that either keep `None` as `None`, or set `None -> latest` ?

Train of thought is that if you want to pull _all tags_, no `tag` is allowed to be set, so it feels a bit more natural to having a separate option (the original "no tag means all tags" is a bit awkward, and, well, bad choices in the past)

aiordache: @thaJeztah I added an `all_tags` parameter to the pull methods. Is the logic ok? If we set it, we ignore any tag set with `tag` or `repository` parameter.

docker/models/images.py
```diff
@@ -413,10 +413,11 @@ def pull(self, repository, tag=None, **kwargs):
                 config for this request.  ``auth_config`` should contain the
                 ``username`` and ``password`` keys to be valid.
             platform (str): Platform in the format ``os[/arch[/variant]]``
+            all_tags (bool): Pull all image tags
 
         Returns:
             (:py:class:`Image` or list): The image that has been pulled.
-                If no ``tag`` was specified, the method will return a list
+                If ``tag`` is None, the method will return a list
```
dan-hipschman-od, line 420: I think this comment needs updating. Should it be something like this?
> The image that has been pulled. If `all_tags` is True, then the method will return a list of :py:class:`Image` objects belonging to this repository.

docker/models/images.py
```diff
@@ -395,12 +395,12 @@ def load(self, data):
...
 
-    def pull(self, repository, tag=None, **kwargs):
+    def pull(self, repository, tag=None, all_tags=False, **kwargs):
         """
         Pull an image of the given name and return it. Similar to the
         ``docker pull`` command.
-        If no tag is specified, all tags from that repository will be
-        pulled.
+        If ``all_tags`` is set, the ``tag`` parameter is ignored and all image
+        tags will be pulled.
```
dan-hipschman-od, line 403: Even though it's intuitive, it's probably good to add explicit documentation that if `tag` is `None` or empty, then it will be set to `"latest"`. Also in the `api.image` module.

ulyssessouza (approved):
LGTM

ulyssessouza:
@thaJeztah PTAL

Issue #2659, opened 2020-09-01
Replace paramiko with the local ssh client


aiordache:
For the current implementation, we use paramiko to implement the ssh connection. Many users have had different issues and feature requests that paramiko doesn't address at the moment.

To mirror the docker cli behaviour, we should rely on the ssh client for connections to remote docker engines.

PR #2680, opened 2020-10-05
Shell out to local SSH client as alternative to a paramiko connection

aiordache:
We have several issues raised for failures of the paramiko SSH connection, support for SSH config option etc. Shelling out to the local ssh client may provide a better experience and it mirrors the behaviour of docker cli. We keep the paramiko connection as the default one.

To use the local ssh client instead of paramiko, we must set the parameter `use_ssh_client` when creating the docker client as below:

```python
def test_ssh_client():
    client = APIClient(base_url='ssh://remote', use_ssh_client=True)

    containers = client.containers()
    print("Running container IDs:")
    for c in containers:
        print("\t{}".format(c['Id']))
```
Make sure we have configured key-based ssh login to the target docker host:
```
Host remote
	Hostname 192.168.0.10
	User root
	IdentityFile ~/test/id_rsa
	Port 22
```

Output
```
Running container IDs:
        bd07c66651afc84f4223df96f59c4ffb3ba4ebb425feb6a6e0365e1a8987a024
        8625e8663fcaaa512d54444c08d99057ca893c806e05dd702cbd1dc86764065e
```

Closes https://github.com/docker/docker-py/issues/2659


Makefile
```diff
@@ -69,6 +73,29 @@ integration-dind-py3: build-py3 setup-network
 		--network dpy-tests docker-sdk-python3 py.test tests/integration/${file}
 	docker rm -vf dpy-dind-py3
 
+.PHONY: integration-ssh-py2
+integration-ssh-py2: build-dind-ssh build setup-network
```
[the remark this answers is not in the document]
aiordache: Yes, only in docker-compose.

docker/api/client.py
```diff
@@ -89,6 +89,9 @@ class APIClient(
         user_agent (str): Set a custom user agent for requests to the server.
         credstore_env (dict): Override environment variables when calling the
             credential store process.
+        use_ssh_client (bool): If set to `True`, an ssh connection is made
```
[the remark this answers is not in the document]
aiordache: Yep, that is the plan.

Dockerfile
```diff
@@ -13,3 +13,11 @@ RUN pip install -r test-requirements.txt
 
 COPY . /src
 RUN pip install .
+
+# install SSHD
+RUN apt-get install -y openssh-client
```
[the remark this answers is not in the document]
thaJeztah: better to move this to the top of this stage, so that it doesn't re-install with every code change.

Also needs a apt-get update, --no-install-recommends, and cleaning up after

Makefile
```diff
@@ -1,3 +1,6 @@
+TEST_API_VERSION ?= 1.39
+TEST_ENGINE_VERSION ?= 19.03.12
```
chris-crone, line 2: nit: 19.03.13 is out now

docker/transport/sshconn.py
```diff
@@ -20,33 +23,144 @@
...
+            port,
+            'docker system dial-stdio'
+        ]
+        self.proc = subprocess.Popen(
+            ' '.join(args),
+            shell=True,
+            stdout=subprocess.PIPE,
+            stdin=subprocess.PIPE)
+
+    def sendall(self, msg):
```
chris-crone, line 85: This looks identical except for the return to the `send` function so you can call that here and just not return the result.

docker/transport/sshconn.py
```diff
@@ -20,33 +23,144 @@
...
             'localhost', timeout=timeout, maxsize=maxsize
         )
-        self.ssh_transport = ssh_client.get_transport()
+        self.ssh_transport = None
+        if ssh_client:
+            self.ssh_transport = ssh_client.get_transport()
         self.timeout = timeout
+        self.host = host
+
+        # self.base_url = six.moves.urllib_parse.urlparse(host)
```
chris-crone, line 157: Can we remove this? Do we know why it was added originally?

tests/Dockerfile
```diff
@@ -1,15 +1,20 @@
...
 ARG APT_MIRROR
 RUN sed -ri "s/(httpredir|deb).debian.org/${APT_MIRROR:-deb.debian.org}/g" /etc/apt/sources.list \
     && sed -ri "s/(security).debian.org/${APT_MIRROR:-security.debian.org}/g" /etc/apt/sources.list
 
-RUN apt-get update && apt-get -y install \
+RUN apt-get update && apt-get -y install --no-install-recommends \
     gnupg2 \
     pass \
-    curl
+    curl \
```
chris-crone, line 11: I think that `curl` and `ssh` are already in the Python 3.7 Debian image:
```
docker run --rm python:3.7 sh -c "which curl && which ssh"
/usr/bin/curl
/usr/bin/ssh
```

If there's another reason you need them, I prefer having lists like this sorted so that it's easy to check if something's being installed.

chris-crone (commented):
Thanks @aiordache! I think we're nearly there

Dockerfile
```diff
@@ -1,7 +1,11 @@
 ARG PYTHON_VERSION=2.7
-
 FROM python:${PYTHON_VERSION}
 
+# Add SSH keys and set permissions
+COPY tests/ssh-keys /root/.ssh
+RUN chmod 600 /root/.ssh/id_rsa && \
+    chmod 600 /root/.ssh/id_rsa.pub
```
thaJeztah, line 7: I think with engine 20.x, we'll be able to use `COPY --chmod` https://github.com/moby/buildkit/pull/1492

would a `chmod -R` work for this?

```suggestion
RUN chmod -R 600 /root/.ssh/
```

tests/Dockerfile
```diff
@@ -6,10 +6,16 @@ ARG APT_MIRROR
...
     gnupg2 \
-    pass \
-    curl
+    openssh-client \
+    pass
+
+# Add SSH keys and set permissions
+COPY tests/ssh-keys /root/.ssh
+RUN chmod 600 /root/.ssh/id_rsa && \
+    chmod 600 /root/.ssh/id_rsa.pub
```
thaJeztah, line 18:
```suggestion
RUN chmod -R 600 /root/.ssh
```

docker/transport/sshconn.py
```diff
@@ -20,33 +23,140 @@
...
+    def sendall(self, msg):
+        if not self.proc or self.proc.stdin.closed:
+            raise Exception('SSH subprocess not initiated.'
+                            'connect() must be called first.')
+        self.proc.stdin.write(msg)
+        self.proc.stdin.flush()
+
+    def send(self, msg):
+        self.sendall(msg)
+        return len(msg)
```
chris-crone, line 94: nit: To avoid having to compute the message length separately, I would call `send` from `sendall` and use the byte count from the `write()` function's return.

```suggestion
        self.send(msg)

    def send(self, msg):
        if not self.proc or self.proc.stdin.closed:
            raise Exception('SSH subprocess not initiated.'
                            'connect() must be called first.')
        written = self.proc.stdin.write(msg)
        self.proc.stdin.flush()
        return written
```

docker/transport/sshconn.py
chris-crone: Re-ping on this as I see it's still here

tests/integration/api_build_test.py
```diff
@@ -339,7 +339,6 @@ def test_build_with_extra_hosts(self):
 
         assert self.client.inspect_image(img_name)
         ctnr = self.run_container(img_name, 'cat /hosts-file')
-        self.tmp_containers.append(ctnr)
```
chris-crone, line 342: I haven't dived into the code but why this change?

chris-crone (approved):
Once my comments are addressed, this LGTM! Thanks @aiordache!

docker/transport/sshconn.py
aiordache: Undid it by mistake.

docker/transport/sshconn.py
aiordache: Py2.7 request library seems to have a problem if we call `send` from `sendall`. Moved the logic intro a private method to be called in these two.

tests/integration/api_build_test.py
aiordache: The `self.run_container` already appends it to the list so there is no need to do it after.

PR #2699, opened 2020-11-12
Add max_pool_size parameter

Skazzino:
All the connection pools (UNIX socket/NPipe and SSH) are limited to a pool of 10 threads. This value is hardcoded in the Python code, so it can't be changed.

This pull request adds a `max_pool_size` parameter in the Client that makes able to tweak the number of thread for each connection pool used by the library. Of couse, by default this value is set to 10.

This feature was requested in #2477. We already opened a PR 1 year ago but we had some problems with the commits signatures.

### DIFF
--- a/docker/api/client.py
+++ b/docker/api/client.py
@@ -9,9 +9,9 @@ import websocket
 
 from .. import auth
 from ..constants import (DEFAULT_NUM_POOLS, DEFAULT_NUM_POOLS_SSH,
-                         DEFAULT_TIMEOUT_SECONDS, DEFAULT_USER_AGENT,
-                         IS_WINDOWS_PLATFORM, MINIMUM_DOCKER_API_VERSION,
-                         STREAM_HEADER_SIZE_BYTES)
+                         DEFAULT_MAX_POOL_SIZE, DEFAULT_TIMEOUT_SECONDS,
+                         DEFAULT_USER_AGENT, IS_WINDOWS_PLATFORM,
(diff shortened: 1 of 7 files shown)


tcaiazzi:
I've closed my old PR, hoping that this time it is ok!

aiordache (approved):
SGTM. Thank you!

docker/api/client.py
```diff
@@ -103,7 +103,8 @@ class APIClient(
     def __init__(self, base_url=None, version=None,
                  timeout=DEFAULT_TIMEOUT_SECONDS, tls=False,
                  user_agent=DEFAULT_USER_AGENT, num_pools=None,
-                 credstore_env=None, use_ssh_client=False):
+                 max_pool_size=DEFAULT_MAX_POOL_SIZE, credstore_env=None,
+                 use_ssh_client=False):
```
chris-crone, line 107: I would move the newest parameter to the end of the list so that if users are using this constructer with positional parameters, it will continue to work as expected.
```suggestion
                 credstore_env=None, use_ssh_client=False,
                 max_pool_size=DEFAULT_MAX_POOL_SIZE):
```

chris-crone (commented):
One minor issue then LGTM

docker/api/client.py
Skazzino: @chris-crone, I did it. Thanks for the suggestion! :smile:

PR #2705, opened 2020-11-19
Update Jenkinsfile with docker registry credentials

aiordache:
Add registry credentials to avoid being rate limited.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -25,10 +25,11 @@ def buildImages = { ->
       imageNamePy2 = "${imageNameBase}:py2-${gitCommit()}"
       imageNamePy3 = "${imageNameBase}:py3-${gitCommit()}"
       imageDindSSH = "${imageNameBase}:sshdind-${gitCommit()}"
-
-      buildImage(imageDindSSH, "-f tests/Dockerfile-ssh-dind .", "")
-      buildImage(imageNamePy2, "-f tests/Dockerfile --build-arg PYTHON_VERSION=2.7 .", "py2.7")
-      buildImage(imageNamePy3, "-f tests/Dockerfile --build-arg PYTHON_VERSION=3.7 .", "py3.7")
+      withDockerRegistry(credentialsId:'dockerbuildbot-index.docker.io') {
+        buildImage(imageDindSSH, "-f tests/Dockerfile-ssh-dind .", "")
+        buildImage(imageNamePy2, "-f tests/Dockerfile --build-arg PYTHON_VERSION=2.7 .", "py2.7")
+        buildImage(imageNamePy3, "-f tests/Dockerfile --build-arg PYTHON_VERSION=3.7 .", "py3.7")
+      }
     }
   }
 }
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -25,10 +25,11 @@ def buildImages = { ->
       imageNamePy2 = "${imageNameBase}:py2-${gitCommit()}"
       imageNamePy3 = "${imageNameBase}:py3-${gitCommit()}"
       imageDindSSH = "${imageNameBase}:sshdind-${gitCommit()}"
-
-      buildImage(imageDindSSH, "-f tests/Dockerfile-ssh-dind .", "")
-      buildImage(imageNamePy2, "-f tests/Dockerfile --build-arg PYTHON_VERSION=2.7 .", "py2.7")
-      buildImage(imageNamePy3, "-f tests/Dockerfile --build-arg PYTHON_VERSION=3.7 .", "py3.7")
+      withDockerRegistry(credentialsId:'dockerbuildbot-index.docker.io') {
```
thaJeztah, line 28: can we use an unprivileged account, and no access to any private images in the docker namespaces? ideally a PAT

thaJeztah (changes requested):

Jenkinsfile
StefanScherer: We just compared it to the docker/compose repo and used the same here. 😅 
And this pipeline also pushes the images. 🤔

StefanScherer: Do we have a better hub account that we could use?

thaJeztah: if needed, we could create one perhaps

thaJeztah: If it's currently pushing; what credentials does it use for that?

(sorry for brevity; on my phone 😂)

StefanScherer: I've updated the Jenkins secret to use a PAT instead of the real password.

StefanScherer:
According to https://ci-next.docker.com/public/blue/rest/organizations/jenkins/pipelines/docker-py/branches/master/runs/48/nodes/81/steps/135/log/?start=0 we probably have to change the tests and not the Jenkinsfile to fix the rate limiting.
It's running DIND

Jenkinsfile
```diff
@@ -82,41 +83,44 @@ def runTests = { Map settings ->
...
+            sh """docker run --rm -d --name ${dindContainerName} -v /tmp --privileged --network ${testNetwork} \\
+              ${imageDindSSH} dockerd -H tcp://0.0.0.0:2375
+            """
+            sh """docker run --rm \\
+              --name ${testContainerName} \\
+              -e "DOCKER_HOST=tcp://${dindContainerName}:2375" \\
+              -e 'DOCKER_TEST_API_VERSION=${apiVersion}' \\
+              --network ${testNetwork} \\
+              --volumes-from ${dindContainerName} \\
+              -v ~/.docker/config.json:/root/.docker/config.json \\
```
StefanScherer, line 98: Better watched in diff mode without white space changes.
This does the trick for the DIND environment to login.

StefanScherer (approved):
LGTM
we could change the Docker Hub account in another PR, this PR helps @aiordache to get a release out.

PR #2708, opened 2020-11-24
docker/api/image: replace use of deprecated "filter" argument

thaJeztah:
relates to the failures seen in https://github.com/moby/moby/pull/41709#issuecomment-732789163

The "filter" argument was deprecated in docker 1.13 (API version 1.25), and removed from API v1.41 and up. See https://github.com/docker/cli/blob/v20.10.0-rc1/docs/deprecated.md#filter-param-for-imagesjson-endpoint

This patch applies the name as "reference" filter, instead of "filter" for API 1.25 and up.

### DIFF
--- a/docker/api/image.py
+++ b/docker/api/image.py
@@ -81,10 +81,18 @@ class ImageApiMixin(object):
                 If the server returns an error.
         """
         params = {
-            'filter': name,
             'only_ids': 1 if quiet else 0,
             'all': 1 if all else 0,
         }
+        if name:
+            if utils.version_lt(self._version, '1.25'):
+                # only use "filter" on API 1.24 and under, as it is deprecated
+                params['filter'] = name
+            else:
+                if filters:
+                    filters['reference'] = name
+                else:
+                    filters = {'reference': name}
         if filters:
             params['filters'] = utils.convert_filters(filters)
         res = self._result(self._get(self._url("/images/json"), params=params),
--- a/tests/unit/api_image_test.py
+++ b/tests/unit/api_image_test.py
@@ -26,7 +26,18 @@ class ImageTest(BaseAPIClientTest):
         fake_request.assert_called_with(
             'GET',
             url_prefix + 'images/json',
-            params={'filter': None, 'only_ids': 0, 'all': 1},
+            params={'only_ids': 0, 'all': 1},
+            timeout=DEFAULT_TIMEOUT_SECONDS
+        )
+
+    def test_images_name(self):
+        self.client.images('foo:bar')
+
+        fake_request.assert_called_with(
+            'GET',
+            url_prefix + 'images/json',
+            params={'only_ids': 0, 'all': 0,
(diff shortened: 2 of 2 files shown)


docker/api/image.py
```diff
@@ -81,10 +81,14 @@ def images(self, name=None, quiet=False, all=False, filters=None):
                 If the server returns an error.
         """
         params = {
-            'filter': name,
             'only_ids': 1 if quiet else 0,
             'all': 1 if all else 0,
         }
+        if utils.version_lt(self._version, 1.25):
```
thaJeztah, line 87: Could use some eyes on this one; my concern with this was; if no explicit version is set, will this pick this branch, or the other one (below)?

We should "assume" API v1.25 or up if no API version is set, so if the logic needs to be reversed for that, let me know

aiordache: With 'auto' as default, we now retrieve the engine's api version so `self._version` should be set to that or to one that was explicitly set via the version parameter. In case there is no version retrieved from the engine or none was set, I think we raise an error.

thaJeztah: Thanks! So current logic in this PR should probably good to go then

thaJeztah: LOL, except that it doesn't work if I see the CI output;

```
tests/integration/api_client_test.py:71: in test_resource_warnings
    client.images()
docker/api/image.py:87: in images
    if utils.version_lt(self._version, 1.25):
docker/utils/utils.py:70: in version_lt
    return compare_version(v1, v2) > 0
docker/utils/utils.py:60: in compare_version
    s2 = StrictVersion(v2)
/usr/local/lib/python2.7/distutils/version.py:40: in __init__
    self.parse(vstring)
/usr/local/lib/python2.7/distutils/version.py:105: in parse
    match = self.version_re.match(vstring)
E   TypeError: expected string or buffer
```

thaJeztah: Ah, need a string; pushed 🤞

thaJeztah:
Ugh, more to fix looks like;

```
__________________________ ImageTest.test_image_ids ___________________________
2721tests\unit\api_image_test.py:50: in test_image_ids
2722    timeout=DEFAULT_TIMEOUT_SECONDS
2723c:\python37\Lib\unittest\mock.py:873: in assert_called_with
2724    raise AssertionError(_error_message()) from cause
2725E   AssertionError: Expected call: mock('GET', 'http+docker://localnpipe/v1.39/images/json', params={'filter': None, 'only_ids': 1, 'all': 0}, timeout=60)
2726E   Actual call: mock('GET', 'http+docker://localnpipe/v1.39/images/json', params={'only_ids': 1, 'all': 0, 'filters': '{"reference": ["None"]}'}, timeout=60)
```

thaJeztah:
Whoop! All green now; ptal 👍

aiordache (approved):
Thank you!

thaJeztah:
@ulyssessouza PTAL

ulyssessouza (approved):
LGTM

PR #2739, opened 2020-12-24
setup.py: Add support for Python 3.8 and 3.9

cclauss:
Signed-off-by: Christian Clauss <cclauss@me.com>

Improves the output at https://pypi.org/project/docker/ to match the GitHub Actions tests.
https://devguide.python.org/#status-of-python-branches
https://devguide.python.org/devcycle/#end-of-life-branches

PR #2757, opened 2021-02-09
Setup.py: Add support for current versions of Python

cclauss:
https://devguide.python.org/#status-of-python-branches
https://devguide.python.org/devcycle/#end-of-life-branches


cclauss:
Dup of #2739

PR #2761, opened 2021-02-10
Remove wrappedNode

StefanScherer:
Looks like the problem is within the `wrappedNode`. I don't think we really need it because the agents get wiped after some idle time.


aiordache (approved):
LGTM! Thank you :smile:

PR #2782, opened 2021-03-03
Verify TLS keys loaded from docker contexts

hakanardo:
This maches the behaviour of the docker cli when using contexts.


aiordache (approved):
Thank you!

PR #2795, opened 2021-04-02
Upgrade cryptography library to version 3.4.7

### LINKED ISSUE #2791: version requirements for cryptography should be consistent
Hi
It seems that version requirements for cryptography in setup.py and requirements.txt are not consistent
In setup.py, it is cryptography>=1.3.4
In requirements.txt, it is cryptography==3.2
Note that in pypi, the version of cryptography is always updating( now 3.4.6). Inconsistent version requirements will result in installing different version for cryptography if I use different  ways of installation.

feliperuhland:
Dependabot opened a pull request
93bcc0497d8302aa2d78bd7ef756fc2ff3fd0912 to upgrade cryptography from
2.3 to 3.2.

However, only `requirements.txt` was updated.
The extra requirements were kept outdated.

This commit was made to update the library to the last version.

Fix #2791

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -2,7 +2,7 @@ appdirs==1.4.3
 asn1crypto==0.22.0
 backports.ssl-match-hostname==3.5.0.1
 cffi==1.14.4
-cryptography==3.2
+cryptography==3.4.7
 enum34==1.1.6
 idna==2.5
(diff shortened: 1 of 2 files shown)


ulyssessouza (approved):
LGTM

Issue #2800, opened 2021-04-05
Check low-level documentation to be sure that is up to date


feliperuhland:
**Low-level API**
  - [ ] Configs
  - [ ] Containers
  - [ ] Images #2799
  - [ ] Building images
  - [ ] Networks
  - [ ] Volumes
  - [ ] Executing commands in containers
  - [ ] Swarms
  - [ ] Services
  - [ ] Plugins
  - [ ] Secrets
  - [ ] The Docker daemon
  - [ ] Configuration types

feliperuhland:
@ulyssessouza could you assign that issue to me?

PR #2801, opened 2021-04-05
Fix containers low-level documentation examples

feliperuhland:
I realize that low-level documentation has outdated examples, so I
created issue #2800 to fix that


aiordache (approved):
Thank you

PR #2812, opened 2021-04-08
Fix daemon low-level documentation examples

feliperuhland:
I realize that low-level documentation has outdated examples, so I
created issue #2800 to fix that

### DIFF
--- a/docker/api/daemon.py
+++ b/docker/api/daemon.py
@@ -42,7 +42,7 @@ class DaemonApiMixin(object):
 
         Example:
 
-            >>> for event in client.events(decode=True)
+            >>> for event in client.api.events(decode=True)
             ...   print(event)
             {u'from': u'image/with:tag',
              u'id': u'container-id',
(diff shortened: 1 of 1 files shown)


milas:
`events()` is exposed on `DockerClient`, so there's no need to access via the API object: https://github.com/docker/docker-py/blob/58aa62bb154a2ccea433cf475aefbd695fb5abc8/docker/client.py#L186-L188

```
>>> import docker
>>> events = client.events()
>>> for event in events:
...     print(event)
...
b'{"status":"create","id":"da6c7a2530f414815a8b46521a381e7860c3a0948672ed9186aad3a6674b4546","from":"alpine","Type":"container","Action":"create","Actor":{"ID":"da6c7a2530f414815a8b46521a381e7860c3a0948672ed9186aad3a6674b4546","Attributes":{"image":"alpine","name":"charming_jennings"}},"scope":"local","time":1660309782,"timeNano":1660309782826414637}\n'
```

Issue #2457, opened 2019-11-07
Dockerclient as a context manager


srdjan-catalyst:
Would be great to have Dockerclient as a context manager that runs close() on exit

Issue #2808, opened 2021-04-07
Make `DockerClient` a Context Manager


TBBle:
It's not obvious that `DockerClient` needs to have its `close()` method called to clean up resources, which can lead to unclosed sockets. We noticed this using asyncio and pytest together, which complains about unclosed sockets when tearing down test sessions.

Since `DockerClient` is wrapping `ApiClient`, and `ApiClient` inherits its `close()` method from `requests.Sessions`, it's not even obvious from a brief glance source what `close()` is for.

To make it easier to avoid mistakes, `DockerClient` could have `__enter__` and `__exit__` methods added, [identical to `requests.Sessions`](https://2.python-requests.org/en/master/_modules/requests/sessions/#Session), which would allow safer usage patterns such as

```Python
with docker.from_env() as docker_client:
 # Do stuff with docker_client
```

when that's feasible.

feliperuhland:
Hi, @TBBle I´m going to take a look at your request.

@ulyssessouza could you assign this issue to me?

johnthagen:
When this is implemented, it would be great if the docs could also be updated to show context manager usage as the first thing new users find: https://docker-py.readthedocs.io/en/stable/client.html#docker.client.from_env

I too was surprised to find a `close()` method and no context manager support to call this automatically.

johnthagen:
This also seems to be a duplicate of #2457.

TBBle:
Erk, how did I misss #2457? I 100% searched for Context Manager in the issues. Unless I searched for 'ContextManager'? >_<

TBBle:
In case anyone wants behaviour like this in the meantime, I recently came across [`contextlib.closing`](https://docs.python.org/3/library/contextlib.html#contextlib.closing) which could probably be used like
```
with contextlib.closing(docker.from_env()) as docker_client:
  # Do stuff with docker_client
```
but I haven't tested that myself.

Issue #2766, opened 2021-02-15
Leaking file descriptors


srdjan-catalyst:
There are two places where file descriptors are being leaked , at least with python 3.8:

1. `_get_raw_response_socket():` in case of Unix sockets, additional referencing seems to stop them from being closed
2. When `DockerClient` is not used as a context manager it does not call `close()` on the api session object.

I have managed to work around with:
```
class FixedAPIClient(docker.APIClient):
    def _get_raw_response_socket(self, response):
        sock = super()._get_raw_response_socket(response)
        if self.base_url.startswith('http+docker://'):
            sock._response = None
        return sock


class FixedDockerClient(docker.DockerClient):
    def __init__(self, *args, **kwargs):
        self.api = FixedAPIClient(*args, **kwargs)

    def __del__(self):
        try:
            if self.api:
                self.api.close()
        except AttributeError:
            pass
```

Our setup is quite complex. If you believe this should not happen and it is a corner case I will try to come up with a minimal example.

feliperuhland:
Hi @srdjan-catalyst 

I´m working on #2808, and your request is essential to be delivered together.

I will mention this issue in the commit message.

Thanks and have a nice day :)

Hoohaha:
I have same problem in my local. As a workaround you can close client to force release the sockets.

PR #2815, opened 2021-04-12
Add context manager to the client API

feliperuhland:
`ApiClient` inherence from `requests.session` that implements
`__enter__` and `__exit__` methods and the issue #2808 asks to add the
same behaviour to the `DockerClient`.

At the same time, issue #2766 reports a file descriptor leak when
`_get_raw_response_socket` is called.

To fix both requests, I added a `set` to the `ApiClient` instance to
track every socket open by `_get_raw_response_socket` and close when
the `close` method is called.

I added a test, and documentation now brings the context manager as an
example.

### DIFF
--- a/docker/api/client.py
+++ b/docker/api/client.py
@@ -76,6 +76,14 @@ class APIClient(
          u'Os': u'linux',
          u'Version': u'17.10.0-ce'}
 
+    It's possible to use `DockerClient` as a context manager and make sure every
+    socket will be closed:
+
+        >>> import docker
+        >>> docker.DockerClient(base_url='unix://var/run/docker/sock') as cli:
+        ...     cli.info()
+        ...
+
     Args:
         base_url (str): URL to the Docker server. For example,
             ``unix:///var/run/docker.sock`` or ``tcp://127.0.0.1:1234``.
@@ -118,6 +126,7 @@ class APIClient(
         self.timeout = timeout
         self.headers['User-Agent'] = user_agent
 
+        self._sockets_ref = set()
         self._general_configs = config.load_general_config()
 
         proxy_config = self._general_configs.get('proxies', {})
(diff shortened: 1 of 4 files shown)


docker/api/client.py
```diff
@@ -209,6 +218,20 @@ def __init__(self, base_url=None, version=None,
                 'library.'.format(MINIMUM_DOCKER_API_VERSION)
             )
 
+    def __exit__(self, exc_type, exc_value, traceback):
+        super().__exit__(exc_type, exc_value, traceback)
```
TBBle, line 222: We're losing any potential return value here, hence overriding any attempt by the super class to suppress an in-flight exception.

docker/api/client.py
```diff
@@ -209,6 +218,20 @@ def __init__(self, base_url=None, version=None,
...
+        self._close_sockets()
```
TBBle, line 223: I'm somewhat unclear on the flow here. If we don't define `__exit__`, won't the parent's `__exit__` call our close method anyway, so the `close` definition later would be sufficient.

I might be overlooking something though.

TBBle: I also feel like in the current flow, we'll super-call close twice, once in `super().__exit__`, and then again in our own `close` called from here. That seems odd, and I see a double-close in one of the tests too. Is there some _other_ issue with `close` that we're covering up by recalling `close()` so much?

tests/unit/api_test.py
```diff
@@ -544,17 +544,17 @@ def frame_header(stream, data):
...
-            return client._read_from_socket(
-                resp, stream=stream, tty=tty, demux=demux)
+                )
+        if tty:
+            url = client._url('/tty')
+        else:
+            url = client._url('/no-tty')
+        resp = client._post(url, stream=True)
+        return client._read_from_socket(
+            resp, stream=stream, tty=tty, demux=demux)
```
TBBle, line 557: Is this deliberately introducing a leak by lifting the APIClient out of a `with`?

Would it be better to make this a generator instead or something?

TBBle:
This should also close #2457.

PR #2828, opened 2021-04-23
Prevent pip cache in Docker image to save image size

PeterDaveHello:
This will help save some disk space from the images, both of the images are nearly 1GB 😅


afshin2020 (approved):

milas (approved):
Something went awry with CI here - DCO manually confirmed

PR #2838, opened 2021-05-13
expose session context manager interface from ApiClient in DockerClient

graingert:


docker/client.py
```diff
@@ -211,6 +212,14 @@ def close(self):
         return self.api.close()
     close.__doc__ = APIClient.close.__doc__
 
+    def __enter__(self):
+        return self.api.__enter__()
+    __enter__.__doc__ = APIClient.__enter__.__doc__
```
graingert, line 217:
```suggestion
    def __enter__(self):
        self.api.__enter__()
        return self

    __enter__.__doc__ = APIClient.__enter__.__doc__
```

docker/client.py
```diff
@@ -211,6 +212,14 @@ def close(self):
...
+
+    def __exit__(self, *args, **kwargs):
+        return self.api.__exit__(self, *args, **kwargs)
```
graingert, line 220:
```suggestion
        return self.api.__exit__(*args, **kwargs)
```

milos-korenciak:
Please, sign your commits https://github.com/apps/dco

Issue #2840, opened 2021-05-18
websocket-client dependency range isn't compatible with newly released 1.0.0


pelletier2017:
When doing a `pip install` with a fresh venv it installs websocket-client 1.0.0 due to the websocket-client>=0.32.0 requirement.

requirements.txt
```
docker==5.0.0
```
test.py
```
import docker
```

Running `python test.py` gives the following error
```
Traceback (most recent call last):
  File "test_failing_docker_sdk.py", line 1, in <module>
    import docker
  File "/projects/myproj/src/main/python/env/lib/python3.6/site-packages/docker/__init__.py", line 2, in <module>
    from .api import APIClient
  File "/projects/myproj/src/main/python/env/lib/python3.6/site-packages/docker/api/__init__.py", line 2, in <module>
    from .client import APIClient
  File "/projects/myproj/src/main/python/env/lib/python3.6/site-packages/docker/api/client.py", line 10, in <module>
    from .. import auth
  File "/projects/myproj/src/main/python/env/lib/python3.6/site-packages/docker/auth.py", line 5, in <module>
    import six
ModuleNotFoundError: No module named 'six'
```

logs when installing with pip
```
Collecting websocket-client>=0.32.0 (from docker==5.0.0->-r requirements.txt (line 2))
  Using cached https://files.pythonhosted.org/packages/ba/d1/501076b54481412df1bc4cdd1fe479f66e17857c63ec5981bedcdc2ca793/websocket_client-1.0.0-py2.py3-none-any.whl
```

```
$  pip freeze && python --version && docker version
certifi==2020.12.5
chardet==4.0.0
docker==5.0.0
idna==2.10
requests==2.25.1
urllib3==1.26.4
websocket-client==1.0.0
Python 3.6.8
Client: Docker Engine - Community
 Version:           19.03.8
 API version:       1.40
 Go version:        go1.12.17
 Git commit:        afacb8b
 Built:             Wed Mar 11 01:21:11 2020
 OS/Arch:           darwin/amd64
 Experimental:      false

Server: Docker Engine - Community
 Engine:
  Version:          19.03.8
  API version:      1.40 (minimum version 1.12)
  Go version:       go1.12.17
  Git commit:       afacb8b
  Built:            Wed Mar 11 01:29:16 2020
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          v1.2.13
  GitCommit:        7ad184331fa3e55e52b890ea95e65ba581ae3429
 runc:
[...]

PR #2841, opened 2021-05-18
limit upper bound of dependencies in setup.py and add six as explicit dependency

pelletier2017:
The dependency `websocket-client` recently released version 1.0.0 which removed `six` as a dependency. This breaks because `docker/auth.py` uses `six` without adding it as a dependency in `setup.py`. 

Version `1.16.0` of `six` is used because it was the only one I've tested it with but the lower bound could likely be lowered.

This addresses https://github.com/docker/docker-py/issues/2840

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -11,8 +11,9 @@ ROOT_DIR = os.path.dirname(__file__)
 SOURCE_DIR = os.path.join(ROOT_DIR)
 
 requirements = [
-    'websocket-client >= 0.32.0',
-    'requests >= 2.14.2, != 2.18.0',
+    'websocket-client >= 0.32.0, < 1',
+    'requests >= 2.14.2, != 2.18.0, < 3',
+    'six >= 1.16.0, < 2',
 ]
 
 extras_require = {


setup.py
```diff
@@ -11,7 +11,7 @@
 SOURCE_DIR = os.path.join(ROOT_DIR)
 
 requirements = [
-    'websocket-client >= 0.32.0',
+    'websocket-client >= 0.32.0, < 1.0.0',
     'requests >= 2.14.2, != 2.18.0',
```
till, line 15: Can someone take a moment and also fix this? :D Otherwise we are back here in x days or weeks?

```suggestion
    'requests >= 2.14.2, != 2.18.0, < 3',
```

pelletier2017:
Updated with input from @till about the other dependencies and added explicit dependency for six which is used in the project.

setup.py
```diff
@@ -11,8 +11,9 @@
 SOURCE_DIR = os.path.join(ROOT_DIR)
 
 requirements = [
-    'websocket-client >= 0.32.0',
-    'requests >= 2.14.2, != 2.18.0',
+    'websocket-client >= 0.32.0, < 1',
+    'requests >= 2.14.2, != 2.18.0, < 3',
+    'six >= 1.16.0, < 2',
```
JonZeolla, line 16: This should not be added here, this was resolved in https://github.com/docker/docker-py/commit/a9748a8b702a3c75b46ba8c8d0490e4b8ec5ab04

till: @JonZeolla what's the best way to get people's attention to PRs? 😅

milas:
We no longer use `six` / support Python 2

Issue #2842, opened 2021-05-19
undeclared dependency on six package


kapilt:
installing docker python in a fresh py3, and attempt to use, fails due to undeclared dependency on the python six package (auth.py)

cliffwoolley:
@ulyssessouza - This seems related to your recent change https://github.com/docker/docker-py/pull/2788 ?

cliffwoolley:
There seem to still be quite a few files in the 5.0.0 package that `import six`.
```
docker/transport/npipeconn.py
docker/transport/unixconn.py
docker/transport/npipesocket.py
docker/transport/sshconn.py
docker/api/plugin.py
docker/api/secret.py
docker/api/image.py
docker/api/exec_api.py
docker/api/container.py
docker/api/swarm.py
docker/api/config.py
docker/credentials/store.py
docker/models/images.py
docker/types/healthcheck.py
docker/types/services.py
docker/types/base.py
docker/types/containers.py
docker/auth.py
docker/utils/build.py
docker/utils/json_stream.py
docker/utils/socket.py
```

cliffwoolley:
Looks like this duplicates https://github.com/docker/docker-py/issues/2807 .

MartinWallgren:
six was removed from setup.py in [this commit](https://github.com/docker/docker-py/commit/c8fba210a222d4f7fde90da8f48db1e7faa637ec)

geerlingguy:
Yeah... a lot of my automation has been failing since a couple weeks ago when this new release was made. Finally found the source of it :) (ref: https://github.com/geerlingguy/drupal-container/issues/33)

Issue #2807, opened 2021-04-06
[5.0.0] Dropped dependency six but remains importing six in 32 Python files?


hartwork:
Hi!

While bumping docker-py for Gentoo I noticed that commit c8fba210a222d4f7fde90da8f48db1e7faa637ec dropped the dependency on six from `setup.py` and `requirements.txt`. When grepping the sources I find 32 more Python files that still import six so I would expect some trouble at runtime in an environment where no six is available (if we ignore dependencies pulling six in, indirectly). Is that known and intended?

Thanks and best, Sebastian

feliperuhland:
Hi @hartwork 

That makes total sense, and we will remove those dependencies.

@ulyssessouza could you assign this issue to me?

Thanks and have a good day :)

consideRatio:
Thank you @hartwork for reporting this issue, and everyone involved contributing to this project! :heart: :tada:

@feliperuhland are you still interested in working on this? I would suggest going for it if you don't find an open PR already about it.

Thank you again everyone contributing to this project, I'm one among many users who rely on it indirectly :heart:

feliperuhland:
I won't be able to finish that task.

Have a nice day.

hartwork:
@feliperuhland thanks for being transparent about it.

@aiordache this seems like a low hanging fruit, yet somewhat important. Are you aware of someone who can take over? Else: If I came up with a pull request myself, would you have a minute for review in the next few days?

feliperuhland:
Hi @hartwork

I'm going to dedicate my time to other projects.
I have a small WIP here https://github.com/feliperuhland/docker-py/pull/new/remove-six

I hope that helps.

hartwork:
Thanks.

Here's a direct link to @feliperuhland's commit if anyone else wants to have a look:
https://github.com/feliperuhland/docker-py/commit/71411f5596dd0b3ef523af3b0af4c81e2827cc04

asottile:
fixed in #2863

hartwork:
@asottile awesome, thank you! :pray:

hartwork:
@docker any chance for a new release?

thaJeztah:
Let me close this one, as it was fixed in https://github.com/docker/docker-py/pull/2863

I'll ask @ulyssessouza if it's possible to tag a new release

PR #2844, opened 2021-05-20
Add six to requirements.txt

NitzanDevOpser:
fixes [#2807](https://github.com/docker/docker-py/issues/2807)
fixes [#2842](https://github.com/docker/docker-py/issues/2842)

Made sure all imports contained in the package would have a corresponding requirement:
- grep'ed for 'import' or from, under docker/ dir 
- Filtered out local imports (starting with dot), and 'docker.' imports
- Print only imported module (awk)
- Sort, unique
$ grep -rh -e "^import\b" -e "^from\b"  docker/ | grep -v " \." | grep -v docker |  awk '{print $2}' |sort -u

Imported all listed output in a clean python:3.8-slim docker and caught ImportError exceptions - printing the missing packages
Made sure every* error'ed package is included in the requirements.txt, installed updated requirement.txt and ran import script again - No errors.

* Only exception is the windows-related modules which are imported only for windows.

** Unclear why 'six' was removed on March 2021 (commit: c8fba210)

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -13,5 +13,6 @@ pyOpenSSL==18.0.0
 pyparsing==2.2.0
 pywin32==227; sys_platform == 'win32'
 requests==2.20.0
+six==1.16.0
 urllib3==1.24.3
 websocket-client==0.56.0
--- a/setup.py
+++ b/setup.py
@@ -13,6 +13,7 @@ SOURCE_DIR = os.path.join(ROOT_DIR)
 requirements = [
     'websocket-client >= 0.32.0',
     'requests >= 2.14.2, != 2.18.0',
+    'six >= 1.4.0'
 ]
 
 extras_require = {


thaJeztah:
@NitzanDevOpser thanks! looks like I don't have edit permissions in this repository, but could you add:

fixes https://github.com/docker/docker-py/issues/2807
fixes https://github.com/docker/docker-py/issues/2842

to your PR description?


@ulyssessouza @ndeloof ptal

NitzanDevOpser:
Done.

Both on commit msg and this PR's 1st comment

thaJeztah:
Thanks!

MartinWallgren:
It needs to be re-added to setup.py too

NitzanDevOpser:
@MartinWallgren, Right.

Done

e0ne (approved):

JonZeolla (approved):

JonZeolla:
@aiordache or @ulyssessouza would one of you mind taking a look a this?  It seems like a rather simple bug introduced in 5.0.0 that likely warrants a patch release.

JonZeolla (changes requested):
I would have expected CI to fail, but it [succeeded](https://github.com/docker/docker-py/runs/2280319072#step:4:62) because CI installs `six` due to a bunch of transitive dependencies in `test-requirements.txt` and `requirements.txt`[1].

I suggest this PR gets some improved tests to ensure this kind of issue doesn't reoccur.  It's also requested by the [contributor docs](https://github.com/docker/docker-py/blob/master/CONTRIBUTING.md#7-add-tests).

One way to fix this is to pick up [this work in progress](https://github.com/docker/docker-py/commit/71411f5596dd0b3ef523af3b0af4c81e2827cc04).  I may give that a try if that's the preferred direction.

There does seem to be a preference for using `requirements.txt` and `test-requirements.txt` for testing (`make test` does this, as does CI and `tox.ini`), but I would think we would at least want to also test with `setup.py`.  Any thoughts on that?

Alternatively, since `setup.py` indicates `websocket-client >= 0.32.0` but `requirements.txt` requires `websocket-client==0.56.0` perhaps `websocket-client >= 0.32.0, < 1` in `setup.py` is another way to solve this problem.  Not my favourite approach, but keeps `install_requires` pretty minimal.  This would also require testing the `setup.py` installation approach.

Open to feedback on the preferred approach, especially from the maintainers.  Sorry for all the questions; I'd be happy to contribute once there's some more clear preferences/direction.

1: `pytest-cov==2.6.1` and `pytest-timeout==1.3.3` in `test-requirements.txt`, as well as `websocket-client==0.56.0`, `paramiko==2.4.2`, `packaging==16.8`, and `pyOpenSSL==18.0.0` in `requirements.txt`

JonZeolla:
Duplicate of https://github.com/docker/docker-py/pull/2863

thaJeztah:
> Duplicate of #2863

I see that one was merged; is this one closable?

JonZeolla:
@thaJeztah yes I believe this should be closed.

thaJeztah:
Thanks! Let me close this one. Thanks for contributing, @NitzanDevOpser !!

Issue #836, opened 2015-11-03
[Master ticket] Better documentation


shin-:
#454 Making use of Sphinx autodoc for API docs
#499 Document possible exceptions
#821 Clean up documentation

Main takeaways:

- API docs are poorly formatted and difficult to navigate.
- Some users find it valuable to have the documentation closer to the code itself.
- More information could be given about the behavior of our methods (exceptions raised, etc.)
- User guides in addition to functional documentation could prove useful.

dnephin:
If I could add one more point, I think it would be awesome to have a link from the docker-py docs to the official API for the endpoint as well.

shin-:
I didn't think of that, but you're right. Noting this down!

bfirsh:
I've added some check boxes to see how we're doing here! I think the only things left are:

- Really comprehensive exception docs. This is much improved in #1186, which at least documents what _internal_ exceptions are raised, but I think documenting external exceptions is far too large of a job. We could handle on a case-by-case basis based on what bugs we get, etc.
- `ulimits` docs. Don't think this is done, yet.
- Some users guides are in #1186, the rest will be in https://github.com/docker/docker.github.io/pull/210

PR #2851, opened 2021-06-03
Update the Image.save documentation with a working example. Issue #836

huogerac:
The former save example doesn't work, so I updated it for one that works.

### DIFF
--- a/docker/models/images.py
+++ b/docker/models/images.py
@@ -84,9 +84,9 @@ class Image(Model):
 
         Example:
 
-            >>> image = cli.get_image("busybox:latest")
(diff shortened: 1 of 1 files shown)


huogerac:
Hi,
Is something wrong in the PR? Am I missing something, it's about a quick fix in the documentation, no code was changed.
It's to help others use the library using the documentation. Indeed, I have more example to add to the docker-py.readthedocs.io, however, I'm waiting for this PR to be accepted or not..
Cheers
Roger

mat007 (approved):
Thanks!

PR #2861, opened 2021-06-29
Create Python3 only wheel.

adiroiban:
bdist_wheel universal should only be use if the project does not have any C extensions and supports Python 2 and 3.
The current project only supports py3.6 and above.


milas:
This was fixed in #2898 when Python 3.10 support was added. Thanks for the PR!

PR #2862, opened 2021-06-29
Return 12 character short_ids

### LINKED ISSUE #1491: container id only 10
# code:
print client.containers.list(all=True)
print docker.version

# output:
[<Container: 4ba3e59b9f>, <Container: fc5d8c6f28>, <Container: 72b6c75141>, <Container: 42685100cd>, <Container: 1d383bbd99>]
2.1.0

This is container id only 10 ? 
Here is my docker host container ID
# docker ps -a -q
4ba3e59b9fed
fc5d8c6f2845
72b6c751412a
42685100cdbc
1d383bbd9958

### LINKED ISSUE #2660: Change container.short_id to 12 characters
The 12 character short id is used in quite a few places.

For example, the default alias for all containers in a network is their respective 12 character id.
The default representation in the cli is 12 character id.
The usual way to reference containers is by their 12 character id.

The current 10 digit short_id is quite meaningless and creates more confusion than help. Non of the usual uses for the short id will work with it.

Why not change it to 12?

benfasoli:
Hi all 👋 Thanks for the work here - I've found it really helpful for several projects.

I'd like to discuss changing the behavior of `Model.short_id` from returning a 10 character string to a 12 character string which aligns more closely with the docker cli. It looks like this has been requested previously by #1491 and #2660. 

For example (from the [docs](https://docs.docker.com/engine/reference/commandline/ps/#examples)), running `docker ps` without explicitly setting `--no-trunc`:

```bash
$ docker ps

CONTAINER ID        IMAGE                        COMMAND                CREATED              STATUS              PORTS               NAMES
4c01db0b339c        ubuntu:12.04                 bash                   17 seconds ago       Up 16 seconds       3300-3310/tcp       webapp
d7886598dbe2        crosbymichael/redis:latest   /redis-server --dir    33 minutes ago       Up 33 minutes       6379/tcp            redis,webapp/db
```

It looks like 12 character ids were once the intended behavior - the object reprs in the [docker-py documentation](https://github.com/docker/docker-py#usage) from #1186 show 12 digit ids:

```python
>>> client.containers.list()
[<Container '45e6d2de7c54'>, <Container 'db18e4f20eaa'>, ...]
```

The specific issue I'm having is that the 12 character container ID is used by docker as a container's default hostname. Here's a minimal example of a container trying to connect to itself by hostname:

```python
import docker


client = docker.from_env()


nginx_container = client.containers.run("nginx:latest", detach=True, auto_remove=True)


# Networking with 64 character id as hostname will fail.
exit_code, response = nginx_container.exec_run(f"curl {nginx_container.id}")
assert exit_code == 6
assert b"Could not resolve host:" in response

# Networking with 10 character id as hostname will fail.
exit_code, response = nginx_container.exec_run(f"curl {nginx_container.short_id}")
assert exit_code == 6
assert b"Could not resolve host:" in response

# Networking with 12 character id as hostname will succeed 🚀
exit_code, response = nginx_container.exec_run(f"curl {nginx_container.id[:12]}")
assert exit_code == 0, response
assert b"Welcome to nginx!" in response


nginx_container.stop()
```

API requests treat the id as a variable width prefix and returns valid responses as long as the prefix doesn't match multiple resources. So I think using 12 characters shouldn't break backwards compatibility for code that currently uses the 10 character `short_id`.

Here's a first pass at the implementation.

1. Return a 12 character `short_id` for classes that inherit from `docker.models.resource.Model` and a 19 character `short_id` for `Image` ids prefixed by `sha256:`.
2. Expand the `fake_api` ids to 64 characters.
3. Modify `ImageTest.test_short_id` to expect a 12 character response.
4. Add `ContainerTest.test_short_id` to expect a 12 character response.

This currently fail flake8's E501 (line > 79 characters) requirement in a few places but I left the formatting as is for now since there are existing docstrings which fail E501 and it made the changes more readable. Happy to tweak to your preference.

Closes #1491
Closes #2660


CorentinNeovision:
Up, that doesn't seems like a big deal but it would improve the python API. I was thinking about creating a database storing container stats but I couldn't choose a coherent length for the short ID because of this python/docker cli discrepancy.
It is always possible to use the full length ID, but it is almost never used because it's non user-friendly. As said before even the internal container reprs use the 12 char short id.

milas (approved):
Thanks for the improvement and tests! Will get this included in the upcoming 6.0.0 release 🙂

PR #2863, opened 2021-07-05
use python3.6+ constructs

### LINKED ISSUE #2842: undeclared dependency on six package

asottile:
closes #2842

I originally found that installing `pip install docker` was broken because `six` is missing as a dependency -- but since `docker-py` is python3.6+ there's really no reason to continue to use `six`

so what I did is I ran [pyupgrade](https://github.com/asottile/pyupgrade) against docker-py and then I fixed the few things that `flake8` pointed out afterwards

this should "officially" drop the python-2 compatibility and use python3.6+-only constructs


ulyssessouza (approved):
LGTM

MrMino:
Hey @ulyssessouza. Do you know when's the next release that will contain this fix?

jamesmishra:
It's great to see this Pull Request merged into master, but it would be even nicer to release it as version 5.0.1 and then yank 5.0.0 from PyPI.

Issue #2867, opened 2021-07-24
ModuleNotFoundError: No module named 'six'


levsh:
Hi!
Is this a bug or am I missing something?

```bash
levsh@book:/tmp/test $ python -m venv tenv
levsh@book:/tmp/test $ source tenv/bin/activate
(tenv) levsh@book:/tmp/test $ pip install -U pip
Collecting pip
  Using cached pip-21.1.3-py3-none-any.whl (1.5 MB)
Installing collected packages: pip
  Attempting uninstall: pip
    Found existing installation: pip 20.2.1
    Uninstalling pip-20.2.1:
      Successfully uninstalled pip-20.2.1
Successfully installed pip-21.1.3
(tenv) levsh@book:/tmp/test $ pip install docker
Collecting docker
  Using cached docker-5.0.0-py2.py3-none-any.whl (146 kB)
Collecting websocket-client>=0.32.0
  Using cached websocket_client-1.1.0-py2.py3-none-any.whl (68 kB)
Collecting requests!=2.18.0,>=2.14.2
  Using cached requests-2.26.0-py2.py3-none-any.whl (62 kB)
Collecting urllib3<1.27,>=1.21.1
  Using cached urllib3-1.26.6-py2.py3-none-any.whl (138 kB)
Collecting certifi>=2017.4.17
  Using cached certifi-2021.5.30-py2.py3-none-any.whl (145 kB)
Collecting idna<4,>=2.5
  Using cached idna-3.2-py3-none-any.whl (59 kB)
Collecting charset-normalizer~=2.0.0
  Using cached charset_normalizer-2.0.3-py3-none-any.whl (35 kB)
Installing collected packages: urllib3, idna, charset-normalizer, certifi, websocket-client, requests, docker
Successfully installed certifi-2021.5.30 charset-normalizer-2.0.3 docker-5.0.0 idna-3.2 requests-2.26.0 urllib3-1.26.6 websocket-client-1.1.0
(tenv) levsh@book:/tmp/test $ python
Python 3.8.6 (default, Oct  6 2020, 15:18:31) 
[Clang 11.0.0 (clang-1100.0.33.16)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import docker
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/private/tmp/test/tenv/lib/python3.8/site-packages/docker/__init__.py", line 2, in <module>
    from .api import APIClient
[...]

PR #2870, opened 2021-07-26
Add to README about temporary fix for six dependency

colinbut:
As a temporary measure to fix the issue identified in #2867 we can install the `six` dependency whilst we install this `docker-py`.

This PR updates the `README.md` to signal this potential fix.


JonZeolla (commented):
This should not be needed as of https://github.com/docker/docker-py/commit/a9748a8b702a3c75b46ba8c8d0490e4b8ec5ab04

Issue #2885, opened 2021-08-31
Latest release 5.0.1 container exec_run fails


joelcomp1:
Running urllib3-1.26.6, requests-2.26.0 and latest version of docker 5.0.1 I now can't use the exec_run() cmd any longer to run a script on my container I keep getting:

  container.exec_run(f"/opt/jboss/keycloak/bin/add-user-keycloak.sh -u {keycloak_user} -p {keycloak_password}")
  File "/usr/local/lib/python3.9/site-packages/docker/models/containers.py", line 198, in exec_run
    exec_output = self.client.api.exec_start(
  File "/usr/local/lib/python3.9/site-packages/docker/utils/decorators.py", line 19, in wrapped
    return f(self, resource_id, *args, **kwargs)
  File "/usr/local/lib/python3.9/site-packages/docker/api/exec_api.py", line 157, in exec_start
    res = self._post_json(
  File "/usr/local/lib/python3.9/site-packages/docker/api/client.py", line 296, in _post_json
    return self._post(url, data=json.dumps(data2), **kwargs)
  File "/usr/local/lib/python3.9/site-packages/docker/utils/decorators.py", line 46, in inner
    return f(self, *args, **kwargs)
  File "/usr/local/lib/python3.9/site-packages/docker/api/client.py", line 233, in _post
    return self.post(url, **self._set_request_timeout(kwargs))
  File "/usr/local/lib/python3.9/site-packages/requests/sessions.py", line 590, in post
    return self.request('POST', url, data=data, json=json, **kwargs)
  File "/usr/local/lib/python3.9/site-packages/requests/sessions.py", line 542, in request
    resp = self.send(prep, **send_kwargs)
  File "/usr/local/lib/python3.9/site-packages/requests/sessions.py", line 655, in send
    r = adapter.send(request, **kwargs)
  File "/usr/local/lib/python3.9/site-packages/requests/adapters.py", line 439, in send
    resp = conn.urlopen(
  File "/usr/local/lib/python3.9/site-packages/urllib3/connectionpool.py", line 699, in urlopen
    httplib_response = self._make_request(
  File "/usr/local/lib/python3.9/site-packages/urllib3/connectionpool.py", line 445, in _make_request
    six.raise_from(e, None)
  File "<string>", line 3, in raise_from
  File "/usr/local/lib/python3.9/site-packages/urllib3/connectionpool.py", line 440, in _make_request
    httplib_response = conn.getresponse()
[...]

Issue #1799, opened 2017-11-08
AttributeError: 'SocketIO' object has no attribute 'raw'


graingert:
As raised in https://github.com/docker/compose/issues/5353 but I think it's actually a docker-py issue:
docker-composeup redis
Creating procensus_redis_1 ... 
Creating procensus_redis_1

ERROR: for procensus_redis_1  'SocketIO' object has no attribute 'raw'

ERROR: for redis  'SocketIO' object has no attribute 'raw'
```pytb
Traceback (most recent call last):
  File "/home/graingert/.virtualenvs/gitlab/bin/docker-compose", line 11, in <module>
    sys.exit(main())
  File "/home/graingert/.virtualenvs/gitlab/lib/python3.6/site-packages/compose/cli/main.py", line 68, in main
    command()
  File "/home/graingert/.virtualenvs/gitlab/lib/python3.6/site-packages/compose/cli/main.py", line 121, in perform_command
    handler(command, command_options)
  File "/home/graingert/.virtualenvs/gitlab/lib/python3.6/site-packages/compose/cli/main.py", line 952, in up
    start=not no_start
  File "/home/graingert/.virtualenvs/gitlab/lib/python3.6/site-packages/compose/project.py", line 455, in up
    get_deps,
  File "/home/graingert/.virtualenvs/gitlab/lib/python3.6/site-packages/compose/parallel.py", line 70, in parallel_execute
    raise error_to_reraise
  File "/home/graingert/.virtualenvs/gitlab/lib/python3.6/site-packages/compose/parallel.py", line 168, in producer
    result = func(obj)
  File "/home/graingert/.virtualenvs/gitlab/lib/python3.6/site-packages/compose/project.py", line 441, in do
    start=start
  File "/home/graingert/.virtualenvs/gitlab/lib/python3.6/site-packages/compose/service.py", line 471, in execute_convergence_plan
    scale, detached, start
  File "/home/graingert/.virtualenvs/gitlab/lib/python3.6/site-packages/compose/service.py", line 396, in _execute_convergence_create
    "Creating",
  File "/home/graingert/.virtualenvs/gitlab/lib/python3.6/site-packages/compose/parallel.py", line 70, in parallel_execute
    raise error_to_reraise
  File "/home/graingert/.virtualenvs/gitlab/lib/python3.6/site-packages/compose/parallel.py", line 168, in producer
    result = func(obj)
[...]

PR #2886, opened 2021-08-31
fix(transport): remove disable_buffering option

Aposhian:
#2885 documents a breakage in 5.0.1 because `disable_buffering` is being passed to a `HTTPResponse` object (caused by commit https://github.com/docker/docker-py/commit/5fcc293ba268a89ea1535114d36fbdcb73ec3d88).

As mentioned in #1799, there is no way to disable buffering in Python3 on `HTTPResponse` for now. This removes the option, since the Python2 dependency was removed anyway.


ulyssessouza (approved):
LGTM

PR #2903, opened 2021-10-18
fix pywin32 dependency

keerthiadu:
Issue: https://github.com/docker/docker-py/issues/2902

PR #2912, opened 2021-11-11
Bump pywin32 to version 301 in setup requirements

avnes:
CVE-2021-32559
moderate severity
Vulnerable versions: < 301
Patched version: 301
An integer overflow exists in pywin32 prior to version b301 when adding an access control entry (ACE) to an access control list (ACL) that would cause the size to be greater than 65535 bytes. An attacker who successfully exploited this vulnerability could crash the vulnerable process.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -16,7 +16,7 @@ requirements = [
 
 extras_require = {
     # win32 APIs if on Windows (required for npipe support)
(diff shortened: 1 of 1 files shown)


avnes:
I am closing my pull request, because I just discovered that is is a duplicate of pull request https://github.com/docker/docker-py/pull/2903

PR #2920, opened 2021-12-01
Use `preexec_func` always

q0w:
`preexec_func` is still None if it is win32

### DIFF
--- a/docker/transport/sshconn.py
+++ b/docker/transport/sshconn.py
@@ -63,7 +63,7 @@ class SSHSocket(socket.socket):
             shell=True,
             stdout=subprocess.PIPE,
             stdin=subprocess.PIPE,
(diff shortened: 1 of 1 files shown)


milas (commented):
Can you explain the issue you're trying to solve here?

`preexec_fn` is set for non-Windows platforms *only*: https://github.com/docker/docker-py/blob/b2a18d7209f827d83cc33acb80aa31bf404ffd4b/docker/transport/sshconn.py#L48-L52

q0w:
so if not windows it would still be none https://github.com/docker/docker-py/blob/b2a18d7209f827d83cc33acb80aa31bf404ffd4b/docker/transport/sshconn.py#L48

milas (approved):

PR #2932, opened 2022-01-06
Fix for CWE-295: Improper Certificate Validation

avnes:
In the Secure Shell (SSH) protocol, host keys are used to verify the identity of remote hosts. Accepting unknown host keys may leave the connection open to man-in-the-middle attacks.

Do not accept unknown host keys. In particular, do not set the default missing host key policy for the Paramiko library to either AutoAddPolicy or WarningPolicy. Both of these policies continue even when the host key is unknown. The default setting of RejectPolicy is secure because it throws an exception when it encounters an unknown host key.

Reference: https://cwe.mitre.org/data/definitions/295.html


milas (approved):
Thanks for the PR! Tested locally against SSH integration tests including a new test specifically for this - see #3012.

PR #2933, opened 2022-01-06
Fix for adding static code analyzes to PR and merge process

avnes:
The docker-py code is actually in very state from a static code analyzes point of view, but I did find four issues with it. Three of them were false-positives as it flagged up unit tests using dummy usernames. The fourth one was more severe though since it found a security vulnerability in the usage of paramiko. I have raised another PR for the paramiko issue.

Signed-off-by: Audun Nes <audun.nes@gmail.com>

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -1,0 +1,39 @@
+name: "CodeQL"
+
+on:
+  push:
+    branches: [ master ]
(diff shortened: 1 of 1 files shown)


milas:
Thanks for raising the PR around Paramiko! I appreciate the work you did adding CodeQL here, but we are not looking to use this as part of our CI process at this time, so I'm going to close this PR.

PR #2944, opened 2022-01-27
Avoid deprecated splitnport function.

DanCardin:
I saw three ways forward to avoid the use of the deprecated `splitnport` function.

* Just use the parsed results' hostname. I couldn't track down the actual reason why the stripping of brackets is **actually** problematic. So I figured that would be harder to get merged.
* Inline the use of `splitnport`. I see https://github.com/docker/docker-py/pull/2567, but also that it's been unchanged since 2020.
* Do something else. the behavior you're getting from `splitnport` is essentially just chopping off the port part to find the hostname. But the rest of the logic seems unnecessary because you've already run `parse` and have `parsed_url.port` handy. So I opted to just directly perform the behavior of chopping off the port instead.

I have no opinion on which of the three options are chosen, I'm just aiming to stop this warning on my own code. But figured that if copying the whole of `splitnport` was the reason for no action, that this might be preferable; or else might cause https://github.com/docker/docker-py/pull/2567 to be merged

### DIFF
--- a/docker/utils/utils.py
+++ b/docker/utils/utils.py
@@ -14,7 +14,7 @@ from ..constants import DEFAULT_UNIX_SOCKET
 from ..constants import DEFAULT_NPIPE
 from ..constants import BYTE_UNITS
 
-from urllib.parse import splitnport, urlparse
+from urllib.parse import urlparse
 
 
 def create_ipam_pool(*args, **kwargs):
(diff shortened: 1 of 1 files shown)


DanCardin:
and fwiw, the current failure seems to be the same on that's on master

milas:
Hi! Sorry for not reviewing this PR sooner, and thanks for your contribution!

I merged similar changes that should fix this in #3006. For context, the CI configuration and more has changed in this repo since your PR was opened, so given the already lengthy delay, we wanted to be respectful of your time and not require you to rebase and get things fixed up.

Issue #996, opened 2016-03-17
Unit tests are broken?


adityamarella:
I get whole bunch of these errors when I run `make test`. I am using docker 1.10.3 

> docker run --volumes-from dpy-dind-ssl --env="DOCKER_HOST=tcp://docker:2375" --env="DOCKER_TLS_VERIFY=1" --env="DOCKER_CERT_PATH=/certs" --link=dpy-dind-ssl:docker docker-py py.test tests/integration
> ============================= test session starts ==============================
> platform linux2 -- Python 2.7.11 -- py-1.4.31 -- pytest-2.7.2
> rootdir: /home/docker-py, inifile: pytest.ini
> plugins: cov
> collected 111 items
> 
> tests/integration/api_test.py EEEEEEEEEE
> tests/integration/build_test.py EEEEE
> tests/integration/container_test.py EEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEEs
> tests/integration/exec_test.py EEEEEEE
> tests/integration/image_test.py EEEEEEEEEs
> tests/integration/network_test.py EEssssEEssEEE
> tests/integration/regression_test.py EEEEEE
> tests/integration/volume_test.py EEEEEEE
> 
> ==================================== ERRORS ====================================
> _________________ ERROR at setup of InformationTest.test_info __________________
> tests/integration/conftest.py:19: in setup_test_session
>     c.inspect_image(BUSYBOX)
> docker/utils/decorators.py:21: in wrapped
>     return f(self, resource_id, *args, **kwargs)
> docker/api/image.py:143: in inspect_image
>     self._get(self._url("/images/{0}/json", image)), True
> docker/utils/decorators.py:47: in inner
>     return f(self, *args, **kwargs)
> docker/client.py:112: in _get
>     return self.get(url, **self._set_request_timeout(kwargs))
> /usr/local/lib/python2.7/site-packages/requests/sessions.py:473: in get
>     return self.request('GET', url, **kwargs)
> /usr/local/lib/python2.7/site-packages/requests/sessions.py:461: in request
>     resp = self.send(prep, **send_kwargs)
> /usr/local/lib/python2.7/site-packages/requests/sessions.py:573: in send
>     r = adapter.send(request, **kwargs)
> /usr/local/lib/python2.7/site-packages/requests/adapters.py:415: in send
>     raise ConnectionError(err, request=request)
[...]

PR #2947, opened 2022-02-02
Fixes and improvements

kinday:
This PR fixes various issues and introduces CI improvements. There are no functional changes, only stylistic and/or infrastructure fixes. Issues were encountered during work on other contribution when I attempted to run `make test`.

## Fix flake8 errors (e1a203fc3ea43f2ff621f632935d5c1b80977b88)

`make test` was failing due to several formatting issues, mostly line width. I have reformatted code in multiple places to make `flake8` happy[^1].

## Fix integration tests race condition (89bce107230ede17efe2653aa033a6311c21ad55)

Integration tests were crashing as they were starting before Docker-in-Docker container was ready to accept connections. This was reported in #996 and `sleep` is not a viable fix as DinD startup time varies.

I have updated `Makefile`[^2] to run `netcat` in `busybox` image. This solution is used instead of running `netcat` on host system for two reasons: the binary might not be present and container could be unreachable from host network[^3]

## Run integration tests on CI (d28f522d0d2c40b9ce5f1d6bc0f09f9e22716efc)

Integration test suite does not fully pass on macOS thus I have updated workflow to run them on CI. This could help people with compatibility issues and ensure that proposed changes are not breaking anything.

**Update:** I’ve noticed integration tests are being run on Jenkins. Maybe this one could be removed, but running them via GHA allows to get them executed before PR was created, i.e. in contributor’s repository and give early feedback.

## Use existing DIND version (924e09a098ebb0c7a8a53503e854a8b4c89d4f22)

The `20.10.05` tag is no longer available on Docker Hub. I have replaced it with `20.10` as this seems to be more resilient than specifying full tag.

[^1]: I am not sure if I formatted the code according to Python code style; please tell if it should it be indented/splitted differently.
[^2]: I took liberty to partially reformat `Makefile` and replace shorthand flags with longhand ones for readability. I can either revert the changes, or apply the same style to whole file  or keep things as is. Your call.
[^3]: Could be observed on macOS as Docker is being run in virtual machine and the only way to access container is to bind ports to host network.


docker/api/image.py
```diff
@@ -377,7 +377,11 @@ def pull(self, repository, tag=None, stream=False, auth_config=None,
 
         Example:
 
-            >>> for line in client.api.pull('busybox', stream=True, decode=True):
+            >>> for line in client.api.pull(
+                'busybox',
+                stream=True,
+                decode=True
+            ):
```
kinday, line 384: This should probably look like this according to other places:
```suggestion
                decode=True):
```

docker/api/container.py
```diff
@@ -256,7 +256,10 @@ def create_container(self, image, command=None, hostname=None, user=None,
 
         .. code-block:: python
 
-            client.api.create_host_config(port_bindings={1111: ('127.0.0.1', 4567)})
+            client.api.create_host_config(port_bindings={1111: (
+                '127.0.0.1',
+                4567
+            )})
```
kinday, line 262: This should probably look like this according to other places:
```suggestion
            client.api.create_host_config(
                port_bindings={1111: ('127.0.0.1', 4567)})
```

milas:
Thank you for the fixes and improvements! I just brought this branch up to date. The integration test wait fix is much appreciated ✌️

Issue #2951, opened 2022-02-15
container.remove() does not support timeout


thigm85:
It is possible to increase `container.stop()` timeout by passing a `timeout` parameter. The same seems not to be possible for `container.remove()`. 

Is there a way to increase `container.remove()` timeout?

milas:
This is not supported by the engine API, see https://docs.docker.com/engine/api/v1.41/#tag/Container/operation/ContainerDelete

You can pass `force=True` to _immediately_ kill the container before removing it.

If you'd like to gracefully stop, call `stop()` followed by `remove()`.

PR #2959, opened 2022-03-02
Added timeout to container.remove()

cibinmathew:
Added timeout to `remove()` call (https://github.com/docker/docker-py/issues/2951)


milas:
See comment in #2951 - there is no timeout (`t`) parameter for this API.

PR #2962, opened 2022-03-04
feat: add py.typed file

QuentinN42:
According to PEP 561, we just need to add this file to mark the lib as typed.

I don't know if all the lib is typed, maybe we also need to do that before merging.


milas:
There are no type annotations currently, so exposing it as typed does not make sense AFAICT.

Issue #2902, opened 2021-10-18
installing latest 5.0.3 on windows machines is still using pywin32==227 but not pywin32==301


keerthiadu:
[Bump pywin32 from 227 to 301 ]( https://github.com/docker/docker-py/commit/e0d186d754693feb7d27c2352e455c5febb4a5cd) was already merged in to bump pywin32 from 227 to 301. But, when installing latest 5.0.3 on windows machines is resulting in install of pywin32==227

Most likely extras_require needs updated
https://github.com/docker/docker-py/blob/a48a5a9647761406d66e8271f19fab7fa0c5f582/setup.py#L19

xoxys:
This issue exist on Python Poetry projects as well even if v5.0.3 is used already, https://github.com/thegeeklab/docker-tidy/security/dependabot/poetry.lock/pywin32/open

jackwhelpton:
Also being bitten by this. FYI, there's a further update to `pywin32`, to 302. 302 is the first version to support Python 3.10, so is a key thing for us... is that covered by this ticket, or should I cut a new one for pywin32==302?

tharun634:
> Also being bitten by this. FYI, there's a further update to `pywin32`, to 302. 302 is the first version to support Python 3.10, so is a key thing for us... is that covered by this ticket, or should I cut a new one for pywin32==302?

This ticket doesn't do it, but there already is a PR which covers it, #2857

gitpushdashf:
Any update on this? There's known vulnerabilities with pywin 227: https://security.snyk.io/vuln/SNYK-PYTHON-PYWIN32-1316226

Martin-Molinero:
Ping!
This issue means that you basically **can't use python 3.10 and docker in windows**. @StefanScherer @ulyssessouza @aiordache there are a few PRs already open to solve this, can they be reviewed/shipped anytime soon?

jmackay-godaddy:
Docker team, **PLEASE** can you accept any of the many already submitted fixes for this? People using docker through python and also on github cloud are getting security notifications due to your library. This bug has been open for 8 months.

aabmets:
I got around this problem by removing all docker software and using an alternative, as all docker software seems to have become abandonware and incompatible with the current Python version.

fproulx-boostsecurity:
> I got around this problem by removing all docker software and using an alternative, as all docker software seems to have become abandonware and incompatible with the current Python version.

Which alternative did you end up using @aabmets ?

[...]

PR #2972, opened 2022-03-29
Pywin32 upgrade

igor-sadchenko:
Fix issue #2902

@aiordache @ulyssessouza, please, accept this PR to fix this annoying bug


igor-sadchenko:
@StefanScherer could you help with this?

igor-sadchenko:
@ justincormack @thaJeztah maybe you know the answer who can help with that?
An annoying problem that is more than 6 months old, the fix of which is to update one dependency. Are you seriously?

PR #3006, opened 2022-07-26
utils: fix IPv6 address w/ port parsing

### LINKED ISSUE #2566: DeprecationWarning: urllib.parse.splitnport() is deprecated as of 3.8, use urllib.parse.urlparse() instead

milas:
This was using a deprecated function (`urllib.splitnport`),
ostensibly to work around issues with brackets on IPv6 addresses.

Ironically, its usage was broken, and would result in mangled IPv6
addresses if they had a port specified in some instances.

Usage of the deprecated function has been eliminated and extra test
cases added where missing. All existing cases pass as-is. (The only
other change to the test was to improve assertion messages.)

Fixes #2566.
Closes #2567 and closes #2944.

### DIFF
--- a/docker/utils/utils.py
+++ b/docker/utils/utils.py
@@ -1,4 +1,5 @@
 import base64
+import collections
 import json
 import os
 import os.path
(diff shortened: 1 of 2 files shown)


docker/utils/utils.py
```diff
@@ -351,7 +361,7 @@ def kwargs_from_env(ssl_version=None, assert_hostname=None, environment=None):
         # so if it's not set already then set it to false.
         assert_hostname = False
 
-    params['tls'] = tls.TLSConfig(
+    params['tls'] = TLSConfig(
```
milas, line 364: (This was a small change to prevent shadowing `tls` in the `parse_host` method)

nicksieger (approved):
LGTM

glours (approved):
LGTM

PR #3009, opened 2022-07-26
ci: add flake8 job

milas:
Project is already configured for flake8 but it never gets run in
CI.

(Since flake8 wasn't actually being run in CI, we'd accumulated some
violations, which are fixed here.)


nicksieger (approved):

glours (approved):
LGTM

PR #3012, opened 2022-07-27
ci: run SSH integration tests

milas:
Fix & enable SSH integration test suite.

This also adds a new test for connecting to unknown hosts when
using the Python SSH implementation (Paramiko). See #2932 for
more info.

Because of the above, some of the config/static key files have been
moved around and adjusted.


nicksieger (approved):
Seems good? A lot of duct tape, but if it gets more tests working, 👍

milas:
It's not _QUITE_ as bad as it seems - I moved some files around in the process to make it easier to tell what's client vs server and tried to eliminate some duplicates (e.g. `known_hosts` is constructed _from_ the client pubkey file so we don't need to maintain 2x copies), which made it look a bit more involved than it really is 😓

PR #3014, opened 2022-07-27
deps: upgrade & remove unnecessary dependencies

milas:
The `requirements.txt` and `setup.py` had a lot of extra transitive
dependencies to try and address various SSL shortcomings from the
Python ecosystem.

Thankfully, between modern Python versions (3.6+) and corresponding
`requests` versions (2.26+), this is all unnecessary now!

As a result, a bunch of transitive dependencies have been removed
from `requirements.txt`, the minimum version of `requests` increased,
and the `tls` extra made into a no-op.


README.md
```diff
@@ -10,9 +10,8 @@ The latest stable version [is available on PyPI](https://pypi.python.org/pypi/do
 
     pip install docker
 
-If you are intending to connect to a docker host via TLS, add `docker[tls]` to your requirements instead, or install with pip:
-
-    pip install docker[tls]
+> Older versions (< 6.0) required installing `docker[tls]` for SSL/TLS support.
```
milas, line 13: See #3013  - thinking we should release as v6.0.0 next 🙃

glours (approved):

Issue #2829, opened 2021-04-26
Add support for one-off jobs in Docker Swarm


pklejch:
Docker Engine 20.10 introduced option to run one-off jobs in Docker Swarm (https://docs.docker.com/engine/release-notes/#version-2010), this is achieved by two new modes for Swarm service: `replicated-job` and `global-job`. 

It would be nice to have this possibility when creating Swarm service in docker-py.

Related PR which brought support for those new modes into Docker CLI: https://github.com/docker/cli/pull/2262

feliperuhland:
Hi @pklejch 

Do you think you can open a pull request for that?

Have a nice day :)

pklejch:
Hi,
sure, I could look into it, but I don't know when I am going to have enough time ;)

ieugen:
This is also useful for ansible module https://github.com/ansible-collections/community.docker/issues/28 . 
Added link for better tracking.

disastrous-charly:
Hi,

Any update on this? Any PR has been made or could I help? 
It would be useful.

pklejch:
Hi,

I have partial, working but ugly prototype which kinda does what I needed to do. But I haven't able give it enough time to make it pretty for PR. Now with summer holidays and due to the fact that everything is opening again (because of coronavirus) I don't have that much time.

It would be nice, if someone could do it the right way.

frapell:
Hello, any progress on this? is that fix from @kinday good enough for submitting a PR ?

kinday:
My [preparation PR](https://github.com/docker/docker-py/pull/2947) got merged (kudos to @milas) so I’ll open PR with the fix for this issue soon.

PR #3016, opened 2022-07-28
Support `global-job` and `replicated-job` modes in Docker Swarm

kinday:
Fixes: #2829

[Docker Engine 20.10.0](https://docs.docker.com/engine/release-notes/#20100) introduced support for Swarm jobs via `replicated-job` and `global-job` service modes.

This PR adds support for these service modes.


milas (approved):
Thanks for the PR! 🥳

PR #3018, opened 2022-07-28
ci: add workflow for releases

milas:
GitHub Actions workflow to create a release: will upload to PyPI
and create a GitHub release with the `sdist` and `bdist_wheel`
as well.

The version code is switched to `setuptools_scm` to work well
with this flow (e.g. avoid needing to write a script that does
a `sed` on the version file and commits as part of release).


.github/workflows/release.yml
```diff
@@ -0,0 +1,37 @@
+name: Release
+
+on:
+  workflow_dispatch:
+    inputs:
+      tag:
+        description: "Release Tag (e.g. 6.0.0)"
```
nicksieger, line 7: Suggest putting the "v" in the example to make it clear what the tag will be

```suggestion
        description: "Release Tag (e.g. v6.0.0)"
```

.github/workflows/release.yml
```diff
@@ -0,0 +1,37 @@
+name: Release
+
+on:
+  workflow_dispatch:
+    inputs:
```
nicksieger, line 5: What do you think of adding a "dry-run" flag for testing the release, like in [compose-cli](https://github.com/docker/compose-cli/blob/main/.github/workflows/cli-release.yaml#L9)?

nicksieger (changes requested):
Mostly LGTM, the setuptools_scm stuff is opaque so I assume you've got it right and it's working

.github/workflows/release.yml
nicksieger: Nevermind, I see that the existing convention is w/o the `v` prefix

milas:
> the setuptools_scm stuff is opaque to me so I assume you've got it right and it's working

Ha yeah everything is pretty much lifted straight from their docs! It's funky but less crazy than most other solutions I've seen (and is "blessed" by pypa that maintains pip etc)

.github/workflows/release.yml
milas: Yes! I'll make this clearer in the description, that was why I gave the example w/o `v` - Python versions/tags do NOT want it as a prefix 😬

.github/workflows/release.yml
milas: Fancy! I stole it and made it fancier (✨ types ✨)

nicksieger (approved):

PR #3021, opened 2022-07-28
Replace deprecated distutils.spawn with shutil

n1ngu:
The distutils package is deprecated and slated for removal in Python 3.12.

The distutils.spawn.find_executable is replaced by shutil.which.

### DIFF
--- a/docker/credentials/utils.py
+++ b/docker/credentials/utils.py
@@ -1,15 +1,15 @@
-import distutils.spawn
 import os
+import shutil
 import sys
 
 
 def find_executable(executable, path=None):
     """
-    As distutils.spawn.find_executable, but on Windows, look up
+    As shutil.which, but on Windows, look up
     every extension declared in PATHEXT instead of just `.exe`
     """
     if sys.platform != 'win32':
-        return distutils.spawn.find_executable(executable, path)
+        return shutil.which(executable, path=path)
 
     if path is None:
         path = os.environ['PATH']


milas:
Thanks for the PR! It looks like your commit is missing a sign-off (DCO). There are instructions about [how to rebase and add it to your commit](https://github.com/docker/docker-py/pull/3021/checks?check_run_id=7566947835) on the failure.

milas (approved):
Thanks!

n1ngu:
Wooha didn't see that coming
```
________________ ERROR at setup of TestStore.test_store_and_get ________________
tests/integration/credentials/store_test.py:28: in setup_method
    self.store = Store('pass')
docker/credentials/store.py:18: in __init__
    self.exe = find_executable(self.program)
docker/credentials/utils.py:12: in find_executable
    return shutil.which(executable, path)
/usr/local/lib/python3.10/shutil.py:1474: in which
    if _access_check(name, mode):
/usr/local/lib/python3.10/shutil.py:1394: in _access_check
    return (os.path.exists(fn) and os.access(fn, mode)
E   TypeError: 'NoneType' object cannot be interpreted as an integer
```

Reading the shutil.which implementation, I can't imagine how the default `os.F_OK | os.X_OK` mode would result in a None :(

n1ngu:
Also, reading its docs more carefully https://docs.python.org/3/library/shutil.html#shutil.which

> On Windows, the current directory is always prepended to the path whether or not you use the default or provide your own, which is the behavior the command shell uses when finding executables. Additionally, when finding the cmd in the path, the PATHEXT environment variable is checked. For example, if you call shutil.which("python"), [which()](https://docs.python.org/3/library/shutil.html#shutil.which) will search PATHEXT to know that it should look for python.exe within the path directories.

It might be the time to completely remove this `find_executable` utility which does essentially the same.

n1ngu:
Ok I could debug it. It was the positional path parameter being confused for the mode parameter. Damn.

Passing it by keyword fixes the test suite in my workstation, would you be kind to approve the testing workflow again?

n1ngu:
Better merge #3028 and close this one IMHO.

milas:
See #3028

PR #3022, opened 2022-07-28
deps: upgrade websocket-client to latest

milas:
This is the last of the dependency upgrades - I didn't do it with the others because there was no test coverage for the websocket version of log streaming, so this PR adds a basic one along with the upgrade.


glours (approved):

milas:
FYI I had to mark the test as `xfail` for the SSL case -- it looks like that's actually broken right now when you're using `DOCKER_CERT_PATH ` because we don't seem to propagate the certs to the websocket connection 🙈

PR #3025, opened 2022-07-29
types: add missing log type enums

milas:
See https://docs.docker.com/engine/api/v1.41/#tag/Container/operation/ContainerCreate.

Closes #2849.

### DIFF
--- a/docker/types/containers.py
+++ b/docker/types/containers.py
@@ -15,9 +15,12 @@ class LogConfigTypesEnum:
         'journald',
         'gelf',
         'fluentd',
+        'awslogs',
+        'splunk',
+        'etwlogs',
         'none'
     )
-    JSON, SYSLOG, JOURNALD, GELF, FLUENTD, NONE = _values
+    JSON, SYSLOG, JOURNALD, GELF, FLUENTD, AWS, SPLUNK, ETW, NONE = _values
 
 
 class LogConfig(DictType):
--- a/tests/unit/dockertypes_test.py
+++ b/tests/unit/dockertypes_test.py
@@ -262,9 +262,13 @@ class UlimitTest(unittest.TestCase):
             Ulimit(name='hello', hard='456')
 
 
-class LogConfigTest(unittest.TestCase):
-    def test_create_host_config_dict_logconfig(self):
-        dct = {'type': LogConfig.types.SYSLOG, 'config': {'key1': 'val1'}}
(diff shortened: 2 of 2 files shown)


docker/types/containers.py
```diff
@@ -15,9 +15,12 @@ class LogConfigTypesEnum:
         'journald',
         'gelf',
         'fluentd',
+        'awslogs',
+        'splunk',
+        'etwlogs',
```
thaJeztah, line 20: Not sure if relevant for how it's used, but the list of available logging drivers may depend on the daemon; static binaries may omit some (IIRC), as well as differences between windows and linux daemons. In addition, additional logging drivers can be installed as plugins.

The list of logging drivers can also be found in the `/info` endpoint (but the `/info` endpoint itself can be somewhat "heavy" as it collects info from many bits the daemon manages, including container states etc.

```bash
curl -s --unix-socket /var/run/docker.sock 'http://localhost/info' | jq .Plugins.Log
[
  "awslogs",
  "fluentd",
  "gcplogs",
  "gelf",
  "journald",
  "json-file",
  "local",
  "logentries",
  "splunk",
  "syslog"
]
```

thaJeztah: Basically; it's ok to have these for convenience, but it may be good to leave actual validation to the daemon API.

Hm... and just when I suggest that; I think there's something to look into there, as it seems like doing so takes a long time (some timeout hit after 15 seconds to look up the (plugin) driver?);

```
time docker create --name foo --log-driver=nosuchdriver alpine
Error response from daemon: error looking up logging plugin nosuchdriver: plugin "nosuchdriver" not found

real	0m15.048s
user	0m0.014s
sys	0m0.027s
Error response from daemon: error looking up logging plugin nosuchdriver: plugin "nosuchdriver" not found
```

Let me see if there's a ticket for that problem.

milas: Ah, that's good to know! There's no _new_ validation here, but it was already checking to see if it was one of the enum entries 🙈 

I'll probably leave that as-is for now but look at opening a follow-up PR that either just removes that validation or makes the log type field a `typing.Union[LogConfigTypesEnum, str, None]` so that a string can be passed for other types such as from plugins.

thaJeztah: Ah, yes that `str` makes sense (I'm not much of a Python coder, so never know what to suggest for these things 😂 )

Issue #2825, opened 2021-04-19
container logs time precision


fopinappb:
https://github.com/docker/docker-py/blob/master/docker/api/container.py#L845

This discards any microseconds from datetime while the API allows floats to be sent with the same precision as shown in the timestamps of the output (10^-9).

Even if datetime only keeps 10^-9, it'd still be nice to use that precision instead of just seconds.

https://github.com/moby/moby/issues/41784#issuecomment-822371593

PR #3031, opened 2022-08-17
feat: add support for floats to docker logs params since / until sinc…

ArchiMoebius:
feature to add support for floats to docker container logs method params `since / until` due to the fact that [docker engine supports it](https://github.com/moby/moby/issues/41784#issuecomment-822371593)

Related:
https://github.com/docker/docker-py/issues/1515
https://github.com/docker/docker-py/issues/2825

### DIFF
--- a/docker/api/container.py
+++ b/docker/api/container.py
@@ -826,11 +826,12 @@ class ContainerApiMixin:
             tail (str or int): Output specified number of lines at the end of
                 logs. Either an integer of number of lines or the string
                 ``all``. Default ``all``
-            since (datetime or int): Show logs since a given datetime or
-                integer epoch (in seconds)
+            since (datetime, int, or float): Show logs since a given datetime,
+                integer epoch (in seconds) or float (in nanoseconds)
             follow (bool): Follow log output. Default ``False``
-            until (datetime or int): Show logs that occurred before the given
-                datetime or integer epoch (in seconds)
+            until (datetime, int, or float): Show logs that occurred before
+                the given datetime, integer epoch (in seconds), or
+                float (in nanoseconds)
 
         Returns:
             (generator or str)
@@ -855,6 +856,8 @@ class ContainerApiMixin:
                 params['since'] = utils.datetime_to_timestamp(since)
             elif (isinstance(since, int) and since > 0):
                 params['since'] = since
+            elif (isinstance(since, float) and since > 0.0):
+                params['since'] = since
             else:
                 raise errors.InvalidArgument(
                     'since value should be datetime or positive int, '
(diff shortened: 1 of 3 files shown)


docker/api/container.py
```diff
@@ -870,6 +873,8 @@ def logs(self, container, stdout=True, stderr=True, stream=False,
                 params['until'] = utils.datetime_to_timestamp(until)
             elif (isinstance(until, int) and until > 0):
                 params['until'] = until
+            elif (isinstance(until, float) and until > 0.0):
+                params['until'] = until
             else:
                 raise errors.InvalidArgument(
                     'until value should be datetime or positive int, '
```
milas, line 880:
```suggestion
                    'until value should be datetime or positive int/float, '
```

docker/api/container.py
```diff
@@ -855,6 +856,8 @@ def logs(self, container, stdout=True, stderr=True, stream=False,
                 params['since'] = utils.datetime_to_timestamp(since)
             elif (isinstance(since, int) and since > 0):
                 params['since'] = since
+            elif (isinstance(since, float) and since > 0.0):
+                params['since'] = since
             else:
                 raise errors.InvalidArgument(
                     'since value should be datetime or positive int, '
```
milas, line 863:
```suggestion
                    'since value should be datetime or positive int/float, '
```

milas (changes requested):
Thanks for the PR!

Looks good -- just a couple tiny suggestions to update the error messages for unsupported types.

docker/api/container.py
```diff
@@ -826,11 +826,12 @@ def logs(self, container, stdout=True, stderr=True, stream=False,
             tail (str or int): Output specified number of lines at the end of
                 logs. Either an integer of number of lines or the string
                 ``all``. Default ``all``
-            since (datetime or int): Show logs since a given datetime or
-                integer epoch (in seconds)
+            since (datetime, int, or float): Show logs since a given datetime,
+                integer epoch (in seconds) or float (in nanoseconds)
```
milas, line 830:
```suggestion
            since (datetime, int, or float): Show logs since a given datetime,
                integer epoch (in seconds) or float (in fractional seconds)
```

docker/api/container.py
```diff
@@ -826,11 +826,12 @@ def logs(self, container, stdout=True, stderr=True, stream=False,
...
             follow (bool): Follow log output. Default ``False``
-            until (datetime or int): Show logs that occurred before the given
-                datetime or integer epoch (in seconds)
+            until (datetime, int, or float): Show logs that occurred before
+                the given datetime, integer epoch (in seconds), or
+                float (in nanoseconds)
```
milas, line 834:
```suggestion
            until (datetime, int, or float): Show logs that occurred before
                the given datetime, integer epoch (in seconds), or
                float (in fractional seconds)
```

milas (changes requested):
Ah another thing - "in nanoseconds" is misleading

milas (approved):
Thanks for the PR!

PR #3035, opened 2022-08-19
Use latest stable syntax for Dockerfiles

thaJeztah:
I noticed one Dockerfile was pinned to 1.4; given that there's a
backward compatibility guarantee on the stable syntax, the general
recommendation is to use `dockerfile:1`, which makes sure that the
latest stable release of the Dockerfile syntax is pulled before
building.

While changing, I also made some minor changes to some Dockerfiles
to reduce some unneeded layers.

### DIFF
--- a/tests/Dockerfile
+++ b/tests/Dockerfile
@@ -1,5 +1,7 @@
-# syntax = docker/dockerfile:1.4
+# syntax=docker/dockerfile:1
+
 ARG PYTHON_VERSION=3.10
+
 FROM python:${PYTHON_VERSION}
 
 ARG APT_MIRROR
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,3 +1,5 @@
+# syntax=docker/dockerfile:1
+
 ARG PYTHON_VERSION=3.10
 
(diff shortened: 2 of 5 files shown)


tests/Dockerfile
```diff
@@ -1,5 +1,7 @@
-# syntax = docker/dockerfile:1.4
+# syntax=docker/dockerfile:1
```
thaJeztah, line 1: Note that the whitespace is allowed in these; I'm somewhat trying to use a consistent formatting in our docs and repositories (`#<space>syntax=<syntax>`); consistency helps with searching for occurrences, as well as might avoid questions from casual visitors (why does the documentation show X, and here it shows X?)

thaJeztah, line 1: Note that the whitespace is allowed in these; I'm somewhat trying to use a consistent formatting in our docs and repositories (`#<space>syntax=<syntax>`); consistency helps with searching for occurrences, as well as might avoid questions from casual visitors (why does the documentation show X, and here it shows X?)

milas (approved):
✨

Issue #3045, opened 2022-09-20
Docker 4.12.0 on Windows throws "The pipe has been ended" error


mndeveci:
Hi team,

We've started to see error messages with Docker Desktop v4.12.0 on Windows. We are attaching to the container by calling `{container_instance}.attach(stream=True, logs=True, demux=True)` and then iterating through the logs streams. This is working fine with earlier versions but it started to throw following error with Docker Desktop v4.12.0.

```
Traceback (most recent call last):
  File "c:\users\administrator\workplace\aws-sam-cli\samcli\local\docker\container.py", line 401, in _write_container_output
    for stdout_data, stderr_data in output_itr:
  File "c:\users\administrator\workplace\aws-sam-cli\venv\lib\site-packages\docker\types\daemon.py", line 32, in __next__
    return next(self._stream)
  File "c:\users\administrator\workplace\aws-sam-cli\venv\lib\site-packages\docker\api\client.py", line 406, in <genexpr>
    gen = (demux_adaptor(*frame) for frame in gen)
  File "c:\users\administrator\workplace\aws-sam-cli\venv\lib\site-packages\docker\utils\socket.py", line 94, in frames_iter_no_tty
    (stream, n) = next_frame_header(socket)
  File "c:\users\administrator\workplace\aws-sam-cli\venv\lib\site-packages\docker\utils\socket.py", line 66, in next_frame_header
    data = read_exactly(socket, 8)
  File "c:\users\administrator\workplace\aws-sam-cli\venv\lib\site-packages\docker\utils\socket.py", line 51, in read_exactly
    next_data = read(socket, n - len(data))
  File "c:\users\administrator\workplace\aws-sam-cli\venv\lib\site-packages\docker\utils\socket.py", line 35, in read
    return socket.recv(n)
  File "c:\users\administrator\workplace\aws-sam-cli\venv\lib\site-packages\docker\transport\npipesocket.py", line 23, in wrapped
    return f(self, *args, **kwargs)
  File "c:\users\administrator\workplace\aws-sam-cli\venv\lib\site-packages\docker\transport\npipesocket.py", line 117, in recv
    err, data = win32file.ReadFile(self._handle, bufsize)
pywintypes.error: (109, 'ReadFile', 'The pipe has been ended.')
```

A simplified version of how we are reading the logs;
```
# stdout and stderr are StreamWriter's that is been provided earlier
logs_itr = container.attach(stream=True, logs=True, demux=True)

[...]

PR #3056, opened 2022-10-21
socket: handle npipe close

nicks:
Fixes https://github.com/docker/docker-py/issues/3045

I tried to do something minimally invasive (e.g., not importing pywintypes)

Signed-off-by: Nick Santos <nick.santos@docker.com>


nicks:
cc @milas

milas (approved):

PR #3073, opened 2022-12-01
Avoid stripping characters in volume names

loicleyendecker:
When a volume name ends in "r", "o", or "w", it currently gets stripped which results in a mount created on the host with an incorrect name. The fix simply consists in not using the `rstrip` method but manually checking for ":rw" and ":ro" at the end of the string.

Signed-off-by: Loïc Leyendecker <loic.leyendecker@gmail.com>


milas (approved):
Thanks for the PR! Nice find 🙃

Issue #3070, opened 2022-11-22
Could you not check for `credHelpers` on init?


nomagick:
## Request
Could you change the implementation a bit to not check for `credHelpers` binaries on init?

## Context
We are kind of a PaaS/SaaS provider, and we are providing a `credHelper` for a particular registry.

However, the helper script, namely `docker-credential-[the-thing]`, is part of a Python package.

## Motivation
Contrary to the system `credStore` plugins, which are typically binary and installed to system PATH statically,
executables from Python packages tend to have some "Dynamicity", for example, managed by tools like `pyenv` or `conda` and have something to do with "active profile" configuration.  

Now that the `credHelper` plugin is somewhat dynamic, the Docker configuration file, namely `DOCKER_CONFIG/config.json` is not expected to really change with the "active python environment" settings.

So it comes to a situation where people install our `credHelper` in a particular python environment and modifies the Docker configuration file to use the helper plugin. Then they move to some other python environment which does not have the helper plugin installed. 

When people try to load `docker-py` from the environment without the plugin, it fails with `docker.credentials.errors.InitializationError: docker-credential-[name-of-plugin] not installed or not available in PATH`.

This kind of error does not trigger when people use the command-line docker client.

milas:
It seems reasonable to emit a [warning](https://docs.python.org/3/library/warnings.html#warnings.warn) here instead. Any operations that require authentication will end up failing regardless.

Happy to review a PR for this.

PR #3080, opened 2022-12-22
fix(store): warn on init instead of throw

nomagick:
This tries to resolve https://github.com/docker/docker-py/issues/3070


milas (approved):
Thanks for the PR!

Issue #2896, opened 2021-10-06
Feature Request: `driver_opt` parameter of `Networks.connect` in `Containers.create`


Skazza94:
Currently,
the `Networks.connect` method takes the `driver_opt` parameter in order to pass to the driver some custom values: [https://docker-py.readthedocs.io/en/stable/networks.html#docker.models.networks.Network.connect](https://docker-py.readthedocs.io/en/stable/networks.html#docker.models.networks.Network.connect).
I have a custom Network Plugin which uses such parameters to generate the MAC Address.

However, it is not possible to specify any custom value while attaching the first Docker network using [`Containers.create` or `Containers.run`](https://docker-py.readthedocs.io/en/stable/containers.html#docker.models.containers.ContainerCollection.run).

Is it possible to expose this parameter in such methods?

Thanks.

Skazza94:
Hi, any chance to have an update on this? Thanks.

tcaiazzi:
Hi, 
any news about this thread?

PR #3083, opened 2022-12-27
Add `network_driver_opt` to client.containers run and create

Skazza94:
This pull request adds an additional parameter to `client.containers.run` and `client.containers.create` called `network_driver_opt`. 

This parameter allows to specify a dict which allows to pass to the driver some custom values, and it is the same already present in [`Network.connect`](https://docker-py.readthedocs.io/en/stable/networks.html#docker.models.networks.Network.connect). 

I requested this feature in #2896, but it was never implemented. 

I think it could be useful to also export other parameters of `networking_config`.

### DIFF
--- a/docker/models/containers.py
+++ b/docker/models/containers.py
@@ -678,9 +678,10 @@ class ContainerCollection(Collection):
                   This mode is incompatible with ``ports``.
 
                 Incompatible with ``network``.
-            network_driver_opt (dict): A dictionary of options to provide to the
-                network driver. Defaults to ``None``. Used in conjuction
-                with ``network``. Incompatible with ``network_mode``.
+            network_driver_opt (dict): A dictionary of options to provide
+                to the network driver. Defaults to ``None``. Used in
+                conjuction with ``network``. Incompatible
+                with ``network_mode``.
             oom_kill_disable (bool): Whether to disable OOM killer.
             oom_score_adj (int): An integer value containing the score given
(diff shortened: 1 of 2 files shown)


milas (commented):
Thanks for the PR! This looks good to me, I just need you to sign-off on the commits before I can accept it. There's instructions at https://github.com/docker/docker-py/pull/3083/checks?check_run_id=10326470472 if you're unfamiliar with the process.

Also, I agree re: exposing more of `networking_config`. It doesn't feel great to piecemeal expose parts of it, but this is consistent with the other methods (e.g. [`Network::connect()`](https://docker-py.readthedocs.io/en/6.0.1/networks.html#docker.models.networks.Network.connect)) as you pointed out, so let's go with it! 🙃

Skazza94:
Hi @milas,
thanks for reviewing the PR 😄. I signed the commits.

> Also, I agree re: exposing more of `networking_config`. It doesn't feel great to piecemeal expose parts of it, but this is consistent with the other methods (e.g. [`Network::connect()`](https://docker-py.readthedocs.io/en/6.0.1/networks.html#docker.models.networks.Network.connect)) as you pointed out, so let's go with it! 🙃

I agree that it is consistent with the `Network.connect` method. As soon as I have time, I will also export the remaining network parameters to `client.containers` `run` and `create`.

milas (approved):

PR #3085, opened 2023-01-02
Add types

### LINKED ISSUE #2796: Type annotations
Would be nice to have typing annotations to support static typing in projects with mypy.

Viicos:
Fixes #2796.

- [ ] `api`
- [ ] `context`
- [ ] `credentials`
- [ ] `models`
- [ ] `transport`
- [ ] `types`
- [ ] `utils`
- [x] `auth.py`
- [x] `client.py`
- [x] `constants.py`
- [x] `errors.py`
- [x] `tls.py`
- [x] `version.py`

I have used `Dict[str, Any]` in `auth.py` and `client.py` in some places, we might want to narrow the types using `TypedDict` instead (or other solutions, can be discussed).

Notes:

- As adding types will add a lot of imports and changes to the code layout, `isort` and a code formatter could be enforced.

- I suggest we change `json` and `binary` to keyword only arguments. If used as an argument, this could lead to confusion (see [here](https://github.com/docker/docker-py/blob/82cf559b5a641f53e9035b44b91f829f3b4cca80/docker/api/daemon.py#L95) for example).
https://github.com/docker/docker-py/blob/82cf559b5a641f53e9035b44b91f829f3b4cca80/docker/api/client.py#L272

- `config_path` and `config_dict` can be made `Optional`, as this classmethod is already called with optional arguments [here](https://github.com/docker/docker-py/blob/82cf559b5a641f53e9035b44b91f829f3b4cca80/docker/auth.py#L358).
https://github.com/docker/docker-py/blob/82cf559b5a641f53e9035b44b91f829f3b4cca80/docker/auth.py#L153-L164


docker/api/client.py
```diff
@@ -1,10 +1,15 @@
 import json
 import struct
 import urllib
+import ssl
 from functools import partial
+from typing import Any, AnyStr, Optional, Union, Dict, overload, NoReturn, Iterator
+
+from typing_extensions import Literal
```
Dreamsorcerer, line 8: Better to only require typing_extensions on older Python versions.
```suggestion
if sys.version_info >= (3, 8):
    from typing import Literal
else:
    from typing_extensions import Literal
```

GhostLyrics:
What I don't see in this MR is something like instructions for how to check what you added with either mypy or pyright, a test  target for a makefile, info in a Readme or literally anything that can help the next person avoid regressions or help them continue your work, respectively.

Could you check to see what conventions (e.g. auto tests, makefile, similar) are used by the project and try to throw together the simplest integration you could think of? Ask for help if you can't figure it out yourself, that's fine :)

Viicos:
> What I don't see in this MR is:
> 
>     * something like instructions for how to check what you added with either mypy or pyright
> 
>     * a test  target for a makefile
> 
>     * info in a Readme
> 
>     * or literally anything that can help the next person avoid regressions or help them continue your work, respectively.
> 
> 
> Could you check to see what conventions (e.g. auto tests, makefile, similar) are used by the project and try to throw together the simplest integration you could think of? Ask for help if you can't figure it out yourself, that's fine :)

As I said [here](https://github.com/docker/docker-py/pull/3085#issue-1516669374), I was (and still am) waiting for approval from the maintainers of the project. I don't want to add types directly into the project and then see this PR declined because maintainers doesn't want to use them. That's why we are thinking about adding them to `typeshed` (see [discussion](https://github.com/docker/docker-py/issues/2796)), and I will surely do it if we don't have any feedback from maintainers in the next few days

.github/workflows/ci.yml
```diff
@@ -13,9 +13,11 @@ jobs:
             - uses: actions/setup-python@v4
               with:
                   python-version: '3.11'
-            - run: pip install -U ruff==0.0.284
+            - run: pip install -U ruff==0.0.284 mypy==1.5.1
             - name: Run ruff
               run: ruff docker tests
+            - name: Run mypy
+              run: mypy docker
```
Dreamsorcerer, line 20: I'd suggest putting the files in the config and then only running `mypy`.
I'd also suggest copying, for example, this config, to get more in depth checking: https://github.com/aio-libs/aiohttp-jinja2/blob/master/.mypy.ini
Commenting out any options that you don't want to deal with yet, so they can be added in later (note you can also do config options per module).

Viicos: Or maybe we should go with mypy strict in one go?

Dreamsorcerer: strict just enables about half the options listed in that config (there are several other useful ones not enabled by strict), so I'd still look at the whole thing. And if you don't want to resolve everything in 1 PR, then more finegrained settings is going to help.

Viicos: I see, I'll take a look. I'll see if CI passes once I convert this PR as ready to review. But I don't think maintainers will see..

PR #3089, opened 2023-01-25
Add `one-shot` to container API `stats`

aroxby-wayscript:
Add the `one-shot` parameter to container stats to enable pulling just a single stats update.  New in docker engine [20.10.0](https://docs.docker.com/engine/release-notes/#20100).

See the moby PR: https://github.com/moby/moby/pull/40478

### DIFF
--- a/docker/api/container.py
+++ b/docker/api/container.py
@@ -1158,6 +1158,11 @@ class ContainerApiMixin:
                 )
             params['one-shot'] = one_shot
         if stream:
+            if one_shot is not None:
+                raise errors.InvalidArgument(
+                    'one_shot is only available in conjunction with '
+                    'stream=False'
+                )
             return self._stream_helper(self._get(url, params=params),
                                        decode=decode)
         else:
--- a/docker/api/container.py
+++ b/docker/api/container.py
@@ -1126,7 +1126,7 @@ class ContainerApiMixin:
         self._raise_for_status(res)
 
     @utils.check_resource('container')
-    def stats(self, container, decode=None, stream=True, one_shot=False):
+    def stats(self, container, decode=None, stream=True, one_shot=None):
         """
         Stream statistics for a specific container. Similar to the
         ``docker stats`` command.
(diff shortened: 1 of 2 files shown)


milas (changes requested):
Thanks for the PR!

Two adjustments:

1. Since this was added in v1.41, we should check the API version if the param is set, which will realistically mean changing the default to `one_shot=None`. See [an example in another API call](https://github.com/docker/docker-py/blob/34e6829dd40e99e9ba47ea02fcfabda09e08d36e/docker/api/container.py#L441-L446)
2. `one_shot=True` is only valid if `stream=False`, so we should check that as an invariant here and raise an exception, as otherwise it'll be rather hard to debug because of the stream helper in some cases (normally I'd say rely on server-side validation, but in this case there's no sensible meaning to the combo, so it seems safe enough to add a redundant check for)

aroxby-wayscript:
@milas, thanks for that very helpful and clear feedback.  I believe this is what you'd want to see.

docker/api/container.py
```diff
@@ -1138,23 +1138,39 @@ def stats(self, container, decode=None, stream=True):
...
+        }
+        if one_shot is not None:
+            if utils.version_lt(self._version, '1.41'):
+                raise errors.InvalidVersion(
+                    'one_shot is not supported for API version < 1.41'
+                )
+            params['one-shot'] = one_shot
         if stream:
-            return self._stream_helper(self._get(url, stream=True),
+            if one_shot is not None:
```
milas, line 1161:
```suggestion
            if one_shot:
```
It'd also be valid to explicitly pass `one_shot=False`

milas (changes requested):
One small tweak but LGTM otherwise. Thanks for making the changes!

aroxby-wayscript:
@milas thanks again for all your helpful feedback.  How are we looking on this?

aroxby-wayscript:
@milas just sending another ping on this.  Look good?

milas (approved):

PR #3090, opened 2023-01-26
Fix typos

kianmeng:
Found via `codespell -L keep-alives`


milas (commented):
Thanks for the fixes! You'll need to sign-off on your commit before I can accept it, however.

The GitHub checks for the PR has instructions on how to do this: https://github.com/docker/docker-py/pull/3090/checks?check_run_id=10894214150

Issue #2416, opened 2019-08-26
Custom SSH Credentials


dnut:
I can't rely on an ssh agent to provide a private key. Sometimes I need to provide a password, sometimes I need to manually load the key file. But the docker-py code only connects with this in SSHHTTPAdapter:
```python
self.ssh_client.connect(
    parsed.hostname, parsed.port, parsed.username,
)
```
There is no flexibility here. How am I supposed to connect with a custom key or password? I thought I might hack my way in by reassigning APIClient.\_custom\_adapter to my own subclass of SSHHTTPAdapter, but then I realized the APIClient.\_\_init__ is a huge mess that does way too much. That method would always raise an exception so I would also have to totally reimplement that method in a subclass. This is too much maintenance overhead for my deployment script that I would like to keep as simple as possible.

It should be exposed in SSHHTTPAdapter. I might even recommend doing both of these:
1. Add optional key and password parameters to SSHHTTPAdapter so you don't have to write your own subclass of it just to use custom credentials
2. Add optional ssh_client parameter to SSHHTTPAdapter to be used instead of instantiating one in \_\_init__.

It should also be exposed somehow in APIClient.\_\_init__, for example:
1. Expose password and private key with...
    a. Explicit parameters (a bit messy considering this is not an SSH specific class)
    b. General purpose configuration object (dict or better yet custom configuration class)
2. Enable programmer to provide their own SSHClient to APIClient. This is not ideal for the same reason as 1a.
3. Enable programmer to provide their own instance of a BaseHTTPAdapter to be used as self._custom_adapter. I think this is the best solution but there are some open questions. Like what's the best way to handle all the usages of base_url when it is no longer required to establish a connection.

I'm happy to take the lead on this and submit a PR, but I would like to get some feedback first.

cmcga1125:
@dnut FYI I just got SSH to work to answer your question.  It's not documented and i had to read through some of the source code but here are a few things i learned.
1. If you're wanting to use an ssh key, it has to be in the ~/.ssh/id_rsa location - or - /root/.ssh/id_rsa (if alpine)
2. If you don't have a known_hosts file you'll get a warning but the connection actually connects.
3. the syntax for the base_url is like mentioned above - which as a great help :)

dnut:
@cmcga1125 I can get SSH to work on my desktop, but I need this to run on our build servers with custom credentials.

[...]

PR #3092, opened 2023-01-27
Adding Support for encrypted SSH Private Keys(Resolving #2416)

devmaxde:
Resolving #2416 by passing SSH Filename and Passphrase


mibkvr:
I didn't update the docs tho. Some should do that :)

PR #3093, opened 2023-02-01
Add `status` parameter to services list API

lorinbucher:
Adds the optional `status` parameter to the services list API. If true, the services returned will include `ServiceStatus`, which provides Desired, Running, and Completed task counts for the service.

This has been added in 20.10.0, with API version 1.41. ([Release Notes](https://docs.docker.com/engine/release-notes/#20100))
See moby PR: [moby/moby#39231](https://github.com/moby/moby/pull/39231)

### DIFF
--- a/docker/api/service.py
+++ b/docker/api/service.py
@@ -262,7 +262,7 @@ class ServiceApiMixin:
         return True
 
     @utils.minimum_version('1.24')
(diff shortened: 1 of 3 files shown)


milas (approved):
Thanks for the PR!

PR #3094, opened 2023-02-03
Support checkpoint restore

sdimovv:
Added support for the experimental checkpoint-restore docker API.

Couple of notes:
1. I have intentionally omitted adding a dedicated method for checking if the experimental features are enabled in the docker daemon, since this can be deduced from the version information (`self.client.version()`)
2. I have added docstrings to all new classes and methods, not sure if there is a need to manually add any docs? Can fix if needed.

### DIFF
--- a/docker/api/container.py
+++ b/docker/api/container.py
@@ -670,6 +670,93 @@ class ContainerApiMixin:
         return EndpointConfig(self._version, *args, **kwargs)
 
     @utils.check_resource('container')
+    def container_checkpoints(self, container, checkpoint_dir=None):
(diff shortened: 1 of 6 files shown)


sdimovv:
Hi @aiordache and @ulyssessouza,

does this need anything else to get merged? I'll be happy to resolve any issues you might have.

lumost:
Thanks for this! I'll be using this CR :)

DrShushen:
Ah, shame that this hasn't been merged so far, it's exactly what I'm looking for

PR #3107, opened 2023-04-04
fix: exec_run -- default user-root

### LINKED ISSUE #2979: Documentation for exec_run()
The default user is an empty string, but the docs say it is root. I had to manually specify both user as root and privileged as True in order to run commands as root on a container object. I'm unsure whether that might be a bug in and of itself.

https://docker-py.readthedocs.io/en/stable/containers.html#docker.models.containers.Container.exec_run
The relevant documentation appears to be the same in 4.1.0 (installed), stable, and latest.

Khushiyant:
Fix #2979

According to [Docker docs](https://docker-py.readthedocs.io/en/stable/containers.html#docker.models.containers.Container.exec_run), `exec_run()` has user - `default: root` whereas `exec_run()` has `default:None` in its implementation

Changes Made:
```diff
- privileged=False, user='', detach=False, stream=False,
+ privileged=False, user='root', detach=False, stream=False,
```
> The relevant documentation appears to be the same in 4.1.0 (installed), stable, and latest.

### DIFF
--- a/docker/models/containers.py
+++ b/docker/models/containers.py
@@ -150,7 +150,7 @@ class Container(Model):
         return self.client.api.diff(self.id)
 
     def exec_run(self, cmd, stdout=True, stderr=True, stdin=False, tty=False,
-                 privileged=False, user='', detach=False, stream=False,
+                 privileged=False, user='root', detach=False, stream=False,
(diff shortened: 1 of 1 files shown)


Khushiyant:
@aiordache @ulyssessouza Please review

PR #3112, opened 2023-04-23
Use windows api to honour timeouts (fixes #3111)

### LINKED ISSUE #3111: Timeouts don't work on windows
Currently the windows npipe implementation doesn't honour timeouts. Regardless of which api endpoint you use or pretty much anything else this leads to bugs where the docker api waits until the docker daemon finishes instead of timing out properly.

For example, if there is a dockerfile containing at `timeout/`
```
FROM alpine

RUN sleep 1000
```
and you run
```python
from docker import DockerClient

DockerClient.from_env().images.build(path="timeout/", timeout=3)
```
python will hang for the full 1000 seconds instead of raising an error after 3.

Version info:  
docker-py: 6.0.1  
python: 3.11.3  
docker:
Client:
 Cloud integration: v1.0.24
 Version:           20.10.14
 API version:       1.41
 Go version:        go1.16.15
 Git commit:        a224086
 Built:             Thu Mar 24 01:53:11 2022        
 OS/Arch:           windows/amd64
 Context:           default
 Experimental:      true

Server: Docker Desktop 4.8.1 (78998)
 Engine:
  Version:          20.10.14
  API version:      1.41 (minimum version 1.12)     
  Go version:       go1.16.15
  Git commit:       87a90dc
  Built:            Thu Mar 24 01:46:14 2022        
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          1.5.11
  GitCommit:        3df54a852345ae127d1fa3092b95168e4a88e2f8
 runc:
  Version:          1.0.3
  GitCommit:        v1.0.3-0-gf46b6ba
 docker-init:
  Version:          0.19.0
  GitCommit:        de40ad0

ImogenBits:
This fixes #3111 by using the overlapped file IO mode and associated wait windows api functions.

### DIFF
--- a/docker/transport/npipesocket.py
+++ b/docker/transport/npipesocket.py
@@ -188,7 +188,7 @@ class NpipeSocket:
             raise ValueError('Timeout value out of range')
         else:
             # Timeout mode - Value converted to milliseconds
-            self._timeout = value * 1000
+            self._timeout = int(value * 1000)
 
     def gettimeout(self):
         return self._timeout
--- a/docker/transport/npipesocket.py
+++ b/docker/transport/npipesocket.py
@@ -156,8 +156,15 @@ class NpipeSocket:
 
     @check_closed
     def send(self, string, flags=0):
-        err, nbytes = win32file.WriteFile(self._handle, string)
-        return nbytes
+        event = win32event.CreateEvent(None, True, True, None)
+        overlapped = pywintypes.OVERLAPPED()
+        overlapped.hEvent = event
+        win32file.WriteFile(self._handle, string, overlapped)
+        wait_result = win32event.WaitForSingleObject(event, self._timeout)
+        if wait_result == win32event.WAIT_TIMEOUT:
+            win32file.CancelIo(self._handle)
+            raise TimeoutError
+        return win32file.GetOverlappedResult(self._handle, overlapped, 0)
 
     @check_closed
     def sendall(self, string, flags=0):
--- a/docker/transport/npipesocket.py
+++ b/docker/transport/npipesocket.py
@@ -4,6 +4,8 @@ import io
 
 import win32file
 import win32pipe
+import pywintypes
+import win32event
 
 cERROR_PIPE_BUSY = 0xe7
 cSECURITY_SQOS_PRESENT = 0x100000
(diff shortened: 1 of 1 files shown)


docker/transport/npipesocket.py
```diff
@@ -131,22 +133,38 @@ def recv_into(self, buf, nbytes=0):
...
+            win32file.CancelIo(self._handle)
+            raise TimeoutError
+        return win32file.GetOverlappedResult(self._handle, overlapped, 0)
 
     def _recv_into_py2(self, buf, nbytes):
+        raise RuntimeError
         err, data = win32file.ReadFile(self._handle, nbytes or len(buf))
         n = len(data)
         buf[:n] = data
         return n
```
milas, line 156:
```suggestion
```

I just double-checked, this used to be gated by a `six` call, must have been missed when the rest of the Python2 code went away, but it can be removed entirely now

docker/transport/npipesocket.py
```diff
@@ -131,22 +133,38 @@ def recv_into(self, buf, nbytes=0):
         if not isinstance(buf, memoryview):
             readbuf = memoryview(buf)
 
+        event = win32event.CreateEvent(None, True, True, None)
```
milas, line 136: This will leak the event handle (though it looks like `win32event` will close it as part of GC).

Can we do both of these with a `try` / `finally` block with a call to `win32api.CloseHandle(event)`?

milas (changes requested):
See note about ensuring we close the handle explicitly, but LGTM otherwise! Thanks for the PR!

docker/transport/npipesocket.py
ImogenBits: alright, I removed the function entirely

docker/transport/npipesocket.py
ImogenBits: I added the try/finally block, tbh I'm not super familiar with the python windows api so I'm not sure if I added them at the right spots

milas: What you did looks great!

---

(On the Python side, `finally` will execute even if an exception gets thrown, so this ensures that the Windows API handle we create is released ASAP and even if we throw. On the Windows API side, lots of things use handles and it's good to not leak them 😅 )

milas (approved):

docker/transport/npipesocket.py
milas: Thanks for doing the extra clean-up ✨

Issue #3113, opened 2023-04-26
urllib3 v2 incompatibility


zachmullen:
Minimal repro:

```python
Python 3.8.10 (default, Mar 13 2023, 10:26:41) 
[GCC 9.4.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import docker
/opt/worker_venv/lib/python3.8/site-packages/requests/__init__.py:109: RequestsDependencyWarning: urllib3 (2.0.0) or chardet (None)/charset_normalizer (3.1.0) doesn't match a supported version!
  warnings.warn(
>>> client = docker.from_env(version='auto')
Traceback (most recent call last):
  File "/opt/worker_venv/lib/python3.8/site-packages/docker/api/client.py", line 214, in _retrieve_server_version
    return self.version(api_version=False)["ApiVersion"]
  File "/opt/worker_venv/lib/python3.8/site-packages/docker/api/daemon.py", line 181, in version
    return self._result(self._get(url), json=True)
  File "/opt/worker_venv/lib/python3.8/site-packages/docker/utils/decorators.py", line 46, in inner
    return f(self, *args, **kwargs)
  File "/opt/worker_venv/lib/python3.8/site-packages/docker/api/client.py", line 237, in _get
    return self.get(url, **self._set_request_timeout(kwargs))
  File "/opt/worker_venv/lib/python3.8/site-packages/requests/sessions.py", line 600, in get
    return self.request("GET", url, **kwargs)
  File "/opt/worker_venv/lib/python3.8/site-packages/requests/sessions.py", line 587, in request
    resp = self.send(prep, **send_kwargs)
  File "/opt/worker_venv/lib/python3.8/site-packages/requests/sessions.py", line 701, in send
    r = adapter.send(request, **kwargs)
  File "/opt/worker_venv/lib/python3.8/site-packages/requests/adapters.py", line 487, in send
    resp = conn.urlopen(
  File "/opt/worker_venv/lib/python3.8/site-packages/urllib3/connectionpool.py", line 790, in urlopen
    response = self._make_request(
  File "/opt/worker_venv/lib/python3.8/site-packages/urllib3/connectionpool.py", line 496, in _make_request
    conn.request(
TypeError: request() got an unexpected keyword argument 'chunked'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
[...]

PR #3116, opened 2023-05-02
Make compatible with requests 2.29.0 and urllib3 2.0

felixfontein:
Fixes at least part of #3113. (Recent requests versions declare that they need urllib3 < 2.0 since at least 2017, so the urllib3 problems themselves shouldn't really appear except when forcing invalid combinations of requests and urllib3. My hope is that the urllib3 2.0 compatibility is something that only needs changes on the side of requests, and not here - but who knows...)

Since requests 2.29.0, requests uses urllib3's native chunking ability (see https://github.com/psf/requests/pull/6226). By using urllib3's `HTTPConnection` class (which extends httplib.HTTPConnection`) we make sure that our HTTP transports support this as well.

I was able to reproduce the problem by calling `APIClient.put_archive()`; the change in this PR fixes the problem for me. (I'm working on a modified vendored version of Docker SDK for Python, so YMMV.)

### DIFF
--- a/docker/transport/npipeconn.py
+++ b/docker/transport/npipeconn.py
@@ -5,17 +5,13 @@ from docker.transport.basehttpadapter import BaseHTTPAdapter
 from .. import constants
 from .npipesocket import NpipeSocket
 
-try:
-    import requests.packages.urllib3 as urllib3
-    import requests.packages.urllib3.connection as urllib3_connection
-except ImportError:
-    import urllib3
-    import urllib3.connection as urllib3_connection
+import urllib3
+import urllib3.connection
 
 RecentlyUsedContainer = urllib3._collections.RecentlyUsedContainer
 
 
-class NpipeHTTPConnection(urllib3_connection.HTTPConnection):
+class NpipeHTTPConnection(urllib3.connection.HTTPConnection):
     def __init__(self, npipe_path, timeout=60):
         super().__init__(
             'localhost', timeout=timeout
--- a/docker/transport/npipeconn.py
+++ b/docker/transport/npipeconn.py
@@ -5,17 +5,17 @@ from docker.transport.basehttpadapter import BaseHTTPAdapter
 from .. import constants
 from .npipesocket import NpipeSocket
 
-import http.client as httplib
-
 try:
     import requests.packages.urllib3 as urllib3
+    import requests.packages.urllib3.connection as urllib3_connection
 except ImportError:
     import urllib3
+    import urllib3.connection as urllib3_connection
 
 RecentlyUsedContainer = urllib3._collections.RecentlyUsedContainer
 
 
-class NpipeHTTPConnection(httplib.HTTPConnection):
+class NpipeHTTPConnection(urllib3_connection.HTTPConnection):
     def __init__(self, npipe_path, timeout=60):
         super().__init__(
             'localhost', timeout=timeout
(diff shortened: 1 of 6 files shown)


felixfontein:
Failing CI (https://github.com/docker/docker-py/actions/runs/4861956711/jobs/8667665911?pr=3116) seems to be unrelated since it didn't fail before the force push (which only adjusted the commit message). Restarting CI.

felixfontein:
As a reference, the first test run that didn't fail: https://github.com/docker/docker-py/actions/runs/4861929164/jobs/8667601939

felixfontein:
According to #3117 and https://github.com/docker/docker-py/issues/3113#issuecomment-1531570788 this also fixes compatibility with urllib3 2.0, once requests officially supports that.

docker/transport/npipeconn.py
```diff
@@ -5,17 +5,17 @@
...
 from .npipesocket import NpipeSocket
 
-import http.client as httplib
-
 try:
     import requests.packages.urllib3 as urllib3
+    import requests.packages.urllib3.connection as urllib3_connection
 except ImportError:
     import urllib3
+    import urllib3.connection as urllib3_connection
```
pquentin, line 13: I just realized that this try/except Import Error is not needed anymore as docker-py requires requests >= 2.26.0 and 2.16.0 unvendored urllib3 in 2017. Just import what you need directly from urllib3. (This applies to other transports below.)

felixfontein: Good point! I've simplified the imports in 115dbce2cf54457af256001e6d9376377758f58c. This also affects some unrelated pieces of code, I'm not sure whether that's OK for the maintainers, but let's see.

docker/transport/unixconn.py
```diff
@@ -5,16 +5,14 @@
...
-    import urllib3
+import urllib3
+import urllib3.connection
 
 
 RecentlyUsedContainer = urllib3._collections.RecentlyUsedContainer
 
 
-class UnixHTTPConnection(httplib.HTTPConnection):
+class UnixHTTPConnection(urllib3.connection.HTTPConnection):
```
pquentin, line 15: Thanks! I can't comment on unedited lines, but wanted to mention that putheader() and response_class() can both be removed.

felixfontein: Thanks! I've removed these functions in a48dd4c0cf0c34e3d15c2676d23349842af0ad53.

pquentin (approved):

felixfontein:
@milas can you take a look at this PR? Thanks.

milas (approved):
Thanks for the PR!

I verified several scenarios under the integration tests (there are calls to `put_archive` and I see the failures currently without this PR):

* requests 2.30.0 + urllib3 v2.x
* requests 2.28.x + urllib3 v1.x
* requests 2.30.0 + urllib3 v1.x

Looks good across the three. I'll get this merged and a new release tagged momentarily.

josephkishan:
@pquentin  @milas @felixfontein we facing the same issue like  "TypeError: request() got an unexpected keyword argument 'chunked'" We are using docker as 5.0.3 and requests as 2.30.0 and urllib3 as 2.0.2. Is these versions fine or do we need to pin the version until the issue resolves? If we have to pin then please suggest to us the versions we need to resolve this issue

pquentin:
@josephkishan You can keep requests 2.30.0 and urllib3 2.0.2, but you should upgrade docker to 6.1.0 or 6.1.1.

PR #3117, opened 2023-05-02
Fix urllib3 2.0 compatibility

pquentin:
By inheriting from `urllib3.connection.HTTPConnection` (that inherits from `httplib.HTTPConnection` itself), we can adapt to the internal changes in urllib3 2.0 that added a `request()` method that is incompatible with `httplib.HTTPConnection.request`.

See https://github.com/docker/docker-py/issues/3113#issuecomment-1531570788 for details.

### DIFF
--- a/docker/transport/unixconn.py
+++ b/docker/transport/unixconn.py
@@ -1,6 +1,5 @@
 import requests.adapters
 import socket
-import http.client as httplib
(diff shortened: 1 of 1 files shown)


pquentin:
Closing in favor of #3116

PR #3121, opened 2023-05-07
Support all `Network.connect` parameters in `client.containers.run` and `client.containers.create`

Skazza94:
In PR #3083, I introduced the support for the `network_driver_opt` parameter in `client.containers` `run` and `create`.

This PR will further extend this support, by adding all the parameters of `Network.connect` to the aforementioned methods.

I changed the methods signature a bit, removing the `network_driver_opts` parameter and adding a dict called `network_config` that supports all the parameters.

I would kindly request if it possible to release a newer version of dockerpy with this feature merged soon, since I need it to continue developing some [Kathará](https://github.com/KatharaFramework/Kathara) features. Thanks.

### DIFF
--- a/docker/models/containers.py
+++ b/docker/models/containers.py
@@ -8,7 +8,7 @@ from ..errors import (
     ContainerError, DockerException, ImageNotFound,
     NotFound, create_unexpected_kwargs_error
 )
-from ..types import HostConfig
+from ..types import EndpointConfig, HostConfig, NetworkingConfig
 from ..utils import version_gte
 from .images import Image
 from .resource import Collection, Model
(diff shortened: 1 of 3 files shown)


tcaiazzi:
Hi all, 
any news on this PR? 

We are waiting for it for a long time :innocent:

docker/models/containers.py
```diff
@@ -680,10 +680,34 @@ def run(self, image, command=None, stdout=True, stderr=False,
                   This mode is incompatible with ``ports``.
 
                 Incompatible with ``network``.
-            network_driver_opt (dict): A dictionary of options to provide
-                to the network driver. Defaults to ``None``. Used in
-                conjuction with ``network``. Incompatible
-                with ``network_mode``.
+            network_config (dict): A dictionary containing options that are
```
milas, line 683: There is an `EndpointConfig` type: https://github.com/docker/docker-py/blob/c38656dc7894363f32317affecc3e4279e1163f8/docker/types/networks.py#L5-L8

Newer versions of the Moby API (will) allow defining multiple networks on create, so let's keep this the same here, e.g. `Dict[str, EndpointConfig]`, so that we don't have to update this again / make another breaking change.

(I'm ok with this breaking change to replace `network_driver_opt` since it was just introduced in v6, so it is hopefully not that entrenched in codebases yet.)

milas (changes requested):
Thanks for this PR and your prior contributions. Having this sit around for so long is unfortunate; if you're still interested I would love to help get this merged and we can do a v7 release along with a few other changes.

Overall it looks good, I'd just like to re-use the `EndpointConfig` type and make sure we are able to support multiple networks as part of container create.

Skazza94:
Hi @milas,
no worries! Thanks for the reply.

I've made the requested changes. The parameter has now been updated to be of type `Dict[str, EndpointConfig]`. Additionally, I have also refactored the sanity checks implemented within `_create_container_args` to align with the new parameter structure.

Furthermore, I have renamed the new parameter to `networking_config`, since it is much more semantically coherent with the rest of the API, as the class itself is named `NetworkingConfig`. 

Let me know if you require any other change.

Thanks,
Mariano.

milas:
Thanks again for making the changes and apologies with how long this took to get merged. I'm going to get Python 3.12 compatibility fixed and release a 7.0.0 ASAP

PR #3122, opened 2023-05-08
chore: move configurations to `pyproject.toml`

SauravMaheshkar:
This PR aims to minimize the number of configuration files for the project. Instead of a separate `.coveragerc` and `pytest.ini`, we now would have all the configurations within the `pyproject.toml` file leading to a overall minimal code structure.


akx:
This could be followed up by the removal of `setup.py` in favor of a PEP 517 based build with `setuptools` or `hatch` :)

SauravMaheshkar:
> This could be followed up by the removal of `setup.py` in favor of a PEP 517 based build with `setuptools` or `hatch` :)

Yes, that would be the next step. It allows follows nicely with #3126

SauravMaheshkar:
Closing this for now, will reopen if there is interest from maintainers.

PR #3123, opened 2023-05-10
feat: move websocket-client to extra dependency

akx:
Since `websocket-client` is _only_ required when using websockets for attaching to containers (https://github.com/docker/docker-py/pull/62), I don't think it needs to be a required dependency.

This also bumps the minimum required version of `websocket-client` to that pinned by #3022.

### DIFF
--- a/docker/api/client.py
+++ b/docker/api/client.py
@@ -5,7 +5,6 @@ from functools import partial
 
 import requests
 import requests.exceptions
-import websocket
 
 from .. import auth
 from ..constants import (DEFAULT_NUM_POOLS, DEFAULT_NUM_POOLS_SSH,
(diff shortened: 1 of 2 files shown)


akx:
@milas Hi, thanks for getting back on this! We have a situation where we'd like to avoid downloading extra dependencies in general – and from an attack surface/SBOM point of view, the less code there is, the better.

`attach_websocket` doesn't seem to be used [all that much in the wild](https://github.com/search?q=attach_websocket+language%3APython&type=code) but sure, I agree this needs a version bump of some sort. #3126 would probably be nice to get in too at that point :)

milas (approved):
Merging this - I'm going to do a major version bump to v7 for Python 3.12 compatibility, as I had to remove some old functionality in the process that constitutes a breaking change, so it's a good time to accept this as well.

Thanks for the PR!

akx:
@milas Great, thanks! Should v7 also drop Python 3.7 support? It's EOL date was 4 months ago.

PR #3125, opened 2023-05-10
Avoid socket timeouts when executing commands

loicleyendecker:
Only listen to read events when polling a socket in order to avoid incorrectly trying to read from a socket that is not actually ready.

### DIFF
--- a/docker/utils/socket.py
+++ b/docker/utils/socket.py
@@ -37,7 +37,7 @@ def read(socket, n=4096):
             select.select([socket], [], [])
         else:
             poll = select.poll()
-            poll.register(socket)
+            poll.register(socket, select.POLLIN | select.POLLPRI)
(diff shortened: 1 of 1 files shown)


docker/utils/socket.py
```diff
@@ -37,8 +37,7 @@ def read(socket, n=4096):
             select.select([socket], [], [])
         else:
             poll = select.poll()
-            poll.register(socket)
-            poll.poll()
```
domidimi, line 41: and this line is not needed anymore?

loicleyendecker: Oops, yes, it is. I just deleted it by mistake when I deleted some logs for my testing... I just re-added it now.

milas (approved):
Ahh good catch, thanks for the PR!

PR #3126, opened 2023-05-11
Switch linting to ruff, use f-strings, fix lint errors and possible bugs

akx:
This PR is a bit on the big side, but it should be pretty straightforward to review commit by commit. A net line reduction 
too!

It:

* replaces `.format` and `%` string formatting with the more succinct and performant f-strings wherever possible (this was half powered by `flynt`, half manual work)
* replaces the `flake8` linter with the much faster, more capable [ruff](https://github.com/charliermarsh/ruff) linter
* ... and then starts to fix up various issues and possible bugs found by `ruff`. See each commit's message for details.


milas:
FYI the integration test failure was because the test was incorrect - it was missing `assert` so the statements were no-ops. I fixed the status code in the merge after comparing to real engine behavior.

(Also apologies for noise to anyone watching this repo, it appears my Git pushes might have caused a bunch of notification emails 😬)

PR #3130, opened 2023-05-16
fix user_guides/multiplex.rst

Longin-Yu:
The format is wrong and the code in last case (``stream=False`` and ``demux=True``) is the same as the ``stream=True`` and ``demux=True`` one.


milas:
Thanks for the correction - can you amend the commit with a sign-off and I'll get this merged?

See https://github.com/docker/docker-py/pull/3130/checks?check_run_id=13508868314 for instructions.

Longin-Yu:
Thanks for your reminder. I have committed with a sign-off now.

milas (approved):

PR #3143, opened 2023-06-18
fix: Make source a keyword argument for Mounts

### LINKED ISSUE #3139: Cannot create tmpfs with docker.types.Mount
I'm using docker==6.1.2 with server version  23.0.1

Creating a  tmpfs mount doesn't work unfortunately. If not providing the field `source` I get a client error 
`TypeError: Mount.__init__() missing 1 required positional argument: 'source'`. 
If providing the field `source` I get a server error
`Bad Request ("invalid mount config for type "tmpfs": field Source must not be specified")`

Please try out with the following script
```
import docker.types
import docker

client = docker.from_env()

client.containers.run(
    image='ubuntu',
    name='test',
    detach=True,
    mounts=[docker.types.Mount(
        type='tmpfs',
        target='/dev/shm',
        tmpfs_size="1M",
        #source="/tmp"
    )],
    entrypoint="/bin/sh -c",
    command=["while true; do echo hello world; sleep 1; done"],
)
```

I had to revert to using a tmpfs argument as workaround
```
tmpfs=["/dev/shm:size=1G"]
```

Jack-Dane:
* There is no source for tmpfs mounts.
* Raise an exception if a source hasn't been parsed for types other than tmpfs.

https://docs.docker.com/storage/tmpfs/

There is the possibility to make the keyword argument an empty string as it should raise an error anyway when needed for certain types of mounts, but I think this error is easier to understand + is similar to the implementation already. 

fixes #3139


milas (approved):
Thanks!

milas:
Okay, I looked at this more after the unit test failures. Empty `source` is already supported, but you need to explicitly pass `None`. Similarly, it's actually valid for other cases as well, such as anonymous volumes.

As a result, I'm closing this PR, but let me know if you have questions or there's a case that still isn't supported. I gave an example on #3139 which might be useful.

PR #3149, opened 2023-07-03
Update Jenkinsfile

owenwu811:
added comments


owenwu811:
added comments and documentation

milas:
Closing - this file is not actually used anymore and should be deleted, no need to document it.

Issue #3151, opened 2023-07-04
404 API Error is not handled properly


mndeveci:
Hi there,

We recently upgraded dockerpy version to 6.2, and we found out an issue when we try to make a call to a non-existent API.

Before this upgrade, we used to get `docker.errors.NotFound` error with following details;
```
Traceback (most recent call last):
  File "test.py", line 9, in <module>
    main()
  File "test.py", line 5, in main
    registry_data = docker_client.images.get_registry_data("docker/hello-world")
  File "/private/tmp/docker-test/.venv420/lib/python3.8/site-packages/docker/models/images.py", line 337, in get_registry_data
    attrs=self.client.api.inspect_distribution(name, auth_config),
  File "/private/tmp/docker-test/.venv420/lib/python3.8/site-packages/docker/utils/decorators.py", line 34, in wrapper
    return f(self, *args, **kwargs)
  File "/private/tmp/docker-test/.venv420/lib/python3.8/site-packages/docker/utils/decorators.py", line 19, in wrapped
    return f(self, resource_id, *args, **kwargs)
  File "/private/tmp/docker-test/.venv420/lib/python3.8/site-packages/docker/api/image.py", line 281, in inspect_distribution
    return self._result(
  File "/private/tmp/docker-test/.venv420/lib/python3.8/site-packages/docker/api/client.py", line 267, in _result
    self._raise_for_status(response)
  File "/private/tmp/docker-test/.venv420/lib/python3.8/site-packages/docker/api/client.py", line 263, in _raise_for_status
    raise create_api_error_from_http_exception(e)
  File "/private/tmp/docker-test/.venv420/lib/python3.8/site-packages/docker/errors.py", line 31, in create_api_error_from_http_exception
    raise cls(e, response=response, explanation=explanation)
docker.errors.NotFound: 404 Client Error: Not Found ("b'Not Found'")
```

After upgrade, we are now getting this error `Error: a bytes-like object is required, not 'str'`, which is raised during error handling/generation inside dockerpy library;
```
Traceback (most recent call last):
  File "test.py", line 9, in <module>
    main()
  File "test.py", line 5, in main
    registry_data = docker_client.images.get_registry_data("docker/hello-world")
   File "/private/tmp/docker-test/.venv610/lib/python3.8/site-packages/docker/models/images.py", line 356, in get_registry_data
    attrs=self.client.api.inspect_distribution(name, auth_config),
[...]

PR #3156, opened 2023-07-07
fix: use response.text to get string rather than bytes

mndeveci:
Fixes; https://github.com/docker/docker-py/issues/3151

`response.content` returns bytes which eventually fails at line 34 when searching for image not found error message.

This change fixes it by using `response.text` which returns string rather than bytes.


milas (approved):
LGTM, thanks for the fix!

PR #3159, opened 2023-07-12
feat: add pause option to commit api

VincentLeeMax:
When using `container.commit()` function, it default to pause the container before committing. In some case that could't pause the container, we need to use `container.commit(pause=False)`。


milas (approved):
LGTM!

From https://docs.docker.com/engine/api/v1.43/#tag/Image/operation/ImageCommit:
> `pause` [boolean]
> Default: `true`
> Whether to pause the container before committing

PR #3167, opened 2023-08-14
README: fix link for CI status badge

thaJeztah:
The default branch was renamed from master to main, but the badge was still linking to the status for the master branch.

Remove the branch-name so that the badge always refers to the "default" branch


milas (approved):
Thanks :)

PR #3169, opened 2023-08-18
integration: Fix bad subnet declaration

akerouanton:
Some network integration tests are creating networks with subnet `2001:389::1/64`. This is an invalid subnet as the host fragment is non-zero (ie. it should be `2001:389::/64`).

PR moby/moby#45759 is adding strict validation of network configuration. Docker API will now return an error whenever a bad subnet is passed.


akerouanton:
@milas PTAL :hugs:

milas (approved):

PR #3170, opened 2023-08-19
integration: Remove test_create_check_duplicate

akerouanton:
moby/moby#46251 marks CheckDuplicate as deprecated. Any NetworkCreate request with a conflicting network name will now return an error.

### DIFF
--- a/tests/integration/api_network_test.py
+++ b/tests/integration/api_network_test.py
@@ -327,8 +327,6 @@ class TestNetworks(BaseAPIIntegrationTest):
         net_name, net_id = self.create_network()
         with pytest.raises(docker.errors.APIError):
             self.client.create_network(net_name, check_duplicate=True)
-        net_id = self.client.create_network(net_name, check_duplicate=False)
-        self.tmp_networks.append(net_id['Id'])
 
     @requires_api_version('1.22')
     def test_connect_with_links(self):


akerouanton:
@milas PTAL :hugs:

tests/integration/api_network_test.py
```diff
@@ -323,13 +323,6 @@ def test_create_with_links(self):
 
         self.execute(container, ['nslookup', 'bar'])
 
-    def test_create_check_duplicate(self):
-        net_name, net_id = self.create_network()
-        with pytest.raises(docker.errors.APIError):
-            self.client.create_network(net_name, check_duplicate=True)
-        net_id = self.client.create_network(net_name, check_duplicate=False)
```
thaJeztah, line 330: alternatively, you could only remove the "=False" case; the other check in this test ("=True") should still be valid (but the "=True" will be a no-op in future)

thaJeztah:
```
AttachContainerTest.test_attach_no_stream ___________________
tests/integration/api_container_test.py:1299: in test_attach_no_stream
    assert output == 'hello\n'.encode(encoding='ascii')
E   AssertionError: assert b'' == b'hello\n'
E     Use -v to get more diff
___________________ TestNetworks.test_create_check_duplicate ___________________
tests/integration/api_network_test.py:330: in test_create_check_duplicate
    self.tmp_networks.append(net_id['Id'])
E   TypeError: string indices must be integers
```

Ha! Wondering if the last failure is an implicit type conversion; if the network ID could be parsed as a hexadecimal🤔

Let me try restarting CI

thaJeztah:
Although it failed in 2 check, so perhaps that line needs to be removed (and was only needed for the second test?)

thaJeztah (approved):
LGTM

@milas PTAL

milas (approved):

PR #3176, opened 2023-09-08
[Bug report] Add support for Python 3.12

hugovk:
The [second and final Python 3.12 release candidate](https://discuss.python.org/t/python-3-12-0rc2-final-release-candidate-released/33105/5?u=hugovk) is out! :rocket:

> ## Call to action
> 
> We strongly encourage maintainers of third-party Python projects to prepare their projects for 3.12 compatibilities during this phase, and where necessary publish Python 3.12 wheels on PyPI to be ready for the final release of 3.12.0.

See also https://dev.to/hugovk/help-test-python-312-beta-1508/

The full Python 3.12.0 release is in under a month: 2023-10-02.

# Bug report

Consider this a PR a bug report, the Python 3.12 build fails:

```pytb
________________ ERROR collecting tests/unit/ssladapter_test.py ________________
ImportError while importing test module '/home/runner/work/docker-py/docker-py/tests/unit/ssladapter_test.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
/opt/hostedtoolcache/Python/3.12.0-rc.2/x64/lib/python3.12/importlib/__init__.py:90: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
tests/unit/ssladapter_test.py:2: in <module>
    from ssl import match_hostname, CertificateError
E   ImportError: cannot import name 'match_hostname' from 'ssl' (/opt/hostedtoolcache/Python/3.12.0-rc.2/x64/lib/python3.12/ssl.py)
```

This is because `ssl.match_hostname` was removed in Python 3.12 after being deprecated in 3.7:

* https://docs.python.org/3.12/whatsnew/3.12.html#removed
* https://docs.python.org/3.11/library/ssl.html#ssl.match_hostname

# Replacements

Here's how a couple of other projects dealt with it:

* https://github.com/urllib3/urllib3/pull/2448 vendored `ssl.match_hostname`
* https://github.com/aiortc/aioquic/pull/389 replaced `ssl.match_hostname` with PyCA's `service-identity`

Feel free to use this PR as a starting point for fixing it as you see best, or close this as needed.

Thanks!

PR #3185, opened 2023-11-20
breaking: Python 3.12 compatibility & remove custom SSL adapter

milas:
`match_hostname` is gone in Python 3.12 and has been unused by
Python since 3.7.

The custom SSL adapter allows passing a specific SSL version; this
was first introduced a looong time ago to handle some SSL issues
at the time.

See #3176.


glours (approved):
LGTM

PR #3191, opened 2023-12-01
Fix #3153 - Validate tag before build using OCI regex

### LINKED ISSUE #3153: build images with tag with https prefix never ends
## Reproduces

```
docker_client = docker.from_env()
docker_client .images.build(path="{my-path}", tag="https://dummy:latest", dockerfile="{dockerfile-path})
```

Using the debugger, we can see it loop on the _post request here:

https://github.com/docker/docker-py/blob/84414e343e526cf93f285284dd2c2c40f703e4a9/docker/api/build.py#L264-L271

LombardiDaniel:
Fix #3153, but guarantees that it woud raise the appropriate error.

The regex expression is:
`^[a-z0-9]+((\.|_|__|-+)[a-z0-9]+)*(\/[a-z0-9]+((\.|_|__|-+)[a-z0-9]+)*)*(:[a-zA-Z0-9_][a-zA-Z0-9._-]{0,127})?$`

I made it by combining `^[a-z0-9]+((\.|_|__|-+)[a-z0-9]+)*(\/[a-z0-9]+((\.|_|__|-+)[a-z0-9]+)*)*` and the `(:[a-zA-Z0-9_][a-zA-Z0-9._-]{0,127})?` group. Both expressions are from the [opencontainers/distribution-spec](https://github.com/opencontainers/distribution-spec) repo, as cited in the [docker docs](https://docs.docker.com/engine/reference/commandline/tag/).

I also wrote a test for testing a build function call with erroneous tag value.

### DIFF
--- a/docker/api/build.py
+++ b/docker/api/build.py
@@ -129,6 +129,9 @@ class BuildApiMixin:
             raise errors.DockerException(
                 'Can not use custom encoding if gzip is enabled'
             )
+        if tag is not None:
+            if not utils.match_tag(tag):
+                raise errors.DockerException(f"invalid tag '{tag}': invalid reference format")
 
         for key in container_limits.keys():
             if key not in constants.CONTAINER_LIMITS_KEYS:
--- a/docker/api/build.py
+++ b/docker/api/build.py
@@ -131,14 +131,14 @@ class BuildApiMixin:
             )
         if tag is not None:
             if not utils.match_tag(tag):
-                raise errors.DockerException(f"invalid tag '{tag}': invalid reference format")
-
+                raise errors.DockerException(
+                    f"invalid tag '{tag}': invalid reference format"
+            )
         for key in container_limits.keys():
             if key not in constants.CONTAINER_LIMITS_KEYS:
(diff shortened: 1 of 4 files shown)


milas (commented):
Thanks for the PR! I just kicked off CI and there's some failures - once those are taken care of, I'll get this merged:
* Lint errors - `make ruff` (will run inside Docker container)
* Unit tests (need to update example tag) - `make unit-test-py3` (will run inside Docker container)

Let me know if you have any trouble

LombardiDaniel:
@milas Done! the error was due to me changing the exception type and not updating the test!

milas (approved):
👍🏻 

Out of curiosity, was it you or the linter that reformatted the test file? (Not a big deal, just surprised)

LombardiDaniel:
No no, i had made a change to better represent the error, before it was `TypeError` but changed to `errors.DockerException` to better represent the scenario. But had forgotten to change the test accordingly

consideRatio:
I opened https://github.com/docker/docker-py/issues/3240 related to this PR

PR #3192, opened 2023-12-03
feat: accept filters and keep_storage in prune_builds

emranbm:
Make the `api.prune_builds` method accept `filters` and `keep-storage` args to pass to the docker API. The API details can be found [here](https://docs.docker.com/engine/api/v1.43/#tag/Image/operation/BuildPrune).

### DIFF
--- a/docker/api/build.py
+++ b/docker/api/build.py
@@ -276,9 +276,19 @@ class BuildApiMixin:
         return self._stream_helper(response, decode=decode)
 
     @utils.minimum_version('1.31')
-    def prune_builds(self):
+    def prune_builds(self, filters=None, keep_storage=None):
         """
         Delete the builder cache
+
+        Args:
+            filters (dict): Filters to process on the prune list.
+                Available filters:
+                - dangling (bool):  When set to true (or 1), prune only
+                unused and untagged images.
+                - until (str): Can be Unix timestamps, date formatted
+                timestamps, or Go duration strings (e.g. 10m, 1h30m) computed
+                relative to the daemon's local time.
+            keep_storage (int): Amount of disk space in bytes to keep for cache.
 
         Returns:
             (dict): A dictionary containing information about the operation's
@@ -290,7 +300,12 @@ class BuildApiMixin:
                 If the server returns an error.
         """
         url = self._url("/build/prune")
-        return self._result(self._post(url), True)
+        params = {}
+        if filters is not None:
+            params['filters'] = utils.convert_filters(filters)
+        if keep_storage is not None:
+            params['keep-storage'] = keep_storage
+        return self._result(self._post(url, params=params), True)
 
     def _set_auth_headers(self, headers):
         log.debug('Looking for auth config')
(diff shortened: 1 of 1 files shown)


docker/api/build.py
```diff
@@ -276,10 +276,20 @@ def build(self, path=None, tag=None, quiet=False, fileobj=None,
         return self._stream_helper(response, decode=decode)
 
     @utils.minimum_version('1.31')
-    def prune_builds(self):
+    def prune_builds(self, filters=None, keep_storage=None):
```
milas, line 279: There's also `all`, which is a boolean and was added at the same time as `filters` and `keep-storage` - let's add that as well

docker/api/build.py
```diff
@@ -290,7 +300,12 @@ def prune_builds(self):
                 If the server returns an error.
         """
         url = self._url("/build/prune")
-        return self._result(self._post(url), True)
+        params = {}
+        if filters is not None:
+            params['filters'] = utils.convert_filters(filters)
+        if keep_storage is not None:
+            params['keep-storage'] = keep_storage
+        return self._result(self._post(url, params=params), True)
```
milas, line 308: These args were added in API v1.39 - typically, we add a check for them, e.g.
```python
if utils.version_lt(self._version, '1.39'):
    raise errors.InvalidVersion(
        '`keep-storage` arg only available for API version > 1.38'
    )
```

(if you look for `utils.version_lt` or `utils.version_gte` you'll see a bunch of examples - they're not 100% consistent so do whatever makes sense here, e.g. I'm fine if there's a single check if ANY of `filters` / `keep_storage` / `any` are not `None` since they were all added together in the same API version)

milas (changes requested):
Thank you for the PR! I requested a couple of small changes:
* Add the `all` parameter which was added to API v1.39 at the same time: https://docs.docker.com/engine/api/v1.39/#tag/Image/operation/BuildPrune
* Add a programmatic API version check before using the new parameters

docker/api/build.py
emranbm: Done!

docker/api/build.py
emranbm: Done!

docker/api/build.py
```diff
@@ -300,11 +304,19 @@ def prune_builds(self, filters=None, keep_storage=None):
...
+                '`filters`, `keep_storage`, and `all` args are only available '
+                'for API version > 1.38'
+            )
         params = {}
         if filters is not None:
             params['filters'] = utils.convert_filters(filters)
         if keep_storage is not None:
             params['keep-storage'] = keep_storage
+        if keep_storage is not None:
+            params['keep-storage'] = keep_storage
```
milas, line 319:
```suggestion
        if all is not None:
            params['all'] = all
```
😂

milas (changes requested):
Thanks for the speedy turn-around

LGTM, just needs a small fix from copy-paste 🙃

emranbm:
> Thanks for the speedy turn-around
> 
> LGTM, just needs a small fix from copy-paste 🙃

Oh sorry :cry:

milas (approved):

milas:
Merged! Thanks for the PR! ❤️

Issue #3195, opened 2023-12-12
7.0: Docker Tag Parsing Issue with Registry Ports


bbezak:
#### Description:
We're facing a parsing issue in docker.py version 7.0, where registry server addresses with port numbers are incorrectly processed as part of the Docker tag, leading to an `invalid tag format` error.

#### Error:
it' happening in openstack kolla upstream CI:
```docker.errors.DockerException: invalid tag 'primary:4000/lokolla/base:change_903165': invalid reference format```

#### Possible issue:
The current regex introduced in commit: https://github.com/docker/docker-py/commit/a9b5494fd0574e520c735e1d0c4a303528d48063
in `docker.py` does not accommodate port numbers in registry addresses:
```python
_TAG = re.compile(
    r"^[a-z0-9]+((\.|_|__|-+)[a-z0-9]+)*(\/[a-z0-9]+((\.|_|__|-+)[a-z0-9]+)*)*" \
    + "(:[a-zA-Z0-9_][a-zA-Z0-9._-]{0,127})?$"
)

**Comment by 6ax** [comment]
We faced the same problem. Apparently the regular expression should be like this: 
`r"^[a-z0-9]+((\.|_|__|-+)[a-z0-9]+)*(:[0-9]{2,5})?" \
     + "(\/[a-z0-9]+((\.|_|__|-+)[a-z0-9]+)*)*$"`

**Comment by artificial-intelligence** [comment]
well, looking at the [OCI-Spec](https://github.com/opencontainers/distribution-spec/blob/main/spec.md#workflow-categories) I think the new code is more correct/matches what OCI mandates.

a `name` must match ```[a-z0-9]+((\.|_|__|-+)[a-z0-9]+)*(\/[a-z0-9]+((\.|_|__|-+)[a-z0-9]+)*)*```
and a `tag` must match ```[a-zA-Z0-9_][a-zA-Z0-9._-]{0,127}```

**Comment by artificial-intelligence** [comment]
So I did dig deeper, and afaik this is a bug in the OCI spec:

https://github.com/opencontainers/distribution-spec/commit/a73835700327bd1c037e33d0834c46ff98ac1286 introduced the new regex:

```
`[a-z0-9]+([._-][a-z0-9]+)*(/[a-z0-9]+(\.|_|__|-+)[a-z0-9]+)*)*`
```

but this is wrong imho, because the commit message clearly mentions:

```
This change makes the separators consistent with the implementation in the Docker distribution
[...]

PR #3196, opened 2023-12-12
fix: tag regex should allow ports

artificial-intelligence:
Closes: https://github.com/docker/docker-py/issues/3195
Related: https://github.com/opencontainers/distribution-spec/pull/498

notice: there is a false positive warning from ruff during `make test`:

```
docker/utils/build.py:14:33: W605 [*] Invalid escape sequence: `\.`
```

but it's a valid escape sequence, because it's a regex, had no idea yet how to fix this without introducing more complexity.

### DIFF
--- a/docker/utils/build.py
+++ b/docker/utils/build.py
@@ -10,8 +10,9 @@ from ..constants import IS_WINDOWS_PLATFORM
 
 _SEP = re.compile('/|\\\\') if IS_WINDOWS_PLATFORM else re.compile('/')
 _TAG = re.compile(
-    r"^[a-z0-9]+((\.|_|__|-+)[a-z0-9]+)*(\/[a-z0-9]+((\.|_|__|-+)[a-z0-9]+)*)*" \
-        + "(:[a-zA-Z0-9_][a-zA-Z0-9._-]{0,127})?$"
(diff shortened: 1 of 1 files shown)


artificial-intelligence:
further notice:

I did fix the ruff warning about `\/` because that escape sequence is not needed.

LombardiDaniel:
Make sure to add a simple test like:

```py
def test_build_with_ports_on_tag(self):
        self.client.build(".", tag="registry.example.com:443/my-img:latest")
```

on `tests/build/api_build_and_test.py`. And why not use raw string literals? i.e.

```py
r"my-regex"
```

instead of 

```py
"my-regex"
```

milas (approved):
Thanks for the bugfix!

I addressed the lint issue (it was valid but very subtle btw: needed to use `r""` strings each time for the concatenation 😵) and added some test cases from the `distribution/reference` codebase.

(There's some xfails on the IPv6 cases which can be addressed separately.)

SecT0uch:
When can we expect a release including this fix ?

artificial-intelligence:
Thanks @milas for taking over (I was first on PTO, then this issue was drowned in my inbox due to other issues)!

a new release would be highly appreciated for our downstream usage!

PR #3199, opened 2023-12-15
Bump default API version to 1.43 (Moby 24.0)

vvoland:
Update API version to the latest maintained release.


docker/constants.py
```diff
@@ -1,7 +1,7 @@
 import sys
 from .version import __version__
 
-DEFAULT_DOCKER_API_VERSION = '1.41'
+DEFAULT_DOCKER_API_VERSION = '1.43'
 MINIMUM_DOCKER_API_VERSION = '1.21'
```
thaJeztah, line 5: We should also start looking at raising this minimum, but I don't know if there's code specific in docker-py that handles differences between versions @milas ?

We're starting to deprecate old API versions, and the initial step is to disable (by default) API versions < 1.24. API 1.24 is the last version of the API before introduction of API version negotiation, so that's the version the CLI falls back to if it's unable to negotiate.

- https://github.com/moby/moby/pull/46887

thaJeztah (approved):
LGTM

glours (approved):
LGTM

PR #3200, opened 2023-12-17
fix: keyerror when creating new config(#3110)

### LINKED ISSUE #3110: Can't create config object
Much like https://github.com/docker/docker-py/issues/2025 the config model is failing to create a new object due to 'name' KeyError

```
  File "docker\models\configs.py", line 10, in __repr__
    return f"<{self.__class__.__name__}: '{self.name}'>"
  File "docker\models\configs.py", line 14, in name
    return self.attrs['Spec']['Name']
```

This https://github.com/docker/docker-py/pull/2793 appears to be the fix that was implemented and should likely be implements for configs as well (if not other models that might have this issue)

Khushiyant:
fix #3110


Khushiyant:
@milas PTAL

milas (approved):
LGTM - behavior of the API is weird as well as the `docker-py` handling, but this is identical to how it's handled for secrets, so good enough

PR #3201, opened 2023-12-17
fix: none output type in exec_run(#3172)

### LINKED ISSUE #3172: `exec_run` with `demux=True`: output is `None` instead of empty `bytes`
When calling `exec_run` with `demux=True`, the documentation states that a pair of `bytes` is returned.

However, if no standard output (or no standard error) is produced, then the corresponding tuple entry is `None` instead of an empty `bytes` (`b''`).

Minimum working example:

```python
import docker
client = docker.from_env()
container = client.containers.run("alpine:latest", detach=True, tty=True)
container.exec_run(("sh", "-c", "echo out >&1"), demux=True)  # ExecResult(exit_code=0, output=(b'out\n', None))
container.exec_run(("sh", "-c", "echo err >&2"), demux=True)  # ExecResult(exit_code=0, output=(None, b'err\n'))
```

The expected behaviour is that both tuple entries are always `bytes`. Alternatively, the possibility of returning `None` should be documented clearly!

Khushiyant:
fix #3172

This is added with refernce to [docker-py docs](https://docker-py.readthedocs.io/en/stable/containers.html?highlight=exec_run#docker.models.containers.Container.exec_run)

Changes:
``` diff
+ exec_output = tuple(b'' if x is None else x for x in exec_output)
```


Khushiyant:
@milas PTAL

PR #2931, opened 2022-01-03
Use `packaging` instead of `distutils` for `Version`

FrancescoCasalegno:
Fixes #2928.

## Context

`setuptools` has deprecated the usage of `distutils` `Version` classes since version `60.3.0`. See https://github.com/pypa/setuptools/commit/1701579e0827317d8888c2254a17b5786b6b5246 for details.

As a consequence, a `warning` is raised any time we try, e.g., to do `from docker import from_env`. Notice that this `warning` can be fatal if the user has set a warning filter to `error` mode, which is not too uncommon when running unit tests in CI.

## Proposed solution

1. Replace occurrences of `distutils.Version` (and its subclasses) with `packaging.Version`.
2. Add `packaging` to the `install_requires`.

PR #3205, opened 2023-12-21
Drop `packaging` dependency

akx:
<del>It doesn't seem to be required.</del>

Implement [the same version comparison algorithm](https://github.com/moby/moby/blob/e7001c14558af124d7959c1d1cfe3a16b507c521/api/types/versions/compare.go#L8-L40) Moby uses instead.

### DIFF
--- a/docker/utils/utils.py
+++ b/docker/utils/utils.py
@@ -6,7 +6,8 @@ import os.path
 import shlex
 import string
 from datetime import datetime, timezone
-from packaging.version import Version
+from functools import lru_cache
+from itertools import zip_longest
 
 from .. import errors
 from ..constants import DEFAULT_HTTP_HOST
@@ -43,6 +44,7 @@ def decode_json_header(header):
     return json.loads(data)
 
 
+@lru_cache(maxsize=None)
 def compare_version(v1, v2):
     """Compare docker versions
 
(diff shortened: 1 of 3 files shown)


milas (changes requested):
`packaging.Version` is used in `docker/utils/utils.py`:
https://github.com/docker/docker-py/blob/f467fd9df9929cbf72455aee7bdc6b114eb83b82/docker/utils/utils.py#L58-L59

That was changed from `distutils.Version`:
* #2931

I'm happy to get rid of the dependency, especially given the small surface area. Do you have a suggested alternative?

akx:
> `packaging.Version` is used in `docker/utils/utils.py`:

Ah, sorry, yeah, good catch.

> Do you have a suggested alternative?

How are Docker versions specified? `packaging.Version` is, after all, meant to parse [Python version specifiers](https://packaging.python.org/en/latest/specifications/version-specifiers/), not versions in general...

If they're specified to be just `A.B[.C]` where each part is an integer, then it's trivial to split, cast to integers in a tuple and compare them :)

milas:
> If they're specified to be just `A.B[.C]` where each part is an integer, then it's trivial to split, cast to integers in a tuple and compare them :)

Yep, that's effectively what Moby does: https://github.com/moby/moby/blob/e7001c14558af124d7959c1d1cfe3a16b507c521/api/types/versions/compare.go#L8-L40

(Technically there can be an arbitrary number of version sections; in practice, I hope we never make an API version `1.2.3.4.5.6.7.8.9.0` 😛)

akx:
@milas Great :) Rebased, implemented the same algorithm here, added tests.

milas (approved):
LGTM and 😲 with that turn-around time 🎉

PR #3207, opened 2023-12-21
Modernize build to use `hatchling` and `hatch-vcs`

akx:
This PR switches the packaging for `docker-py` from the old and busted `setup.cfg` + `setup.py` stuff to [PEP517](https://peps.python.org/pep-0517/) [Hatch](https://hatch.pypa.io)'s `hatchling` and the `hatch-vcs` plugin to provide what `setuptools_scm` did.

Aside from superficial changes in metadata structure (and file times being static for more reproducible builds), the resulting wheels are the same (and of course the `_version` file is different because there's two extra commits here...):

comp1

There's also an added GitHub Actions workflow that builds the wheel even when not publishing, so you can look at the results for any given PR.

> [!NOTE]  
> This PR includes #3205 – it drops `packaging` out of the requirements.


milas (approved):
A HUGE thanks from me, this is long overdue and looks great.

PR #3212, opened 2024-01-03
Enable Ruff I (import sort), autofix

akx:
I didn't have a good time sorting imports by hand in the tests for #3205, so maybe a machine should do it for us 🤖.

The first commit is entirely mechanical, but it turns out it uncovered circular import issues, so the second commit takes care of them.


akx:
Ping @milas? :)

milas (approved):
LGTM (merging to main manually to handle merge conflict because `setup.py` is gone now)

PR #3222, opened 2024-01-26
Automatically add many type hints, with ruff

adamtheturtle:
If this library is to ship type hints, it is important that those type hints are valid. A type checker (e.g. mypy or pyright) can check that those type hints are valid.

This PR is one small step towards that goal.

I ran:

```
ruff --extend-select=ANN201,ANN202,ANN204,ANN205,ANN206 --fix --unsafe-fixes docker
```

and I did not make any manual changes.

I chose not to also type other directories (such as `tests`) as that is a decision I'd first like to hear from core maintainers on.

### DIFF
--- a/docker/api/build.py
+++ b/docker/api/build.py
@@ -322,7 +322,7 @@ class BuildApiMixin:
             params['all'] = all
         return self._result(self._post(url, params=params), True)
 
-    def _set_auth_headers(self, headers):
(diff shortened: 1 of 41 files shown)


akx:
https://github.com/JelleZijlstra/autotyping, https://github.com/orsinium-labs/infer-types, https://github.com/instagram/MonkeyType could be used to add more types (but they'll require more hand-holding and manual review).

PR #3233, opened 2024-03-13
Fix tests that look at 'Aliases'

robmry:
With Docker Engine 24.x, 25.x (and probably older versions), `inspect` output for `NetworkSettings.Networks.<network>.Aliases` includes the container's short-id.

(The short-id will be removed from 'Aliases' in API v1.45, in moby 26.0. So, this change will need to be reverted when docker-py is updated to that version.)

### DIFF
--- a/tests/integration/models_containers_test.py
+++ b/tests/integration/models_containers_test.py
@@ -110,12 +110,12 @@ class ContainerCollectionTest(BaseIntegrationTest):
         client.networks.create(net_name)
         self.tmp_networks.append(net_name)
 
-        test_aliases = ['hello']
+        test_alias = 'hello'
         test_driver_opt = {'key1': 'a'}
 
         networking_config = {
             net_name: client.api.create_endpoint_config(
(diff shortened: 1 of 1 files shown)


robmry:
It turns out that, while the docker <= 25.0 does include the short-id in inspect output for a running container, the result of docker-py's `run()` returns the state of the container after the 'create' but before the 'start' ... and that didn't include the short-id until docker 25.0.

Which explains why the tests had been passing ... and highlights an inadvertent change in docker 25.0.

(In 25.0, docker uses the API version to decide whether to include the short-id, but it doesn't precisely match the 24.0 behaviour. In 24.0, inspect after the container's started includes short-id, even once the container's been stopped again.)

We're not planning to fix that, as it's a bit subtle and inconsistent - but it means we need to include this change to the tests in the change that bumps docker-py to 25.0. (It'll still need to be reverted when we bump to 26.0 / API version 1.45, because the API is deliberately changed to exclude it in that version.)

PR #3242, opened 2024-04-04
Add equivalent of bind-recursive option to the Mount type class

sairon:
With the recursive mount behavior change in Docker 25, it is not possible to make recursive mounts writable with the current API. Add the `recursive` option which is equivalent of bind-recursive in CLI. This also allows for setting the mount to be non-recursive (added earlier in API v1.41).

Ugly workarounds [like this](https://github.com/home-assistant/supervisor/pull/4997/files#diff-0ea86e6e87f3633d75b11c96e6244085c1fdf008e50f63b11db2f7e2f2ec4e6bR77) are needed to adjust the behavior now, it will be nice to have this in the standard Python API.

Related:
* https://github.com/moby/moby/issues/47158
* https://github.com/moby/moby/pull/45278


docker/types/services.py
```diff
@@ -235,6 +235,9 @@ class Mount(dict):
         ``default```, ``consistent``, ``cached``, ``delegated``.
         propagation (string): A propagation mode with the value ``[r]private``,
           ``[r]shared``, or ``[r]slave``. Only valid for the ``bind`` type.
+        recursive (string): Bind mount recursive mode,  one of ``enabled``,
```
agners, line 238:
```suggestion
        recursive (string): Bind mount recursive mode, one of ``enabled``,
```

docker/types/services.py
```diff
@@ -267,6 +270,21 @@ def __init__(self, target, source, type='volume', read_only=False,
...
+                    bind_options['NonRecursive'] = True
+                elif recursive == "writable":
+                    bind_options['ReadOnlyNonRecursive'] = True
+                elif recursive == "readonly":
+                    bind_options['ReadOnlyForceRecursive'] = True
+                else:
+                    raise errors.InvalidArgument(
+                        'Invalid recursive bind option, must be one of '
+                        '"enabled", "disabled", "writable", or "readonly".'
+                    )
```
agners, line 287: Looks good.

The CLI also does some validation, not sure if we should do those too? :thinking: 
https://github.com/docker/cli/blob/v25.0.5/opts/mount.go#L183-L196

sairon: I'm not sure - some other options are also validated loosely here or not at all, so I'm sticking with that, only added this `InvalidArgument` error for unknown options since we need to enumerate the possible values to map them to mount config options anyway.

thaJeztah: (disclaimer: I'm a maintainer for the docker CLI, but I'm not very good at Python, or very familiar with existing code in this repository for that matter).

In general (in docker/cli), we try to keep client-side validation "light" where possible, so that we can defer to the API as source of truth (also accounting for situations where only the daemon can fully validate, as well as preventing the CLI from disallowing options that _may_ be supported at some point in future).

That said, for some cases we do some validation on the client side, either where we are sure (or "very confident") certain validation would never change, or for situations where performing the API request (when invalid) would be much more "heavyweight" than a local check, and where a client-side validation can provide a better experience to the user (e.g. producing a more specific error message).

All of the above out of the way, it would be good to verify the behavior when using one of those invalid combinations; the API _should_ validate those, but it'd be good to check if it does, and if so, if the error-message is useful enough to the user to resolve the problem.

sairon: Cool, thanks for the insights, @thaJeztah. For me the only concern is this:
> we are sure (or "very confident") certain validation would never change

I _think_ it's the case for the referred checks done in CLI for the bind options, and the check in Python will have minimal overhead, so I'm fine with adding it here. Let's see what others have to say.

thaJeztah: Just to emphasize; If the API already responds with a reasonable error, then I'm (personally) perfectly fine with keeping it light.

PR #3263, opened 2024-05-23
Fix env var name in release pipeline

krissetto:
Fixed the env var name passed into the `Generate Package` job of the release pipeline to match what is used in the Makefile


thaJeztah (approved):
LGTM

laurazard (approved):

krissetto:
For the sake of whoever might come here in the future, this PR was unnecessary because we actually needed the env var name to match what `hatch` was expecting, since the packaging job of the release pipeline doesn't use our Makefile.  

Proper fix was done [here](https://github.com/docker/docker-py/pull/3264)

PR #3264, opened 2024-05-23
Change env var name in release pipeline to match hatch expectations

krissetto:
Since the Packaging job of the release pipeline doesn't go through our Makefile, we need the env var defined here to have the exact name `hatch` is expecting


laurazard (approved):

thaJeztah (approved):
LGTM

PR #3266, opened 2024-05-27
Updated outdated deployment steps

krissetto:
Updated CONTRIBUTING docs so the deployment steps reflect the current GitHub action based workflow

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -120,14 +120,21 @@ $ make docs
 $ open _build/index.html
 ```
 
-## Release Checklist
+## Deployment
 
-Before a new release, please go through the following checklist:
+To deploy a new release, please go through the following steps:
 
-* Bump version in docker/version.py
-* Add a release note in docs/change_log.md
-* Git tag the version
-* Upload to pypi
+* Make sure a release note for the changes included since the last published version is present in [docs/change-log.md](./docs/change-log.md);
+* Go to the [`Release`](https://github.com/docker/docker-py/actions/workflows/release.yml) GitHub action page and click on `Run workflow`. Here:
+  * Insert the version number you want to give to the new package, **without** the `v.` prefix (eg. `7.1.0`). This version should correspond with the one written in [docs/change-log.md](./docs/change-log.md);
+  * Make sure `Dry run` is checked;
+  * Click `Run workflow`;
+  * Click on the newly executed workflow, click on the `publish` job, and under `Generate Package` check to see if the package has been built successfully and with the correct version number.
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -120,14 +120,21 @@ $ make docs
...
-* Git tag the version
-* Upload to pypi
+## Deployment
+
+To deploy a new release, please go through the following steps:
+
+* Make sure a release note for the changes included since the last published version is present in [docs/change-log.md](./docs/change-log.md);
+* Go to the [`Release`](https://github.com/docker/docker-py/actions/workflows/release.yml) GitHub action page and click on `Run workflow`. Here:
+  * Insert the version number you want to give to the new package, **without** the `v.` prefix (eg. `7.1.0`). This version should correspond with the one written in [docs/change-log.md](./docs/change-log.md);
+  * Make sure `Dry run` is checked;
```
laurazard, line 130: Not sure we need to document the `dry-run` option, or ossify it as part of the actual release process, as it's just a debugging/testing tool. At most, I'd explain what it does i.e. "If `dry-run` is checked, then the package is build but a release is not created/uploaded to pypi"

krissetto: my logic here was "better safe than sorry" (considering past events), but i'm ok with changing this :)

krissetto: It removes "if-branching" from the steps and can be followed as-is in a sense

laurazard: to clarify my point, I just mean that running a dry-run first is not a necessary step of the release process, so I don't think we should document it as such, since it might introduce confusion "do I have to do a dry run before releasing?". Instead we could just explain what it does (and if we want, mention something like "it's advisable to test the release pipeline before release by checking the "dry-run" option).

krissetto: Yeah I get your point. I agree that technically speaking a dry run is not necessary, but I'd argue it **should** be part of the "release process" so that people double check the version number, and only those in-the-know (or confident enough) should skip it on their discretion.  

I'm saying this mostly because the result of the pipeline is yes a "draft release" on github, but at that point of the process the package has **already** been published to pypi and users might be downloading it before we realize and correct any potential issue.  

No super hard feelings here, lets see if others have also differing opinions on this

laurazard: I don't disagree, I just think that including "dry-run" in that section removes clarity about what it actually does. The same can be achieved by separately explaining what "dry-run" does, and strongly advising it's use.

@thaJeztah ?

PR #3145, opened 2023-06-20
test: remove APT_MIRROR from Dockerfile

thaJeztah:
The official Python images on Docker Hub switched to debian bookworm, which is now the current stable version of Debian.

However, the location of the apt repository config file changed, which causes the Dockerfile build to fail;

    Loaded image: emptyfs:latest
    Loaded image ID: sha256:0df1207206e5288f4a989a2f13d1f5b3c4e70467702c1d5d21dfc9f002b7bd43
    INFO: Building docker-sdk-python3:5.0.3...
    tests/Dockerfile:6
    --------------------
       5 |     ARG APT_MIRROR
       6 | >>> RUN sed -ri "s/(httpredir|deb).debian.org/${APT_MIRROR:-deb.debian.org}/g" /etc/apt/sources.list \
       7 | >>>     && sed -ri "s/(security).debian.org/${APT_MIRROR:-security.debian.org}/g" /etc/apt/sources.list
       8 |
    --------------------
    ERROR: failed to solve: process "/bin/sh -c sed -ri \"s/(httpredir|deb).debian.org/${APT_MIRROR:-deb.debian.org}/g\" /etc/apt/sources.list     && sed -ri \"s/(security).debian.org/${APT_MIRROR:-security.debian.org}/g\" /etc/apt/sources.list" did not complete successfully: exit code: 2

The APT_MIRROR build-arg was originally added when the Debian package repositories were known to be unreliable, but that hasn't been the case for quite a while, so let's remove this altogether.

PR #3267, opened 2024-05-28
Set a dummy-version if none set, and remove unused APT_MIRROR build-arg

thaJeztah:
- relates to https://github.com/docker/docker-py/pull/3145
- relates to https://github.com/moby/moby/pull/47092

### Set a dummy-version if none set

Make sure the Dockerfiles can be built even if no VERSION build-arg is passed.

### Makefile: remove unused APT_MIRROR build-arg

The APT_MIRROR build-arg was removed from the Dockerfile in commit
ee2310595d1362428f2826afac2e17077231473a, but wasn't removed from the
Makefile.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,7 +6,7 @@ FROM python:${PYTHON_VERSION}
 WORKDIR /src
 COPY . .
 
-ARG VERSION
+ARG VERSION=0.0.0.dev0
 RUN --mount=type=cache,target=/cache/pip \
     PIP_CACHE_DIR=/cache/pip \
     SETUPTOOLS_SCM_PRETEND_VERSION=${VERSION} \
--- a/Dockerfile-docs
+++ b/Dockerfile-docs
@@ -13,7 +13,7 @@ RUN addgroup --gid $gid sphinx \
 WORKDIR /src
 COPY . .
 
-ARG VERSION
+ARG VERSION=0.0.0.dev0
 RUN --mount=type=cache,target=/cache/pip \
     PIP_CACHE_DIR=/cache/pip \
     SETUPTOOLS_SCM_PRETEND_VERSION=${VERSION} \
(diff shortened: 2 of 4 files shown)


Dockerfile
```diff
@@ -6,7 +6,7 @@ FROM python:${PYTHON_VERSION}
 WORKDIR /src
 COPY . .
 
-ARG VERSION
+ARG VERSION=0.0.0-dev
```
thaJeztah, line 9: I chose `0.0.0-dev`, but not sure if anything uses the version, so if `99.9.9-dev` or `99.0.0-dev` ("always higher than what exists") would be better.

Happy to change!

laurazard: Okay either way, but maybe someone with more python experience than me can chime in.

laurazard (approved):

Dockerfile
thaJeztah: @neersighted any preference here?

krissetto: [PEP-440](https://peps.python.org/pep-0440/) defines the syntax for declaring a development package as the `.devN` suffix (where N is just a number), so I'd use that instead of `-dev`.

`pip` __should__ only install stable packages when dependencies are declared as `someDep >= 1.2.3`, but it will allow installation of `.devN` dependencies if by chance the user has a dependency declared as `someDep >= 4.5.6.dev2` for example.

Because of ☝️, I'd suggest staying with the lower v `0.0.0.dev` just to be on the safer side in case we accidentally end up publishing a package with that version to pypi.  

I'm not 100% sure this is the best approach though, so if others have more experience please chime in 😄

thaJeztah: If it expects a number after `.dev`, I can make it `0.0.0.dev0` I guess @neersighted SGTY?

neersighted: `0.0.0.dev0` is PEP 440 (though the current spec lives at https://packaging.python.org/en/latest/specifications/version-specifiers/#version-specifiers) compliant, so that works for me.

Notably all numeric components are allowed to be 0; the proposed version does pass Poetry's PEP 440 validation, which is notoriously a bit strict.

neersighted (changes requested):
Use a PEP 440-compliant version (e.g. the proposed `0.0.0.dev0` works).

PR #3270, opened 2024-06-13
Implement Subpath Support for Volumes in Docker-Py

Khushiyant:
This PR aims to implement subpath support for volumes in Docker-Py, allowing users to specify a subpath when creating a volume or a container. This feature is essential for scenarios where a container needs to access a specific directory within a volume.

Added a `subpath` parameter to the `create_volume` method in `docker/api/volume.py`.
Updated the `Volume` model in `docker/models/volumes.py` to include a subpath attribute.
Modified the `create_container` method in `docker/api/container.py` to include the `subpath` parameter when creating a mount.
Updated the `Mount` class in `docker/types/mounts.py` to include a `subpath` attribute.

> This implementation is based on the Docker API documentation and might require further testing and refinements.
The subpath parameter is optional, and users can still create volumes and containers without specifying a subpath.


rayrapetyan:
Hello, can this PR be merged ASAP?

Khushiyant:
@rayrapetyan Even, I'm also waiting for the review

rayrapetyan:
@thaJeztah @krissetto, could you provide a review please, this seems like a very simple and straightforward change. Thanks!

rayrapetyan (approved):

rayrapetyan:
@Khushiyant, could you re-trigger the build as seems one of the integration tests has failed before.

Khushiyant:
> @Khushiyant, could you re-trigger the build as seems one of the integration tests has failed before.

Yeah, I triggered the build and now, all tests have passed

docker/api/volume.py
```diff
@@ -45,6 +45,9 @@ def create_volume(self, name=None, driver=None, driver_opts=None,
             driver (str): Name of the driver used to create the volume
             driver_opts (dict): Driver options as a key-value dictionary
             labels (dict): Labels to set on the volume
+            subpath (str): Subpath within the volume to use as the volume's
```
krissetto, line 48: I'd change this to 
```
subpath (str): Path inside a volume to mount instead of the volume root.
```
To better match documentation we have elsewhere (eg https://github.com/docker/docs/blob/23f3fbd0639a24d0e2280fd0edb7f72b3f0cb1ca/content/reference/compose-file/services.md?plain=1#L1865)

@dvdksn WDYT?

docker/types/services.py
```diff
@@ -242,14 +242,16 @@ class Mount(dict):
           for the ``volume`` type.
         driver_config (DriverConfig): Volume driver configuration. Only valid
           for the ``volume`` type.
+        subpath (string): The path within the volume where the container's
```
krissetto, line 245: I'd keep this the same as the suggestion above ☝️

tests/integration/api_container_test.py
```diff
@@ -620,6 +620,29 @@ def test_create_with_volume_mount(self):
         assert mount['Source'] == mount_data['Name']
         assert mount_data['RW'] is True
 
+    @requires_api_version('1.45')
+    def test_create_with_subpath_volume_mount(self):
+        mount = docker.types.Mount(
+            type="volume", source=helpers.random_name(),
+            target=self.mount_dest, read_only=True,
+            subpath='subdir'
```
krissetto, line 628: From a quick glance, it doesn't seem that the subpath is being checked in this test

docker/types/services.py
```diff
@@ -280,6 +282,8 @@ def __init__(self, target, source, type='volume', read_only=False,
                 volume_opts['Labels'] = labels
             if driver_config:
                 volume_opts['DriverConfig'] = driver_config
+            if subpath:
+                volume_opts['SubPath'] = subpath
```
krissetto, line 286: I think this should be written as `Subpath` here, right @vvoland?

krissetto (commented):
Thanks for the contribution!
I gave this a quick look for now, we'll try to get it in soon

docker/types/services.py
Khushiyant: Yeah, it is `Subpath` instead of `SubPath`: [moby#45687](https://github.com/moby/moby/pull/45687)

docker/api/volume.py
dvdksn: SGTM

tests/integration/api_container_test.py
Khushiyant: Made changes accordingly

docker/types/services.py
```diff
@@ -280,6 +282,8 @@ def __init__(self, target, source, type='volume', read_only=False,
                 volume_opts['Labels'] = labels
             if driver_config:
                 volume_opts['DriverConfig'] = driver_config
+            if subpath:
+                volume_opts['Subpath'] = subpath
```
Khushiyant, line 286: Change the `SubPath` to `Subpath`

tests/integration/api_container_test.py
```diff
@@ -620,6 +620,32 @@ def test_create_with_volume_mount(self):
...
+            lambda x: x['Destination'] == self.mount_dest,
+            inspect_data['Mounts']
+        ))
+
+        print(mount)
+        assert len(filtered) == 1
+        mount_data = filtered[0]
+        assert mount['Source'] == mount_data['Name']
+        assert mount_data['RW'] is False
+        assert mount["VolumeOptions"]["Subpath"] == "subdir"
```
krissetto, line 647: This is testing that the input mount options, as defined above on line 628, are the same. 

The test needs to check that the path mounted _inside the container_ corresponds to the subdir as defined in the mount options. The volume mounted must therefore contain the `subdir`, or else there should be an error.

Maybe you could change to check `mount_data` instead, if it contains the necessary info? I don't have much time now but I'd have to check how these volume tests are defined in a bit more detail later

Khushiyant: Oh, sure

Khushiyant:
I have cleaned up the commit history and remove unnecessary changes and correct tests

PR #3290, opened 2024-09-27
tests/exec: expect 127 exit code for missing executable

laurazard:
The Docker Engine has always returned `126` when starting an exec fails due to a missing binary, but this was due to a bug in the daemon causing the correct exit code to be overwritten in some cases – see: https://github.com/moby/moby/issues/45795

Change tests to expect correct exit code (`127`).

Note: Must take some care to what's the best way to fix this without breaking engine or vice-versa, due to the engine running tests against docker-py and docker-py running tests against the engine. It might be necessary to 1st. skip the test, then fix the engine, then fix/unskip the test.

### DIFF
--- a/tests/integration/models_containers_test.py
+++ b/tests/integration/models_containers_test.py
@@ -353,6 +353,15 @@ class ContainerTest(BaseIntegrationTest):
         assert exec_output[0] == 0
         assert exec_output[1] == b"hello\n"
 
+    def test_exec_run_error_code_from_exec(self):
+        client = docker.from_env(version=TEST_API_VERSION)
+        container = client.containers.run(
+            "alpine", "sh -c 'sleep 20'", detach=True
+        )
+        self.tmp_containers.append(container.id)
+        exec_output = container.exec_run("sh -c 'exit 42'")
+        assert exec_output[0] == 42
+
     def test_exec_run_failed(self):
         client = docker.from_env(version=TEST_API_VERSION)
         container = client.containers.run(
@@ -363,7 +372,7 @@ class ContainerTest(BaseIntegrationTest):
         # older versions of docker return `126` in the case that an exec cannot
         # be started due to a missing executable. We're fixing this for the
         # future, so accept both for now.
-        assert exec_output[0] == 127 or exec_output == 126
+        assert exec_output[0] == 127 or exec_output[0] == 126
 
     def test_kill(self):
         client = docker.from_env(version=TEST_API_VERSION)
--- a/tests/integration/models_containers_test.py
+++ b/tests/integration/models_containers_test.py
@@ -359,8 +359,11 @@ class ContainerTest(BaseIntegrationTest):
             "alpine", "sh -c 'sleep 60'", detach=True
         )
         self.tmp_containers.append(container.id)
-        exec_output = container.exec_run("docker ps")
-        assert exec_output[0] == 126
+        exec_output = container.exec_run("non-existent")
+        # older versions of docker return `126` in the case that an exec cannot
+        # be started due to a missing executable. We're fixing this for the
+        # future, so accept both for now.
+        assert exec_output[0] == 127 or exec_output == 126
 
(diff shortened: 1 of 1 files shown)


laurazard:
PR to fix the breaking docker-py test: https://github.com/docker/docker-py/pull/3290

I called it out there, but we'll need to TAL at the dependencies between the engine and docker-py. Docker-py runs integration tests against the engine, and the engine uses docker-py to run tests, so changing it in either place first is a bit annoying. We might need to skip the test in docker-py -> change the engine behavior -> unskip+fix the test in docker-py, and update the engine version it runs against.

tests/integration/models_containers_test.py
```diff
@@ -360,7 +360,7 @@ def test_exec_run_failed(self):
         )
         self.tmp_containers.append(container.id)
         exec_output = container.exec_run("docker ps")
-        assert exec_output[0] == 126
+        assert exec_output[0] == 127
```
thaJeztah, line 363: Perhaps check for "either 126 or 127" and a comment?

Given that this test was added as a general "include exit code", also wondering if something like this would work (produce a custom exit code, and check for that)?

```bash
docker exec foo sh -c 'exit 42'
echo $?
42
```

laurazard: I _this_ test is just for the exit code for a failed exec (as in, failed to start), not for general exit code. it fails that way because `docker` isn't a binary in the container. The exit code here isn't the exit code from the process executed in the container as a general "include exit code" test because that can only happen if the process in the container can be started.

I'll TAL and see if we're missing tests for that case though, and can add some.

tests/integration/models_containers_test.py
```diff
@@ -353,14 +353,26 @@ def test_exec_run_success(self):
...
             "alpine", "sh -c 'sleep 60'", detach=True
         )
         self.tmp_containers.append(container.id)
-        exec_output = container.exec_run("docker ps")
-        assert exec_output[0] == 126
+        exec_output = container.exec_run("non-existent")
+        # older versions of docker return `126` in the case that an exec cannot
+        # be started due to a missing executable. We're fixing this for the
+        # future, so accept both for now.
+        assert exec_output[0] == 127 || exec_output == 126
```
thaJeztah, line 375:
```
error: Failed to parse tests/integration/models_containers_test.py:375:39: Unexpected token '|'
tests/integration/models_containers_test.py:375:39: E999 SyntaxError: Unexpected token '|'
```

I guess something like `or` ?

```suggestion
        assert exec_output[0] == 127 or exec_output == 126
```

tests/integration/models_containers_test.py
```diff
@@ -353,14 +353,26 @@ def test_exec_run_success(self):
...
             "alpine", "sh -c 'sleep 60'", detach=True
         )
         self.tmp_containers.append(container.id)
-        exec_output = container.exec_run("docker ps")
-        assert exec_output[0] == 126
+        exec_output = container.exec_run("non-existent")
+        # older versions of docker return `126` in the case that an exec cannot
+        # be started due to a missing executable. We're fixing this for the
+        # future, so accept both for now.
+        assert exec_output[0] == 127 or exec_output == 126
```
thaJeztah, line 375: DOH! And I think this;

```
/usr/local/lib/python3.12/shlex.py:191: in read_token
    raise ValueError("No closing quotation")
E   ValueError: No closing quotation
______________________ ContainerTest.test_exec_run_failed ______________________
tests/integration/models_containers_test.py:375: in test_exec_run_failed
    assert exec_output[0] == 127 or exec_output == 126
```

Means it should be

```suggestion
        assert exec_output[0] == 127 or exec_output[0] == 126
```


One day I know how to write python as well! 😂

laurazard: I'm too tired, time to log off for the week 😅

laurazard:
It's passing now @thaJeztah 🎉

PR #3312, opened 2025-02-18
fix waiting for auto_remove container

huguesdk:
use the same logic as the cli (see [here](https://github.com/docker/cli/blob/master/cli/command/container/run.go#L199) and [here](https://github.com/docker/cli/blob/master/cli/command/container/utils.go#L29)) to wait for the exit of a container when running one: ensure that a container run with `auto_remove` set to `True` has been removed when the function returns.

this prevents a race condition when trying to run another container with the same name directly afterwards.

the race condition can be reproduced like this:

```python
import docker
client = docker.from_env()
while True:
    client.containers.run(
        "some_image",
        name="race_condition_test",
        command="sleep 1",
        auto_remove=True,
    )
```

after some time, this error happens:

```
  File "/usr/lib/python3.10/site-packages/docker/api/client.py", line 275, in _raise_for_status
    response.raise_for_status()
  File "/usr/lib/python3.10/site-packages/requests/models.py", line 1024, in raise_for_status
    raise HTTPError(http_error_msg, response=self)
requests.exceptions.HTTPError: 409 Client Error: Conflict for url: http+docker://localhost/v1.41/containers/create?name=race_condition_test

The above exception was the direct cause of the following exception:

  File "<stdin>", line 2, in <module>
  File "/usr/lib/python3.10/site-packages/docker/models/containers.py", line 876, in run
    container = self.create(image=image, command=command,
  File "/usr/lib/python3.10/site-packages/docker/models/containers.py", line 935, in create
    resp = self.client.api.create_container(**create_kwargs)
  File "/usr/lib/python3.10/site-packages/docker/api/container.py", line 440, in create_container
    return self.create_container_from_config(config, name, platform)
  File "/usr/lib/python3.10/site-packages/docker/api/container.py", line 457, in create_container_from_config
    return self._result(res, True)
  File "/usr/lib/python3.10/site-packages/docker/api/client.py", line 281, in _result
  File "/usr/lib/python3.10/site-packages/docker/api/client.py", line 277, in _raise_for_status
    raise create_api_error_from_http_exception(e) from e
  File "/usr/lib/python3.10/site-packages/docker/errors.py", line 39, in create_api_error_from_http_exception
    raise cls(e, response=response, explanation=explanation) from e
docker.errors.APIError: 409 Client Error for http+docker://localhost/v1.41/containers/create?name=race_condition_test: Conflict ("Conflict. The container name "/race_condition_test" is already in use by container "2f632c9f52b2f9286574972a97ee615e3714204bf84aecae399623b609948301". You have to remove (or rename) that container to be able to reuse that name.")
```

### DIFF
--- a/docker/models/containers.py
+++ b/docker/models/containers.py
@@ -894,7 +894,24 @@ class ContainerCollection(Collection):
                 stdout=stdout, stderr=stderr, stream=True, follow=True
             )
 
-        exit_status = container.wait()['StatusCode']
+        if kwargs.get('auto_remove'):
+            wait_condition = 'removed'
+        else:
+            # the wait condition should theoretically be 'next-exit' (as is
+            # used by the cli), but it may have exited already if its run time
+            # was very short, which would cause the wait to hang.
+            # 'not-running' works in both cases.
+            wait_condition = 'not-running'
+        try:
+            exit_status = container.wait(condition=wait_condition)['StatusCode']
+        except NotFound:
+            if wait_condition == 'removed':
+                # it has been already removed, which is why it was not found,
+                # so everything fine here. unfortunately, there is no way to
+                # have its real exit status, so assume success.
+                exit_status = 0
+            else:
+                raise
         if exit_status != 0:
             out = None
             if not kwargs.get('auto_remove'):


legalsylvain:
Hi. Why CI is not executed here ?

huguesdk:
@legalsylvain i think it is because someone should approve this change first. i’ve just pushed some new changes to improve the code even more.

legalsylvain:
hi docker-py contributors. (@thaJeztah, @milas)

Could you take a look on that PR ? Thanks a lot !

huguesdk:
@legalsylvain i adjusted the tests to make them pass.

@thaJeztah do you think i should add tests for the new code?

PR #3318, opened 2025-03-07
Add support for ClusterVolumeSpec in create_volume api

### LINKED ISSUE #3254: Support `ClusterVolumeSpec` in the `create_volume` api
Docker's [`VolumeCreate` API](https://docs.docker.com/engine/api/v1.45/#tag/Volume/operation/VolumeCreate) takes a `ClusterVolumeSpec` parameter that it'd be nice to be able to use. 

Most commonly, `ClusterVolumeSpec.AccessMode.{Scope,Sharing}` will need to be set and the Docker CLI also has arguments to set these. But probably the whole `ClusterVolumeSpec` should be exposed in Python. For example in my case, I'd like to be able to set `ClusterVolumeSpec.AccessMode.MountVolume.MountFlags` (and I turned to the Python API because I could find no way to do this with the CLI...).

Currently the only workaround I'm aware of is to use the API directly instead of using the otherwise nice Python API.

Khushiyant:
Closes #3254 

Added following changes:

- Added `ClusterVolumeSpec ` in `docker.types`
- Added `cluster_volume_spec` arg and its handling to `create_volume` api
- Added tests in `dockertypes_test.py` and `api_volume_test.py`

### DIFF
--- a/docker/api/volume.py
+++ b/docker/api/volume.py
@@ -31,12 +31,11 @@ class VolumeApiMixin:
 
         params = {
             'filters': utils.convert_filters(filters) if filters else None
-        }
+            }
         url = self._url('/volumes')
         return self._result(self._get(url, params=params), True)
 
-    def create_volume(self, name=None, driver=None, driver_opts=None,
-                      labels=None):
+    def create_volume(self, name=None, driver=None, driver_opts=None, labels=None, cluster_volume_spec=None):
         """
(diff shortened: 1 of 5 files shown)


Khushiyant:
@thaJeztah PTAL, The integration test is failing in a different pipeline that is unrelated to the applied changes.

PR #3322, opened 2025-03-21
Signal Parameter to Container Stop and Restart

### LINKED ISSUE #3304: `container.stop` and `container.restart` are missing the `signal` argument
Both `stop` and `restart` have supported customizing the signal that is sent to the container since api version `v1.42` but it seems that it has not been added here. It would be very helpful if it could be added.

I would have a crack at it but I am not very familiar with writing unit tests. I saw that it is implemented for `kill` so it might not be too hard but I'm not sure.

Khushiyant:
Closes #3304 

This PR aims to provide support for signal that is sent to the container in `stop` and `restart` according to `API v1.42`.

Adds following changes:

- `signal` argument to `restart` and `stop` in `docker/api/container.py`
- `signal` docstring in `docker/models/containers.py`
-  unit tests in `tests/unit/models_containers_test.py`

> Refer [API version here](https://docs.docker.com/reference/api/engine/version-history/#v142-api-changes)

### DIFF
--- a/docker/api/container.py
+++ b/docker/api/container.py
@@ -1074,7 +1074,7 @@ class ContainerApiMixin:
         self._raise_for_status(res)
 
     @utils.check_resource('container')
-    def restart(self, container, timeout=10):
+    def restart(self, container, timeout=10, signal=None):
         """
         Restart a container. Similar to the ``docker restart`` command.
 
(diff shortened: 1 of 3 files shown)


ic:
Implementing now, and stopped as you did it a year ago already. Hopefully merged soon!

I was wondering whether replacing plain integer signals for `signal.Signals(int(signal))` or `signal.getsignal(int(signal)).value` (with `ValueError` handling) was attractive to make sure the signal code is valid. Upside may be to get early failure, without querying the Docker API and handle a 400. At the same time I like the transparency of your implementation.

PR #3325, opened 2025-03-28
Add support for platform parameter in image push

### LINKED ISSUE #3324: Support for platform in client.images.api.push
API [version v1.48](https://docs.docker.com/reference/api/engine/version/v1.48/#tag/Image/operation/ImagePush) accepts platform argument as parameter to be passed to `/images/{name}/push` but it is not implemented in [image.py](https://github.com/docker/docker-py/blob/eeb9ea1937ba5e67087551bdabed858fab5e0ef3/docker/api/image.py#L436).

Below function should accept platform i.e `JSON-encoded OCI platform to select the platform-variant to push`:

```python
def push(self, repository, tag=None, stream=False, auth_config=None,
             decode=False):
```

Khushiyant:
Closes #3324 

Adds the following:

- add `platform` parameter handling in `images.api.push`
- add tests in `tests/unit/api_image_test.py::ImageTest::test_push_image_with_platform`

> `platform` parameter was introduced in [v1.46](https://docs.docker.com/engine/release-notes/27/#api-2)


hoangelos:
Any chance someone can look at this?  This causing a bug for me.  What happens without this is tags can’t be pushed to registries.  Only the sha.  Say I copy a docker image from one registry to another.   You want to copy a specific platform, then it won’t push the tag without specifying the platform.

docker/api/image.py
```diff
@@ -448,6 +448,7 @@ def push(self, repository, tag=None, stream=False, auth_config=None,
                 ``username`` and ``password`` keys to be valid.
             decode (bool): Decode the JSON data from the server into dicts.
                 Only applies with ``stream=True``
+            platform (str): JSON-encoded OCI platform to select the platform-variant to push. If not provided, all available variants will attempt to be pushed.
```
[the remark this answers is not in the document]
hoangelos: Whatever the case shouldn’t it match the pull?   There shouldn’t be a string in one and class in another.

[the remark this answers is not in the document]
hoangelos: @vvoland can you take a look at this?

vvoland (changes requested):
CI is red

hoangelos:
@Khushiyant any progress on this?

docker/types/image.py
```diff
@@ -0,0 +1,35 @@
+from .base import DictType
+
+
+class Platform(DictType):
+    def __init__(self, **kwargs):
+        architecture = kwargs.get('architecture')
+        os = kwargs.get('os')
+
+        if architecture is None and os is None:
+            raise ValueError("At least one of 'architecture' or 'os' must be provided")
```
hoangelos: This message doesn't match the logic. you are requiring both and not just one.  I'd either correct the logic or correct the statement.

tests/unit/api_image_test.py
```diff
@@ -271,6 +272,28 @@ def test_push_image_with_auth(self):
             timeout=DEFAULT_TIMEOUT_SECONDS
         )
 
+    @requires_api_version('1.46')
+    def test_push_image_with_platform(self):
```
hoangelos, line 276: Not to be picky, but your function can take either Platform OR dict, but you are just testing one.

docker/types/image.py
Khushiyant: Yeah, this was a leftover from the last changes. I'll fix this.

tests/unit/api_image_test.py
Khushiyant: No issue, good point. I'll see towards that.

hoangelos (approved):

hoangelos:
@vvoland  it looks he did what you asked.

PR #3330, opened 2025-04-18
feat(volumes): add missing labels attribute

Ret2Me:
Hi,
I added "labels" attribute to volume object in the same way as it was done in container model.


Ret2Me:
Hi, it would be very helpful in my project. Would you like to marge my changes? @thaJeztah @krissetto

PR #3344, opened 2025-06-27
Add support for building with buildkit

### LINKED ISSUE #2230: Support BuildKit
Currently the Python SDK performs builds using the older build infrastructure. It would be good to support [BuildKit](https://github.com/moby/buildkit) so that users of the SDK can benefit from its various advantages.

felipecrs:
Requires passing `version='2'` like:

```python
stream = self.client.build(
    fileobj=dockerfile, tag='myimage', version='2'
)
```

Refs https://github.com/moby/moby/blob/3ff85c73436f1c4f8d9764a0d72e41a03b4116f5/api/swagger.yaml#L9343-L9352.

Closes #2230

### DIFF
--- a/docker/api/build.py
+++ b/docker/api/build.py
@@ -103,7 +103,7 @@ class BuildApiMixin:
                 Default: `None`.
             version (str): Version of the builder backend to use.
                 - `1` is the first generation classic (deprecated) builder in the Docker daemon (default)
-                - `2` is [BuildKit](https://github.com/moby/buildkit)
+                - `2` is [BuildKit](https://github.com/moby/buildkit) over the REST API endpoint provided by the Docker daemon. This mode doesn't use the BuildKit gRPC API, and lacks support for advanced features such as secret mounts.
                 Default: `None`.
             use_config_proxy (bool): If ``True``, and if the docker client
                 configuration file (``~/.docker/config.json`` by default)


felipecrs:
@thaJeztah do you think you can take a look at this one? I know the PR backlog is full, but in my opinion this PR is a no-brainer and is of huge help (https://github.com/docker/docker-py/issues/2230 has 200+ upvotes).

docker/api/build.py
```diff
@@ -253,6 +257,13 @@ def build(self, path=None, tag=None, quiet=False, fileobj=None,
                 )
             params['isolation'] = isolation
 
+        if version is not None:
+            if utils.version_lt(self._version, '1.38'):
+                raise errors.InvalidVersion(
+                    'version was only introduced in API version 1.38'
+                )
+            params['version'] = version
```
AkihiroSuda, line 265: Bumping up the version number doesn't magically implement the BuildKit API.
You have to dial `/grpc` to call the BuildKit gRPC API.
https://github.com/docker/buildx/blob/v0.25.0/driver/docker/driver.go
https://github.com/moby/buildkit/blob/v0.23.1/api/services/control/control.proto

In addition to implementing the gRPC client, you also have to implement several "attachable" servers (auth, secret, ssh) via the reverse-gRPC connection.
https://github.com/moby/buildkit/blob/v0.23.1/cmd/buildctl/build.go#L191-L214


This is **quite** complicated than you might imagine; I suggest just shelling out `docker buildx` and call it a day.

tests/integration/api_build_test.py
```diff
@@ -201,6 +201,27 @@ def test_build_isolation(self):
         for _chunk in stream:
             pass
 
+    @requires_api_version('1.38')
+    def test_build_with_buildkit(self):
+        script = io.BytesIO('\n'.join([
+            'FROM scratch',
+            'COPY <<EOF greeting.txt',
+            'hello world',
+            'EOF'
```
AkihiroSuda, line 210: The test should use BuildKit-specific features such as `RUN --mount=type=secret`

docker/api/build.py
felipecrs: Thanks a lot for reviewing it.

> Bumping up the version number doesn't magically implement the BuildKit API.

Tests prove you are wrong. No?

> In addition to implementing the gRPC client, you also have to implement several "attachable" servers (auth, secret, ssh) via the reverse-gRPC connection.

That's not this PR intention. Exposing additional APIs can be done in follow-up PRs. This one is sufficient for my use case which is building modern Dockerfiles that do not require on auth, secret, or ssh.

tests/integration/api_build_test.py
felipecrs: It is using Buildkit-specific features: `COPY` with here documents. Try setting version=1 and you'll watch the test fail.

docker/api/build.py
AkihiroSuda: Only basic BuildKit features can be enabled by just bumping up the version:
https://github.com/moby/moby/blob/v28.3.0/api/server/backend/build/backend.go#L55-L73

The scope of the available features should be documented

felipecrs: Do you mean I should clarify that `squash` and `forcerm` will error when `version=2`?

Otherwise, I expect all options currently supported by `client.build()` to be supported.

tests/integration/api_build_test.py
AkihiroSuda: Sorry, I overlooked the BuildKit-specific heredoc. Could you add a code comment to clarify the intent?

docker/api/build.py
```diff
@@ -101,6 +101,10 @@ def build(self, path=None, tag=None, quiet=False, fileobj=None,
             platform (str): Platform in the format ``os[/arch[/variant]]``
             isolation (str): Isolation technology used during build.
                 Default: `None`.
+            version (str): Version of the builder backend to use.
+                - `1` is the first generation classic (deprecated) builder in the Docker daemon (default)
+                - `2` is [BuildKit](https://github.com/moby/buildkit)
```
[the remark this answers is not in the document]
felipecrs: Thanks a lot, very handy suggestion.

docker/api/build.py
felipecrs: Ok, this is worse than I expected. 

I got lucky with the tests because I only tried with `FROM scratch`.

Even a basic Dockerfile like below does not work:

```python
    @requires_api_version('1.38')
    def test_build_buildkit_alpine(self):
        script = io.BytesIO('\n'.join([
            'FROM alpine',
        ]).encode('ascii'))

        self.tmp_imgs.append('buildkitalpine')

        stream = self.client.build(
            fileobj=script, tag='buildkitalpine',
            version='2'
        )

        for _chunk in stream:
            pass

        assert self.client.inspect_image('buildkitalpine')
```

But it's because of this bug:

- https://github.com/moby/moby/issues/48112

But still, it means not even basic use cases can use this.

@AkihiroSuda do you have any idea? Otherwise, I think it may be the end of this PR (and a dream). :(

AkihiroSuda (approved):

docker/api/build.py
felipecrs: I haven't been able to build anything so far, but I will update when I have any news.

felipecrs: FYI I found some gRPC references from an unfinished buildkit implementation in this library from @shin-:

- https://github.com/docker/docker-py/tree/buildkit

PR #3353, opened 2025-08-19
Fix n² consume socket output

### LINKED ISSUE #3349: Performance of `consume_socket_output` is n² where it could be just `n`
I noticed that [the function](https://github.com/docker/docker-py/blob/6e6a273573fe77f00776b30de0685162a102e43f/docker/utils/socket.py#L140-L174) is repeatedly concatenating `bytes`:

```py
def consume_socket_output(frames, demux=False):
    ...
    # If the streams are demultiplexed, the generator yields tuples
    # (stdout, stderr)
    out = [None, None]
    for frame in frames:
        # It is guaranteed that for each frame, one and only one stream
        # is not None.
        assert frame != (None, None)
        if frame[0] is not None:
            if out[0] is None:
                out[0] = frame[0]
            else:
                out[0] += frame[0]
        else:
            if out[1] is None:
                out[1] = frame[1]
            else:
                out[1] += frame[1]
    return tuple(out)
```

This makes it extremely slow for long inputs. I noticed since I used dockerpy to get about 1GB of process output, which took hours and blocked 100% CPU for the entire time.

The fix for it should be pretty simple, just collect all the frames and use `b''.join(frames)` instead:

```py
# If the streams are demultiplexed, the generator yields tuples
# (stdout, stderr)
stdout = []
stderr = []
for stdout_frame, stderr_frame in frames:
    # It is guaranteed that for each frame, one and only one stream
    # is not None.
    if stdout_frame:
        stdout.append(stdout_frame)
    else:
        stderr.append(stderr_frame)
stdout = b''.join(stdout) if len(stdout) > 0 else None
stderr = b''.join(stderr) if len(stderr) > 0 else None
return stdout, stderr
```

GeorchW:
Fixes #3349 - see there for details


GeorchW:
@kiview MR is here

PR #3374, opened 2025-12-01
Improve tmpfs_mode docs

kasium:
I had recently a great time to find out, why the tmpfs_mode=777 does not work until I've found out that in needs to be in octal. Therefore I created this PR for clarity


kasium:
Does anyone have time to have a look?

PR #3384, opened 2026-02-13
Fix sub-second precision loss in datetime_to_timestamp

### LINKED ISSUE #3342: Flooring `since` and `until` datetime arguments down to second resolution in `container.logs`.
I was surprised to figure out that when you pass `since` and `until` arguments to `container.logs()` as `datetime`s they internally are floored down to the closest second with `utils.datetime_to_timestamp(...)`.

Millisecond part of datetime is discarded for these arguments.

That leads to unwanted extra logs in the output from the past before `since` and absent valid logs before `until`.

For example:

```
since = datetime(2025, 6, 1, hour=12, minute=0, second=0, microsecond=500_000)
logs = container.logs(since=since)
# logs would have EXTRA records in half a second from 12:00:00.000 to 12:00:00.500

until = datetime(2025, 6, 1, hour=12, minute=0, second=0, microsecond=500_000)
logs = container.logs(until=until)
# logs would NOT have records in half a second from 12:00:00.000 to 12:00:00.500
```

Is it by design and I need to manually convert `datetime` to `float` in such usecase?

veeceey:
`datetime_to_timestamp` was using `delta.seconds + delta.days * 24 * 3600` to convert a timedelta to a Unix timestamp. The problem is that `timedelta.seconds` only returns the whole seconds component and throws away microseconds, so any sub-second precision in the input datetime gets silently floored.

This means passing a datetime like `datetime(2025, 6, 1, 12, 0, 0, microsecond=500_000)` as `since` to `container.logs()` would behave as if you passed `12:00:00.000` instead of `12:00:00.500`, returning an extra half-second of unwanted logs.

The fix is straightforward — replace the manual arithmetic with `delta.total_seconds()`, which returns a float that preserves microsecond precision. The Docker Engine API already accepts fractional timestamps, and the calling code in `container.logs()` and `events()` already handles float values.

Added unit tests covering sub-second precision, whole-second datetimes, epoch, and naive datetimes.

Fixes #3342

### DIFF
--- a/docker/utils/utils.py
+++ b/docker/utils/utils.py
@@ -405,7 +405,7 @@ def convert_filters(filters):
 def datetime_to_timestamp(dt):
     """Convert a datetime to a Unix timestamp"""
     delta = dt.astimezone(timezone.utc) - datetime(1970, 1, 1, tzinfo=timezone.utc)
-    return delta.seconds + delta.days * 24 * 3600
+    return delta.total_seconds()
 
 
 def parse_bytes(s):
--- a/tests/unit/utils_test.py
+++ b/tests/unit/utils_test.py
@@ -1,4 +1,5 @@
 import base64
+from datetime import datetime, timezone
 import json
 import os
 import os.path
@@ -15,6 +16,7 @@ from docker.utils import (
     compare_version,
     convert_filters,
     convert_volume_binds,
+    datetime_to_timestamp,
     decode_json_header,
     format_environment,
     kwargs_from_env,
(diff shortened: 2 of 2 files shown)


veeceey:
Friendly ping - any chance someone could take a look at this when they get a chance? Happy to make any changes if needed.

veeceey:
hey, just a gentle nudge on this when you have a sec

veeceey:
hey, just checking in - anything else needed on this one?

PR #3390, opened 2026-03-25
Gracefully handle credential store errors in get_all_credentials

### LINKED ISSUE #3379: Building images fails on auth errors even when auth is not required
## The Problem

When running `DockerClient.images.build(...)`, if a StoreError is raised when running `AuthConfig._resolve_authconfig_credstore` ([link](https://github.com/docker/docker-py/blob/df3f8e2abc5a03de482e37214dddef9e0cee1bb1/docker/auth.py#L267)) when trying to create auth headers for the build request, the whole command fails.

For my specific example, I have gcloud auth artifacts on my local machine that require login once per day - if I have not logged in and try to build a publicly accessible image, that build will still fail since `docker-credentials-gcloud get` exits non-zero by prompting me to log in again, **even though that auth is not required**.

## Proposed Solution

Either:
- try/except each call to `_resolve_authconfig_credstore` in `get_all_credentials` ([link](https://github.com/docker/docker-py/blob/df3f8e2abc5a03de482e37214dddef9e0cee1bb1/docker/auth.py#L298) and [link](https://github.com/docker/docker-py/blob/df3f8e2abc5a03de482e37214dddef9e0cee1bb1/docker/auth.py#L291)) and just don't add them to `auth_data` if the call errors
- do not raise a DockerExecption from the `StoreError` in [_resolve_authconfig_credstore](https://github.com/docker/docker-py/blob/df3f8e2abc5a03de482e37214dddef9e0cee1bb1/docker/auth.py#L267) and treat it the same as a `CredentialsNotFound` error (I could see this breaking other things)

Krishnachaitanyakc:
## Summary

Fixes #3379

When `get_all_credentials()` iterates over credential stores (both the default `credsStore` and per-registry `credHelpers`), a `StoreError` from any single store propagates as a `DockerException` and aborts the entire operation. This causes image builds to fail even when the failing credentials are not required — for example, when expired gcloud auth tokens exist in the local Docker config but the image being built only uses publicly accessible base images.

This change wraps each call to `_resolve_authconfig_credstore` inside `get_all_credentials` with a try/except that:
- Catches `DockerException` (which wraps the underlying `StoreError`)
- Logs a warning with the store name, registry, and error details
- Skips the failing entry and continues collecting other valid credentials

All other valid credentials are still collected and returned as before.

## Changes

- **`docker/auth.py`**: Modified `AuthConfig.get_all_credentials()` to catch `errors.DockerException` per-entry in both the `credsStore` loop and the `credHelpers` loop, logging a warning and continuing instead of propagating the error.
- **`tests/unit/auth_test.py`**: Added two new tests:
  - `test_get_all_credentials_credstore_error_skipped` — verifies that a failing `credHelper` entry is skipped while valid entries from the default store are still returned.
  - `test_get_all_credentials_default_store_error_skipped` — verifies that a failing entry in the default `credsStore` is skipped while other valid entries from the same store are still returned.
  - Added `FailingStore` and `PartiallyFailingStore` test helper classes.

## Test plan

- [x] All existing unit tests in `tests/unit/auth_test.py` continue to pass (58/58)
- [x] Two new unit tests pass covering both `credsStore` and `credHelpers` failure scenarios
- [ ] Manual verification: configure a Docker `config.json` with a credential helper that fails (e.g., expired `gcloud` auth), then run `client.images.build()` for a public image and confirm it no longer raises

### DIFF
--- a/docker/auth.py
+++ b/docker/auth.py
@@ -288,17 +288,35 @@ class AuthConfig(dict):
             # Retrieve all credentials from the default store
             store = self._get_store_instance(self.creds_store)
             for k in store.list().keys():
-                auth_data[k] = self._resolve_authconfig_credstore(
-                    k, self.creds_store
-                )
-                auth_data[convert_to_hostname(k)] = auth_data[k]
+                try:
+                    cred = self._resolve_authconfig_credstore(
+                        k, self.creds_store
+                    )
+                except errors.DockerException as e:
+                    log.warning(
+                        'Failed to retrieve credentials from store '
+                        '"%s" for %s: %s — skipping this entry',
+                        self.creds_store, k, e,
+                    )
+                    continue
+                auth_data[k] = cred
+                auth_data[convert_to_hostname(k)] = cred
 
         # credHelpers entries take priority over all others
         for reg, store_name in self.cred_helpers.items():
-            auth_data[reg] = self._resolve_authconfig_credstore(
-                reg, store_name
-            )
-            auth_data[convert_to_hostname(reg)] = auth_data[reg]
+            try:
+                cred = self._resolve_authconfig_credstore(
+                    reg, store_name
+                )
+            except errors.DockerException as e:
+                log.warning(
+                    'Failed to retrieve credentials from store '
+                    '"%s" for %s: %s — skipping this entry',
+                    store_name, reg, e,
+                )
+                continue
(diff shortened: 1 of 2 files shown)


Krishnachaitanyakc:
@akerouanton can you please review these changes?

Issue #3146, opened 2023-06-22
Docker CLI contexts not used or supported


brettdh:
Testing with Docker Desktop on a mac, I have this docker context state:

```
➜ docker context list
NAME                TYPE                DESCRIPTION                               DOCKER ENDPOINT                                       KUBERNETES ENDPOINT   ORCHESTRATOR
colima              moby                colima                                    unix:///Users/brhiggins/.colima/default/docker.sock
default             moby                Current DOCKER_HOST based configuration   unix:///var/run/docker.sock
desktop-linux *     moby                                                          unix:///Users/brhiggins/.docker/run/docker.sock
```

The `desktop-linux` context is active, and `/var/run/docker.sock` does not exist. `docker version` correctly uses the active context and returns the server version info. However, `docker.from_env()` throws an error:

```
../../../.pyenv/versions/3.11.4/envs/myproj/lib/python3.11/site-packages/docker/client.py:96: in from_env
    return cls(
../../../.pyenv/versions/3.11.4/envs/myproj/lib/python3.11/site-packages/docker/client.py:45: in __init__
    self.api = APIClient(*args, **kwargs)
../../../.pyenv/versions/3.11.4/envs/myproj/lib/python3.11/site-packages/docker/api/client.py:197: in __init__
    self._version = self._retrieve_server_version()
../../../.pyenv/versions/3.11.4/envs/myproj/lib/python3.11/site-packages/docker/api/client.py:221: in _retrieve_server_version
    raise DockerException(
E   docker.errors.DockerException: Error while fetching server API version: ('Connection aborted.', FileNotFoundError(2, 'No such file or directory'))
```

If I set `DOCKER_HOST=unix:///Users/brhiggins/.docker/run/docker.sock` in my shell or in Python before calling `docker.from_env()`, it succeeds.

I'll concede that the name `from_env` sort of sounds like it _only_ considers environment variables, but I had read it as creating a client from "the current environment" more broadly, and I had expected it to behave the same as the `docker` CLI.

dingobar:
Facing the exact issue.

```
NAME       DESCRIPTION                               DOCKER ENDPOINT                                           ERROR
colima *   colima                                    unix:///Users/dingobar/.colima/default/docker.sock
default    Current DOCKER_HOST based configuration   unix:///var/run/docker.sock
```

In this scenario I'd expect `docker-py` to use the Colima context.

AstraLuma:
I'm doing this:

```
[...]

PR #3401, opened 2026-05-13
[DKP-2535] Honour context if present for default client, add contexts support

ebriney:
- related to: https://github.com/docker/docker-py/issues/3146

Summary of changes:

  - docker/context/api.py — new ContextAPI.kwargs_from_context(name=None, environment=None) helper that resolves a Docker CLI context (honouring DOCKER_CONTEXT, currentContext in ~/.docker/config.json, then default) and returns {base_url, tls} ready
  for APIClient.
  - docker/client.py —
    - DockerClient.from_env() now falls back to the current context when DOCKER_HOST is unset, so on a Docker Desktop machine docker.from_env() automatically talks to desktop-linux. Opt out with use_context=False.
    - New DockerClient.from_context(name=None, ...) classmethod for explicit context selection, also exported as docker.from_context.
  - docker/__init__.py — re-exports from_context.
  - tests/unit/client_test.py —
    - Pinned the two from_env pool-size tests with use_context=False so they stay deterministic regardless of the developer's current context.
    - New FromContextTest class covering: current-context fallback, DOCKER_HOST precedence, opt-out via use_context=False, from_context(name=...), unknown-context safety, and default-context resolution.

  All 14 new/adjusted tests pass; the remaining pytest tests/unit/ failures are pre-existing on main (verified by stashing and re-running). The behavior matches the Docker CLI: DOCKER_HOST wins, then DOCKER_CONTEXT, then currentContext, then default.

### DIFF
--- a/docker/client.py
+++ b/docker/client.py
@@ -1,5 +1,8 @@
+import os
+
 from .api.client import APIClient
 from .constants import DEFAULT_MAX_POOL_SIZE, DEFAULT_TIMEOUT_SECONDS
+from .context import ContextAPI
 from .models.configs import ConfigCollection
 from .models.containers import ContainerCollection
 from .models.images import ImageCollection
@@ -78,6 +81,10 @@ class DockerClient:
             use_ssh_client (bool): If set to `True`, an ssh connection is
                 made via shelling out to the ssh client. Ensure the ssh
                 client is installed and configured on the host.
+            use_context (bool): If ``True`` (the default), fall back to the
+                current Docker CLI context (``~/.docker/config.json`` /
+                ``DOCKER_CONTEXT``) when ``DOCKER_HOST`` is not set. This
+                allows the client to talk to Docker Desktop out of the box.
 
         Example:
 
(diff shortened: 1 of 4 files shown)


docker/client.py
```diff
@@ -91,12 +98,66 @@ def from_env(cls, **kwargs):
         max_pool_size = kwargs.pop('max_pool_size', DEFAULT_MAX_POOL_SIZE)
         version = kwargs.pop('version', None)
         use_ssh_client = kwargs.pop('use_ssh_client', False)
+        use_context = kwargs.pop('use_context', True)
```
vvoland, line 101: Making this `True` by default is potentially a breaking change.
Not sure what's the best approach

ebriney: `DOCKER_HOST` stills get precedence.
It will only break for people who changed context that's a bit the goal of the PR.
Default context is targeting default value so for users that didn't touch contexts it will be the same behavior, no?

ebriney: Perhaps we use context as default but switch back to previous implem if DOCKER_API_SKIP_CONTEXT=1 is set.
Lets users escape-hatch if it breaks them.

vvoland: Users could already set `DOCKER_CONTEXT=default` as an escape hatch right? So probably not worth adding another one.

vvoland: I agree this is a good default though.
Maybe we should just release this as next major version to properly signal this.

ebriney: Yes, sounds like a good idea to bump major.
All the more so latest release is quite old: May 23, 2024

vvoland (approved):

PR #3410, opened 2026-06-12
integration/swarm: Preserve rotated unlock key

vvoland:
Fixes flaky SwarmTest.test_update_node erroring out with "Swarm is encrypted and needs to be unlocked before it can be used. Please use "docker swarm unlock" to unlock it."

`test_rotate_manager_unlock_key` enables autolock and then rotates the manager unlock key, but teardown still holds the original key.

If the swarm needs to be unlocked during cleanup, the stale key cannot unlock it and the daemon can remain in locked swarm state for the next test.

Store the rotated unlock key so teardown can unlock and leave the swarm cleanly.


thaJeztah (approved):
SGTM

PR #3414, opened 2026-07-08
tests: fix hardcoded skip in test_import_from_url

Yashasm18:
Replace skipif(True, ...) with a runtime check that detects whether tests are running inside a Docker container by checking for /.dockerenv. This allows the test to run normally outside containers while still skipping inside them where the temporary HTTP server is unreachable.

Relates to moby/moby#52754

### DIFF
--- a/tests/integration/api_image_test.py
+++ b/tests/integration/api_image_test.py
@@ -1,5 +1,6 @@
 import contextlib
 import json
+import os
 import shutil
 import socket
 import socketserver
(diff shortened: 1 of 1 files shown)


tests/integration/api_image_test.py
```diff
@@ -278,7 +279,11 @@ def do_GET(self):
 
         server.shutdown()
 
-    @pytest.mark.skipif(True, reason="Doesn't work inside a container - FIXME")
+    @pytest.mark.skipif(
+    os.environ.get('DOCKER_HOST', '').startswith('unix:///') and
+    os.path.exists('/.dockerenv'),
+    reason="Doesn't work inside a container - HTTP server unreachable from within Docker"
```
thaJeztah, line 285: I see this skip was originally added in https://github.com/docker/docker-py/commit/6c0f7187a8cc1abbf155eae5589c67f9e83cc9bf

- https://github.com/docker/docker-py/pull/747


That PR added various similar skips ("Doesn't work inside a container - FIXME"), but a quick search shows that all of those skips were removed _except_ for this one.


Do we know _why_ this one doesn't work inside a container? And perhaps should we check if that's still the case, or if the Skip could be removed?

If it can't be removed, it would be good to leave a comment that describes _why_ it doesn't work (so that potentially it could be fixed if it's not easy to fix short term)

PR #3417, opened 2026-07-10
harden GitHub Actions workflows with zizmor

crazy-max:
similar to https://github.com/moby/buildkit/pull/6623

This adds zizmor coverage for GitHub Actions workflows and tightens the workflows so the reported findings are either fixed directly or documented where the current trust model is intentional.

Dependabot is also configured to keep GitHub Actions updates moving.

Also the release workflow now uses the better maintained `softprops/action-gh-release` to be consistent with other repos in the org.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,9 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    open-pull-requests-limit: 10
+    directory: "/"
+    schedule:
+      interval: "daily"
+    cooldown:
+      default-days: 7
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,85 +1,130 @@
-name: Python package
+name: ci
 
-on: [push, pull_request]
+permissions:
+  contents: read
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+on:
+  push:
+    branches:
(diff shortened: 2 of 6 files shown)


docker-agent (commented):
### Assessment: 🟢 APPROVE

The workflows are well-hardened: permissions are appropriately scoped, action references are pinned to SHA digests, `pull_request_target` triggers use safe patterns, and the switch to `softprops/action-gh-release` is consistent with org conventions. No bugs or security issues were found in the introduced changes.

docker-agent (commented):
### Assessment: 🟢 APPROVE

This PR cleanly hardens the GitHub Actions workflows: permissions are narrowed at both the workflow and job levels, all action references are pinned to SHA digests, zizmor scanning is integrated, Dependabot is configured for automatic action updates, and the release workflow switches to the better-maintained `softprops/action-gh-release`. No confirmed or likely bugs were found in the added code.

#### Dismissed security findings (review manually)
- `.github/workflows/zizmor.yml`:20 — Third-party reusable workflow (`crazy-max/.github`) with `security-events: write` — transitive action pinning within the external workflow is not guaranteed (verifier mitigation: direct dependency is pinned to commit SHA `46267a6e...`; `crazy-max` is a well-known maintainer and the PR author; `security-events: write` is narrowly scoped for SARIF upload only; the concern is speculative without evidence of a compromised transitive dependency)

vvoland (approved):

